PR #17, opened 2016-10-29
Added (alleged) BSIP-0017 - revive BitAssets after global settlement

pmconrad:
Created a BSIP for resetting BitAssets that have seen global settlement.
Please review + accept. :-)

### DIFF
--- a/bsip-0017.md
+++ b/bsip-0017.md
@@ -4,7 +4,7 @@
     Status: Draft
     Type: Protocol
     Created: 2016-10-27
-    Discussion: <url>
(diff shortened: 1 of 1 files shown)


abitmore:
Some concerns:
* Holders of DUCK are perhaps not in the white list of BACK (but perhaps it doesn't matter).
* Currently, new chain parameter can only be added in `extensions`. Perhaps better implement https://github.com/cryptonomex/graphene/issues/554 first?

pmconrad:
Excellent points, thanks abit! Will copy them over to btstalk.

PR #25, opened 2017-06-23
Updating BSIP19

grctest:
I have removed UIA From BSIP19, focusing on MPA. A separate BSIP will be raised to focus on implementing profit-sharing for UIA.

I have updated BSIP19, taking into account community feedback.


grctest:
I'll submit an updated pull request in the near future for the updated 19, and new 20/21.

PR #26, opened 2017-06-25
Bsip 0018

pmconrad:
Replaced the "auto-revive empty" with auto-recollateralization after price recovery
Detailed description of bid mechanism

### DIFF
--- a/bsip-0018.md
+++ b/bsip-0018.md
@@ -145,7 +145,7 @@ is in the interest of the MPA holders.
 The operation has the following payload:
 
 * `fee`(asset_type): The operation requires a fee to be paid
-* `account`(account_type): This account pays the additional collateral and
+* `bidder`(account_type): This account pays the additional collateral and
   will become the owner of the resulting call position, if the bid is accepted
 * `additional_collateral`(asset_type): Collateral paid by the account in order
   to support the call position
@@ -158,7 +158,7 @@ The operation works as follows:
    cancelled (see below).
 3. If `debt_covered` equals 0, no further action is taken.
 4. It reduces the account's balance by `additional_collateral`.
-5. It creates a `collateral_bid_object` containing the `account` and the partial
+5. It creates a `collateral_bid_object` containing the `bidder` and the partial
    inverted swan price calculated as `additional_collateral` / `debt_covered`.
 
 The required validity checks for the operation are:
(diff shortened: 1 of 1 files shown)


bsip-0018.md
```diff
@@ -88,104 +92,159 @@ This bug will be fixed. See
...
+is in the interest of the MPA holders.
+
+### `bid_collateral_operation`
+
+The operation has the following payload:
+
+* `fee`(asset_type): The operation requires a fee to be paid
+* `bidder`(account_type): This account pays the additional collateral and
+  will become the owner of the resulting call position, if the bid is accepted
+* `additional_collateral`(asset_type): Collateral paid by the account in order
```
xeroc: Can we call this just `collateral`. I don't think the `additional_` has any particular meaning, right?

bsip-0018.md
```diff
@@ -88,104 +92,159 @@ This bug will be fixed. See
...
- * the account has sufficient balance to cover `additional_collateral`
-
-If the checks are successful, a `call_order_object` belonging to the
-investor will be created or updated as described above. Then, the
-settlement price and `settlement_fund` will be cleared.
-
-The fee for this operation will be paid by the investor/recoverer. The
-fee is equal to the fee of the `call_order_update` operation.
+1. It pays a fee
+2. If account has already placed a bid on the same MPA, the existing bid is
```
xeroc: To my understanding this makes the operation state-dependent which isn't the case for any other operation.
From only the operation you cannot see if it is a bid, or a cancel, then.
I would propose to just add another operation for canceling those bids. Or make cancelation implicit with `deb_covered=0`

xeroc:
Good job and thank you for making the bsip more clear.
The only concern that I have is for the state dependency of the `bid_collateral_operation` as describes in a line comment

bsip-0018.md
pmconrad: Perhaps it is not clear what I meant here.
If the account already has a bid, the existing bid is replaced by the new one. This will be implemented as two steps - first cancel the old bid (step 2 in my list), then create the new bid (step 5 in my list).
Step 3 implements your second suggestion - after the old bid has been canceled, a new bid will only be created if debt_covered is > 0, in other words "creating" a bid with debt_covered = 0 cancels an existing bid.

PR #41, opened 2017-11-11
update proposed changes

oxarbitrage:


bsip-0028.md
```diff
@@ -28,9 +28,14 @@ Worker Proposals cannot be updated at all
 
 ## Proposed Changes
 
-* Allow description of the WP to updated.
-* Allow WP Issuer ability to retire WP
-* Possibly allow Committee to retire inactive WP
```
abitmore: I think it's better to keep the descriptive info (above lines) somewhere in the BSIP.

PR #56, opened 2018-02-16
Add BSIP 32: Always Match Orders At Maker Price

abitmore:
BSIP for https://github.com/bitshares/bitshares-core/issues/338.

### DIFF
--- a/bsip-0032.md
+++ b/bsip-0032.md
@@ -11,44 +11,47 @@
 # Abstract
 
 Currently, under most circumstances, when matching two orders, the maker price
-will be used. That said, the order that is placed earlier sets a price,
-another order that is placed later accepts the price, thus a match, two orders
-pay to each other at that price. For example, if one person (A) placed a limit
+will be used to calculate how much each order will pay and receive.
+However, when matching a taker limit order with a maker margin call order,
+the taker price is being used.
+
+This BSIP proposes a principle: always match orders at maker price.
+
+# Motivation
+
+Generally, the order that is placed earlier (the maker) sets a price,
+another order that is placed later (the taker) accepts the price, thus a match,
+two orders pay to each other at that price.
+
+Take the pure limit order matching mechanism in BitShares as an example:
+If one person (A) placed a limit
 order to sell 100 BTS at 0.1 USD per BTS, another person (B) then placed a new
 limit order to buy 100 BTS at 0.105 USD per BTS, the two orders will match at
 0.1 USD per BTS, so A will pay 100 BTS and get 10 USD, B will pay 10 USD and
 get 100 BTS.
 
-However, when matching a taker limit order with a maker margin call order,
-the taker price is being used. For example, if trader A's margin call order is
+However, in BitShares, when matching a taker limit order with a maker margin
+call order, the taker price is being used.
+For example, if trader A's margin call order is
 selling 100 BTS at no less than 0.1 USD per BTS, then trader B placed an order
 that buys 100 BTS at 0.105 USD per BTS, the two order will match at 0.105 USD
 per BTS, so A will pay 100 BTS and get 10.5 USD, B will pay 10.5 USD and get
 100 BTS.
 
-This BSIP proposes a new behavior: always match orders at maker price.
-
-# Motivation
-
-Make the exchange system more user-friendly.
+While not strictly a bug, this behavior is unexpected and irritating for users.
 
 # Rationale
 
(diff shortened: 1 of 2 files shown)


bsip-0032.md
```diff
@@ -0,0 +1,92 @@
...
+    Title: Always Match Orders At Maker Price
+    Author: Abit More <https://github.com/abitmore>
+    Status: Draft
+    Type: Protocol
+    Created: 2018-02-16
+    Discussion: https://github.com/bitshares/bitshares-core/issues/338
+    Replaces: -
+    Worker: To be done
+
+# Abstract
```
pmconrad: IMO most of this Abstract belongs in the Motivation section.
Abstract is only a concise overview of the BSIP.
Motivation explains Why the change is proposed.
Rationale explains specific decisions that were made within the Specification.

bsip-0032.md
```diff
@@ -0,0 +1,92 @@
...
+selling 100 BTS at no less than 0.1 USD per BTS, then trader B placed an order
+that buys 100 BTS at 0.105 USD per BTS, the two order will match at 0.105 USD
+per BTS, so A will pay 100 BTS and get 10.5 USD, B will pay 10.5 USD and get
+100 BTS.
+
+This BSIP proposes a new behavior: always match orders at maker price.
+
+# Motivation
+
+Make the exchange system more user-friendly.
```
pmconrad: Currently, the market engine shows unexpected behaviour in certain situations.
In particular, when placing a limit (e.g. sell) order with a price beyond the highest buy, the order is expected to "walk the book", matching each order on the opposite side in turn, at that order's price, until the new limit order is completely filled, or there are no more buy orders matching its price.
Usually, this expectation is fulfilled. However, when there is a margin call, the margin call will be preferred in matching the incoming order, filling it at its given price. While not strictly a bug, this behaviour is unexpected and irritating for users.

This BSIP proposes more simple order matching rules-

bsip-0032.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Rationale
+
+To attract more users, the system should be fair.
+
+It's common sense that orders should be matched at maker price.
```
pmconrad: "Common sense" is a bad argument.

Suggestion:
> Matching orders at the maker price, with margin calls being inlined in the order book, is an easy to understand rule and matches user expectations, see https://github.com/bitshares/bitshares-core/issues/388.

bsip-0032.md
```diff
@@ -0,0 +1,92 @@
...
+When triggering a black swan event when MSSP is below 100CRP,
+sometimes the short
+position with least collateral ratio may still have more than 100% collateral
+ratio. In this case, the global settlement price is 100CRP but not the actual
+collateral ratio of the short position with least collateral ratio.
+This is current behavior, it's fair, no need to change.
+
+# Discussion
+
+[to be added if any]
```
pmconrad: Suggestion:

> It might seem unfair on the shorter to match at MSSP even if the incoming order specifies a better price. However, in a rationally acting market users will not, in the presence of margin calls, create limit orders below the MSSP. Effectively the new rule doesn't change this situation.

abitmore:
Thank you @pmconrad. I had a hard time organizing the document :-/ Constructive suggestions are welcomed.

bsip-0032.md
```diff
@@ -0,0 +1,92 @@
...
+`check_call_orders(...)` function of `database` class, price of limit order
+is always used. It need to be changed to use MSSP when the call order is maker.
+
+Currently a black swan event will occur when the call order with least
+collateral ratio is going to be matched below 100% collateral ratio price
+(name it `100CRP`). Because the call order will be matched with incoming limit
+order at limit order's price (taker price),
+which can be higher or lower than 100CRP, so even if MSSP is below 100CRP,
+an incoming taker limit order may or may not trigger a black swan.
+So it makes sense to check if a black swan event will occur every time when a
```
pmconrad: "makes sense" is a phrase that indicates this explanation should be moved to the Rationale section.

bsip-0032.md
abitmore: This is "abstract" or "motivation"? Looks like abstract, but it's long.. what should we put in the other section?

abitmore: Actually the call-order-take-precedence rule is discussed in another BSIP, so I don't want to mention it here, but only the match price issue.

xeroc:
+5%

PR #57, opened 2018-02-17
Add BSIP 33: Maker Orders With Better Prices Take Precedence

abitmore:
BSIP for https://github.com/bitshares/bitshares-core/issues/625 and https://github.com/bitshares/bitshares-core/issues/453.

### DIFF
--- a/bsip-0033.md
+++ b/bsip-0033.md
@@ -4,7 +4,8 @@
     Status: Draft
     Type: Protocol
     Created: 2018-02-17
-    Discussion: https://github.com/bitshares/bitshares-core/issues/625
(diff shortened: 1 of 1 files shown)


pmconrad:
Hm, IMO there's a big overlap between this and BSIP-32, perhaps merge them into one?

abitmore:
IMO they're two things, so I prefer two BSIP's.

By the way I'll reorganize this as well.

xeroc:
Can you maybe clarify to the reader what the difference to #56 (bsip32) is?

abitmore:
@xeroc The title explains.

PR #59, opened 2018-02-19
Add BSIP 35: Mitigate Rounding Issue On Order Matching

abitmore:
BSIP for https://github.com/bitshares/bitshares-core/issues/132 and https://github.com/bitshares/bitshares-core/issues/184.

//Update: changed this BSIP to an overall solution for rounding issues, so it will also resolve https://github.com/bitshares/bitshares-core/issues/342. Title changed accordingly.

### DIFF
--- a/bsip-0035.md
+++ b/bsip-0035.md
@@ -182,15 +182,23 @@ The detailed rules proposes in this BSIP (new rules highlighted):
     * **if the smaller order would get nothing after the round-down, cancel it**
   * when matching a limit order with a call order, in favor of call order,
     round down receiving collateral amount
-    * **if the call order is receiving the whole debt (so the short position
-   will be closed) but paying nothing, let it pay 1 Satoshi (round up);**
+    * **if the call order is receiving the whole debt amount (so the short
+   position will be closed) but paying nothing, let it pay 1 Satoshi
+   (round up);**
     * **otherwise, if the limit order would get nothing after the round-down,
    cancel it (it's smaller, so safe to cancel)**
   * when matching a settle order with a call order, in favor of call order,
     round down receiving collateral amount
-    * **if the settle order would get nothing after the round-down, give it 1
-   Satoshi (round up); after paid both side, check if a black swan event would
-   be triggered by the round-up, if yes, trigger it**
+    * **if the call order is receiving the whole debt amount (so the short
+   position will be closed) but paying nothing, let it pay 1 Satoshi
+   (round up);**
+    * **otherwise, if the settle order would be completely filled but would
+   receive nothing, cancel it;**
+    * **otherwise, it means both orders won't be completely filled, which may
+   due to hitting `maximum_force_settlement_volume`, in this case, don't fill
+   any one of the two orders, and stop matching for this asset at this block;**
+    * **that said, only round up when the call order is completely filled, so
+   won't trigger a black swan event, nor need to check for it.**
   * when globally settling, in favor of call order, round down receiving
     collateral amount
     * **when the asset is not a prediction market, if a call order would pay
@@ -243,15 +251,23 @@ so the order will be cancelled later.
 In `check_call_orders(...)` function of `database` class,
 after calculated `order_receives`, check if it is zero.
 If the answer is `true`,
-* if `call_receives` is equal to `call_itr->debt`, set `order_receives` to 1;
+* if `call_receives` is equal to `call_itr->debt`, set `order_receives` to `1`;
 * otherwise, skip filling and cancel the limit order.
 
 ## When Matching A Settle Order With A Call Order
 
 In `match( const call_order_object&, ... )` function of `database` class,
-after calculated `call_pays`, round up it to `1` if it is zero.
+after calculated `call_pays`, check if it is zero.
+If the answer is `true`,
+* if `call_receives` is equal to `call_debt`, set `call_pays` to `1`;
+* otherwise, if `call_receives` is equal to `settle.balance`,
+  call `cancel_order(...)` with parameter set to `settle`,
+  then return a zero-amount collateral asset object;
+* otherwise, return a zero-amount collateral asset object directly.
 
-If rounded up, after filled both orders, check and allow a black swan event.
+After returned, need to check the amount of returned asset at where calling the
+`match(...)` function, specifically, `clear_expired_orders()` function of
+`database` class. If the returned amount is `0`, break out of the `while` loop.
 
 ## When Globally Settling
 
--- a/bsip-0035.md
+++ b/bsip-0035.md
@@ -267,8 +267,9 @@ If the answer is `true`,
 
 After returned, need to check the amount of returned asset at where calling the
 `match(...)` function, specifically, `clear_expired_orders()` function of
-`database` class. If the returned amount is `0`, break out of the `while` loop,
-and label that processing of this asset has completed. Also, in the outer loop,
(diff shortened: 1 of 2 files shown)


xeroc:
Very good bsip and easy to follow.
The proposed solution appears efficient and easy to implement.

This:

>There is an argument says breaking the `min_to_receive` limit is a no-go,
>because that's why it's called a "limit order".

... I agree with.

abitmore:
@xeroc If I'm correct share holder summary is not a must. After this draft listed in the main page we can start stake based voting, although if you as an important proxy don't agree it won't pass, in this case can you help revise the solution, or start a discussion somewhere?

abitmore:
Actually there is a serious issue in this BSIP.

When matching a taker limit order and a margin call order, if `debt * MSSP < 1 Satoshi of collateral`, if round down in favor of the call order, it means the limit order will get nothing. In the proposed solution, the limit order will be cancelled. **This means nobody can ever fill that call order.** As a result, the whole market will get stuck unless something helpful occurs.

abitmore:
I think we need to round up the paying collateral to 1 Satoshi in this case.

abitmore:
Here came more thoughts about settle orders.

A settle order may suffer something-for-nothing issue when
* the matching call order is too small, or
* the settle order itself is too small, or
* reached `max_settlement_volume` in current maintenance interval.

We need to react differently for these scenarios. IMHO,
* if the matching call order is too small, let it pay 1 Satoshi
* if the settle order itself is too small, cancel it
* if reached or about to reach `max_settlement_volume`, stop matching at this block.

bsip-0035.md
```diff
@@ -0,0 +1,306 @@
...
+Downside:  Complicates the implementation of flipped market UI, may require
+re-working part of market GUI, reduces user flexibility, new asset fields
+required to specify precision, if `n` assets exist then `O(n^2)` markets
+could exist and we need to figure out how to determine the precision
+requirement for all of them.
+
+## The Chosen Solution
+
+Current code actually implemented (a) in the first place: when matching two
+orders, if there is a rounding issue, the order with smaller volume will be
```
pmconrad: Please specify how order volumes are compared to determine which is "smaller".

abitmore: Done.

abitmore:
The something-for-nothing issue exists as well when force settling against the global settlement fund (force settlement after a black swan event). [Code](https://github.com/bitshares/bitshares-core/blob/2.0.171105a/libraries/chain/asset_evaluator.cpp#L489):
```
      auto settled_amount = op.amount * bitasset.settlement_price;
```

abitmore:
Something-for-nothing is only a subset of rounding issues, it's the most extreme one. There are other scenarios that one of the matched parties would be paying more than enough, although they're not paying something for nothing overall. There are discussions in https://github.com/bitshares/bitshares-core/issues/342.

I got an idea about how to mitigate the rounding issue, but wonder if it's better to add it to this BSIP (or even replaces this BSIP), or write it as another BSIP.

The principle would be: never pay more than enough when matching orders. Nevertheless, by paying "enough", it's possible that the party still need to pay more than preferred.

Current [code](https://github.com/bitshares/bitshares-core/blob/2.0.171105a/libraries/chain/db_market.cpp#L311-L324) writes (comment added by me):
```
   if( usd_for_sale <= core_for_sale * match_price )
   {
      core_receives = usd_for_sale;
      usd_receives  = usd_for_sale * match_price; // round down, in favor of bigger order
   }
   ...
```

It can be explained as: when `usd_for_sale` is of the smaller order, it's assigned to `core_receives` directly. However, `core_receives` is usually more than enough when `match_price` is smaller than `1`. To mitigate the rounding issue, my idea is to assign `round_up(usd_receives / match_price)` back to `core_receives`, so the code would look like:
```
   if( usd_for_sale <= core_for_sale * match_price )
   {
      core_receives = usd_for_sale;
      usd_receives  = usd_for_sale * match_price; // round down, in favor of bigger order
      core_receives = ceil( usd_receives / match_price ); // Pseudo code, actually can't use operator/() here but need another function
   }
   ...
```

Take a scenario similar to the one described in the 4th comment of https://github.com/bitshares/bitshares-core/issues/132 as an example:
* Alice's order: Sell CORE at `$3 / 80 = $0.0375`, balance `50 CORE`
* Bob's order: Buy CORE at `$19 / 500 = $0.038`, balance `$100`

The original process would be:
* If Alice's order is maker, use `$3 / 80` as match price; since Alice's order is smaller, round in favor of Bob's order, so Alice will pay the whole `50 CORE` and get `round_down(50 CORE * $3 / 80 CORE) = round_down($1.6) = $1`, the effective price would be `$1 / 50 CORE = $0.02`.
* If Bob's order is maker, use `$19 / 500` as match price; since Alice's order is smaller, round in favor of Bob's order, so Alice will pay the whole `50 CORE` and get `round_down(50 CORE * $19 / 500 CORE = round_down($1.9) = $1`, the effective price would still be `$1 / 50 CORE = $0.02`.

Both results are far from Alice's desired price `$0.0375`.

The revised process would be:
* If Alice's order is maker, use `$3 / 80` as match price; since Alice's order is smaller, round in favor of Bob's order, so Alice will get `round_down(50 CORE * $3 / 80 CORE) = round_down($1.6) = $1`, **then, calculate how much Alice need to pay as `round_up($1 * 80 CORE / $3) = round_up(26.67 CORE) = 27 CORE`, so the effective price would be `$1 / 27 CORE = $0.037`**.
* If Bob's order is maker, use `$19 / 500` as match price; since Alice's order is smaller, round in favor of Bob's order, so Alice will get `round_down(50 CORE * $19 / 500 CORE = round_down($1.9) = $1`, **then, calculate how much Alice need to pay as `round_up($1 * 500 CORE / $19) = round_up(26.3 CORE) = 27 CORE`, so the effective price would be `$1 / 27 CORE = $0.037`**.

Both results are still below Alice's desired price `$0.0375`, but they're much closed to `$0.0375` in comparison to `$0.02`.

Above are processes about processing limit orders. After get `$1`, the remaining amount of Alice's order is too small to get anything, so the order will be cancelled. However, call orders can't be cancelled so need to be processed differently. I'll leave this topic here to be discussed later.

abitmore:
Changed this BSIP to an overall solution for rounding issues, so it will also resolve bitshares/bitshares-core#342. Title changed accordingly.

TheTaconator:
Outstanding description of a complicated issue! The enumeration of the various combinations of order types was very helpful to me.

I like the idea of canceling orders that are predicted to be "too small" to be fulfilled in the future. But how can "too small" be determined? Is the answer, "when any future matching would result in a "something-for-nothing" scenario"?

TheTaconator:
Example 2 had me wonder about something.

Could there be a way for someone to make a black swan **more likely** by borrowing such a tiny amount (perhaps with a tiny amount of collateral?) that black swans could more easily be triggered during a market downturn? (Even if my concern is real, perhaps it is not a situation that is introduced by the proposed solution.)

abitmore:
>how can "too small" be determined? Is the answer, "when any future matching would result in a "something-for-nothing" scenario"?

Yes. Only need to check if its `price * amount` is less than `1 Satoshi`.

abitmore:
The black swan scenario introduced in example 2 is not due to amount too small, but due to collateral ratio too low. Tiny call orders are more likely to be filled when entered margin call territory. By the way, the scenario won't happen after [BSIP 32](https://github.com/bitshares/bsips/blob/master/bsip-0032.md) is implemented.

PR #61, opened 2018-02-23
Add BSIP 36: Remove expired feeds

oxarbitrage:


bsip-0036.md
```diff
@@ -0,0 +1,73 @@
...
+This BSIP proposes a protocol change to remove expired feeds at maintenance time.
+
+# Motivation
+
+Improve performance of the bitshares-core.
+
+
+# Rationale
+
+This expired feeds are unnecessary, they clutter up the database and cause unnecessary burden on undo_db.
```
pmconrad: Perhaps better:
> The proposed change only affects the feed price when the feed expiration time is increased. This rarely happens, if ever, so in practice the effect on pricing is negligible. The savings in performance however, is not.

bsip-0036.md
```diff
@@ -0,0 +1,73 @@
+    BSIP: 0036
+    Title: Remove expired feeds
+    Author: oxarbitrage <https://github.com/oxarbitrage>
+    Status: Draft
+    Type: Protocol
+    Created: 2018-02-22
+    Discussion: https://github.com/bitshares/bitshares-core/issues/518
```
pmconrad: Please create a forum thread for discussion.

PR #63, opened 2018-02-25
add bsip 36 and 37 to readme

oxarbitrage:


README.md
```diff
@@ -42,3 +42,7 @@ Number             | Title                                                    |
 [32](bsip-0032.md) | Always Match Orders At Maker Price                                          | Abit More | Protocol | Draft
 [33](bsip-0033.md) | Maker Orders With Better Prices Take Precedence                             | Abit More | Protocol | Draft
 [34](bsip-0034.md) | Always Trigger Margin Call When Call Price Above Or At Price Feed           | Abit More | Protocol | Draft
+[36](bsip-0036.md) | Remove expired feeds                                                        | oxarbitrage | Protocol | Draft
```
abitmore: To be clearer, how about "Remove expired price feeds on maintenance interval"?

README.md
```diff
@@ -42,3 +42,7 @@ Number             | Title                                                    |
...
+[37](bsip-0037.md) | Extend created asset names                                                  | oxarbitrage | Protocol | Draft
```
abitmore: This title looks very strange. Why it's "created"? How will we "extend"?

I think it's better to use something like "Allow new asset name to end with a number".

README.md
oxarbitrage: changed, sent pull.

README.md
oxarbitrage: changed, sent pull.

Issue #51, opened 2018-02-06
New BSIP: Call Order Options (Stop-Loss and etc)


abitmore:
Just write some thoughts here, perhaps will make a detailed BSIP later. People who have interest can help compose as well.

After https://github.com/bitshares/bitshares-core/issues/343 is fixed, call price of call orders (short positions) will be updated after every match, in favor of consistent margin call process logic and black swan detection. But some traders DO NOT want this behavior, for stop-loss purpose for example.

Here we propose some new options that traders can set onto their call positions:
* target CR: an optional collateral ratio, when the position being automatically released, sell no more than required collateral until CR of the position reaches this value.
  * default: not set, which means sell as much collateral as possible
* stop-loss price: an optional price to release collateral automatically when feed_price is below it (in terms of collateral/debt), the behavior will be placing some collateral on market for sale, at the price equals to `feed_price/(1+MSSR)`, until target CR is reached.
  * default: not set, which means will sell collateral only when CR < MCR
* take-profit price: an optional price to release collateral automatically when feed_price is above it (in terms of collateral/debt), the behavior will be placing some collateral on market for sale at the specified price to repay the whole debt (capped by debt amount).
  * default: not set, which means never take profit automatically

bitcrab:
I like these proposed logics, I believe they can help smartcoin producers to avoid risks.

thul3:
Agree on that.Major functions which are very demanded on big exchanges.
If DEX wants to go big this is something it needs

jmjatlanta:
In the US equities world, this is often split into separate orders, managed by the broker's system. The filling of a "parent order" (a.k.a. OSO or "One Sends Other"), even partially, triggers other orders,  which are also interrelated.

Child orders can be set up to cancel the others if they are triggered (a.k.a. OCO (One Cancels Other) a.k.a. "Bracket Order")

Example scenario: a parent order is partially or fully filled, triggering stop and limit orders (bracket orders) to be activated with the same size, both of which will cancel the other upon their fill.

Advantages to the client: automation. They do not need to remain connected to manage the order. The plan is in the hands of the broker. 

It could also be said that concepts are easy to grasp for the client, as simple orders remain simple. An order is an order. If you don't know what a stop order is, you don't have to use it. If you know what you're doing, you can get fairly complex with combinations of OCO and OSO orders.

The disadvantage is to the broker (in this case, the exchange): They are responsible for execution and the overhead of order management. But this also exists with the "all in one order" originally mentioned.

[...]

PR #70, opened 2018-03-05
BSIP 38: add target CR option to short positions

abitmore:
1/3 for https://github.com/bitshares/bsips/issues/51.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,3 +45,4 @@ Number             | Title                                                    | 
 [35](bsip-0035.md) | Mitigate Rounding Issue On Order Matching                                   | Abit More | Protocol | Accepted
 [36](bsip-0036.md) | Remove expired price feeds on maintenance interval                          | oxarbitrage | Protocol | Accepted
 [37](bsip-0037.md) | Allow new asset name to end with a number                                   | oxarbitrage | Protocol | Accepted
+[38](bsip-0038.md) | Add target collateral ratio option to short positions                       | Abit More | Protocol | Draft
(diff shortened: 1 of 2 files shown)


TheTaconator:
Interesting topic and very good BSIP. The guidance for UI/UX developers is very helpful. My suggestions are:

**Math**

As I understand the math, the target_CR variable is a dimensionless variable. (In contrast, a typical collateral ratio actually has implied dimensions of the collateral unit / debt unit). I suggest a sentence to clarify that the target_CR variable has no dimensions/units.

**How to use the feature to address each of the rationales**

If each of the rationales can be addressed with this new mechanism, it would be helpful to an end-user to know how to use it. If my understanding is correct, I think that they are as follows:

> * some want to close their short positions completely to cut losses;

Set the target_CR to a very high value (maybe 6553.5%)

> * some want to sell as little collateral as possible to keep remaining short positions as large as possible;

Do not set the target_CR at all

> * some want to sell more than minimum required collateral to reduce the possibility of being margin called again in the near future, but don't want to close their short positions completely.

Set the target_CR to a value larger than the MCR and less than maximum target_CR (6553.5%?)

abitmore:
@TheTaconator thanks.

If I understood your question correctly, all collateral ratios in BitShares are dimensionless quantities.

About how to use the feature, it's easy to understand: selling collateral increases CR (collateral ratio). When you're margin called,
* if you don't have a target CR, then all your debt will be covered, maximum collateral will be sold;
* if you have a target, then some of your debt will be covered,
  * the higher your target is, more debt will be covered, so more collateral will be sold.

So,
* to cut loss, don't set target CR (same as current behavior), or set it to very high
* to keep as large position as possible, set target CR to as low as possible, the minimum allowed number is MCR
* otherwise, set target CR to a value some higher than MCR

PR #72, opened 2018-04-08
Update worker and status of BSIP 38

abitmore:


oxarbitrage:
need to solve conflicts to change the status of this bsip to accepted.

PR #74, opened 2018-04-10
Add rounding rule for bsip 38

abitmore:


bsip-0038.md
```diff
@@ -131,6 +131,32 @@ After a call order get matched with a limit order and about to fill,
...
+Rules about rounding are defined in [BSIP 35](bsip-0035.md).
+
+The rule for matching a limit order with a call order need to be revised as
+(new rules **in bold**):
+* if the call order is receiving the whole debt amount, which means it's
+smaller and the short position will be closed after the match, round up its
+paying amount;
+* **otherwise,**
+  * **if `target_collateral_ratio` is set and it's receiving the maximum debt
+ amount calculated with `target_collateral_ratio`, see this order is smaller,
```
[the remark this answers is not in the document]
abitmore: I mean "consider the call order is smaller".

bsip-0038.md
```diff
@@ -131,6 +131,32 @@ After a call order get matched with a limit order and about to fill,
...
+ try to round up its paying amount;**
+    * **for edge cases, if its collateral ratio would not increase after being
+   partially filled due to the round-up (which may even cause a black swan
+   event in an extreme scenario), see `target_collateral_ratio` as not set for
+   this time, re-apply the filling rules for this match.**
+  * otherwise, round down its paying amount.
+    * if the limit order would receive nothing, cancel it (it's smaller,
```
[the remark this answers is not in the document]
abitmore: In this case, the amount that call order will pay will be rounded up, so the limit order won't be receiving nothing.

bsip-0038.md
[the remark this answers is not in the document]
abitmore: Reworded, please check again.

abitmore:
@pmconrad I'm aware that rounding up collateral paid is dangerous, so in this PR I tried to avoid such issues.
* In the "definition" section, if `target_collateral_ratio` is lower than `MCR`, we will use `MCR` which is safer then `MSSR`. To be clearer, I'll mention the rule in "specification" section as well.
* In one of the sections you've commented, I came up with the same solution as you suggested: if rounding up collateral paid would cause a decrease in collateral ratio, we ignore the call order's `target_CR` option.

abitmore:
The order of rounding calculation is wrong. Need to get accurate debt then rounded up first, then get collateral with rounded-up debt and rounded up, then get debt with rounded-up collateral and rounded down. I'll update.

pmconrad:
Implementation note: the max_debt_to_cover calculates with prices (TCR * match_price - feed_price). There will be rounding involved, too.

TheTaconator:
Appreciate the updates and clarifications. Just some suggestions to clarify and simplify the reading.

It would be valuable to define collateral ration (e.g. CR = collateral / debt [which actually is not dimensionless] or CR = collateral / (debt * current_feed_price) [which is dimensionless])

The call order object (call_order_object) will store the (target_cr_value / 1000) as an uint16_t.  Will the call_order_update_operation, and call_order_update_operation also store them as uint16_t within the extensions?


> That said, if a short position got matched with a big limit order, after partially filled, its collateral ratio should be just higher than specified target collateral ratio.

What is the significance of it being a "big limit order"?


> Which means if max_debt_to_cover has no fraction, need to plus it by one Satoshi; otherwise, need to round it up. An effectively same approach is to round down then add one Satoshi onto the result.

Minor suggestion for wording: change "has no fraction" to "has no fractional component (e.g. 5.00 as opposed to 5.23)"


> Rounding on Order Matching, and Edge Cases

Helpful discussions on rounding


> When a call order got matched with a limit order, if the call order has no target_CR option set, or has target_CR option set but covering debt is smaller than maximum debt to cover, both means the call order is larger, so its paid collateral will be rounded down, so its collateral ratio will increase after partially filled.

Why "so its paid collateral will be rounded down"?


> This is against the intention of this BSIP. To solve this issue, if detected a decrease in collateral ratio when matching, we propose to ignore the target_CR option of corresponding call order, and re-evaluate the match.

+1


> if the call order is receiving the whole debt amount, which means it's smaller and the short position will be closed after the match, round up its paying amount;

Is "paying amount" the "amount to sell"?

Why "round up" the paying amount? Is it to ensure the trade will be more than fair to the limit order? Is it to ensure that the whole debt is paid off?


> (which may even cause a black swan event in an extreme scenario)

Can you please give an example for my better understanding?


> When and How To Use the Option

+1


> Specifications

+1


> When there are call orders to be filled, if target_collateral_ratio option is set, UI need to show exact amount of collateral that another trader is able to buy and exact amount of debt that need to pay according to the equation described above.

A basic question: When a call order is placed onto the order book, does its price remain constant even if it is unfulfilled/waiting on the book for some time?

abitmore:
Update: I found at this moment I don't have time to edit this document, need to put more time in coding. @TheTaconator: I think you have write access to this repository, can you please update the document as you suggested and perhaps create a new pull request? Thanks a lot.

bsip-0038.md
```diff
@@ -78,7 +88,121 @@ max_amount_to_sell = (debt * target_CR - collateral * feed_price)
...
+Rounding rules about order matching are defined in [BSIP 35](bsip-0035.md).
+
+Generally, when two orders got matched, the order matching engine will favor
+the larger order while rounding.
+
+When a call order got matched with a limit order, if the call order has no
+`target_CR` option set, or has `target_CR` option set but covering debt is
+smaller than maximum debt to cover, both means the call order is larger,
+so its paid collateral will be rounded down, so its collateral ratio will
+increase after partially filled.
```
TheTaconator: I do not understand the conclusion "both means the call order is larger" for four reasons:

1. I do not see how "If the call order has no "target_CR option set" indicates anything about the size of the call order
2. if the "covering debt is smaller than the maximum debt to cover" it makes me think that this call order will only get partially filled. If that is correct, it makes me think that the call order is smaller than the matching limit order.
3. "so its paid collateral will be rounded down" matches up with the suggested approach to handle partial fills in BSIP-35
4. "so its collateral ratio will increase after partially filled." explicitly says that the call order will be partially filled.

If I have understood this BSIP correctly, partial filling of a call order means that the matching order is smaller than the call order.

abitmore: The first point makes sense. Updated. Thanks. Please check again.

TheTaconator: It makes sense to me now, thanks

abitmore:
@pmconrad updated description about rounding, please review again.

bsip-0038.md
```diff
@@ -110,33 +110,81 @@ The result is a rational number as well.
...
+For `max_amount_to_sell_int` above, we can adjust `max_debt_to_cover_int` with:
 
 ```
 max_debt_to_cover_int = round_down(max_amount_to_sell_int * match_price)
 ```
 
+### Review the result
+
+Due to rounding, it's not guaranteed that selling more collateral will
+always result in higher collateral ratio on remaining call order.
```
TheTaconator: This discussion is important and helpful

PR #76, opened 2018-04-17
Initial draft of escrow BSIP (BSIP40)

TheTaconator:
PR for discussion purposes. Initial draft of an escrow feature informational BSIP. Discusses two concepts **so far**. If short on time, I recommend reading "Concept B".

PR #84, opened 2018-07-14
Create auction feature BSIP xxxx

cogutvalera:
PR for discussion purposes. Initial draft of an auction feature BSIP (currently based on Wiki information https://en.wikipedia.org/wiki/Auction)

### DIFF
--- a/auction-feature-bsip-xxx.md
+++ b/auction-feature-bsip-xxx.md
@@ -49,8 +49,10 @@ Live auctioneers, escrow, appraisal experts, delivery services and online auctio
 
 ## Blockchain key features to host decentralized auctions
 
-![Key features]
-(https://cdn-images-1.medium.com/max/800/1*UzKw5Q9qf_0TT2AhV9fxsg.png)
+1. Scalable, reliable and transparent
+2. Interoperable
+3. At a very low transaction cost
+4. Fast and automatic
 
 # Specifications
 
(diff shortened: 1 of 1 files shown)


cogutvalera:
@ryanRfox @abitmore @xeroc @clockworkgr @jmjatlanta @oxarbitrage do you have any suggestions about this ? or maybe some ideas ? what do you think about it ?

cogutvalera:
maybe we don't need at all such feature and I should close this PR immediately and remove all info ? :-) Just help me please to understand the right direction so I can make more researches if required ...

abitmore:
Good initiative. However, it seems the document doesn't contain a workable specification so far. It didn't describe how auction would look like or would work in BitShares. See https://github.com/bitshares/bsips/pull/76 for an example.

cogutvalera:
@abitmore you're absolutely right ! The document currently doesn't contain a workable specification currently. Just wanted to know if it is really good idea to have such Auction feature in BitShares or not. Thanks for your opinion ! I want to research more BitShares-Core C++ project to understand more clearly enough what is the best implementation for such feature. Also I want to discuss here more ideas and suggestions from other developers. Maybe current BitShares-Core C++ project can fully support now different types of Auctions by already implemented features like proposals, UIA and so on ? Maybe we don't need new feature for such feature because everything already done in BitShares ? :-) But currently I'm thinking about developing new plugin for Auction feature. I want to improve this BSIP more and more after more researches and more discussions about it. Currently this BSIP just describes abstractly different types of Auctions and their core features and parameters. So I need approval from core developers to understand clearly which direction should I choose and if we really need such new feature, so I can spend more time on this.

pmconrad:
I agree with abit here. Good initiative.

Still, this needs to be discussed elsewhere, perhaps in the bitsharestalk forum.

Closing for now. Feel free to re-open once the discussion has turned into a clear view of which type of auctions should be integrated into the core, and how.

PR #86, opened 2018-07-25
Custom active permission

sschiessl-bcp:


abitmore:
Things to think about:
* verify_authority function / API, verify_account_authority API
* get_potential_signatures / get_required_signatures API
* how to evaluate proposals
* complexity

sschiessl-bcp:
Complexity I can't argue. For the rest I was hoping it can be implemented on a deep level where the active authority is granted if a custom active is present, without changing any other existing logic. 

Idea, for example, might be better ways: when you are iterating the given signatures  you are checking the required accounts if they have any custom active authority that matches the given signature and the operation and does pass the asserts. If so, active authority of the required account is granted.

I am not familiar with the backend too much, excuse if that was rubbish

abitmore:
Ideally a custom permission item can be defined as
```
struct custom_permission_object
{
   account_id;
   custom_authority;
   operation_id;
   operation_permission_details;
};
```
Among it, `custom_authority` and `operation` are m-n mapped.

The `operation_permission_details` field is operation-specific and can be very complex, for example, for a transfer, people may want to set limitations on:
* what asset can be used as fee
* how much the fee can be
* who to receive
* what asset can be transferred
* how much can be transferred
  * one-time limit
  * daily limit or similar
* frequency of being executed
* memo size

or some combinations of those limitations, E.G. can transfer at most 100 asset X to account A 3 times a day, at most 1000 asset Y to account B 5 times a day, and etc.

sschiessl-bcp:
```
struct custom_permission_object
{
   account_id;
   custom_authority;
   operation_id;
   operation_permission_details;
};
```
a) This object includes the account_id, so you envision this not to be part of the account_object, correct?
Just to say it: This is exactly the initial idea just forumlated more towards an actual implementation, correct?

b) You are even upgrading the possible asserts, or limitations as you called it. Flexibility would be great, e.g. to define allowed ranges for integer arguments. My only wish would be that the logic is so generic that no custom logic has to be done for a specific operation, do you think thats possible? I would rather restrict the flexibility than to do operation specific custom code.

c) How would conditions that requre historic information (like a daily transfer limit) be implemented? Would you store historic information wihtin the custom_permission_object object as well?

abitmore:
I was just writing what I've thought of. The potential use cases or directions to implement. Some of them might be too hard to implement or have some limitations so not worth the efforts. We need more discussion / brainstorming.

sschiessl-bcp:
> I was just writing what I've thought of. The potential use cases or directions to implement. Some of them might be too hard to implement or have some limitations so not worth the efforts. We need more discussion / brainstorming.

I see. My point of the BSIP was to introduce a security layer, but I guess it could also be extended to allow even more sophisticated use cases. I ll try to go through your points

> - what asset can be used as fee
> - who to receive
> - what asset can be transferred

This is a simple restriction of argument value
```
   assert: {
        arguments: {
            to: {
               value: [ 'list', 'of', 'allowed', 'values']
           }
       }
    }
```

> - frequency of being executed
> - one-time limit

This requires history as one needs to know: Last time of execution. One time limit would then be 
having no or infinite interval_in_sec
```
   assert: {
        frequency: {
           last: timestamp of last execution,
           interval_in_sec: integer, if (last == null || now >= last + interval_in_sec) execution is allowed, and set last
        }
    }
```

> - how much the fee can be
> - how much can be transferred
> - memo size

This is a range limitation. For integers it's the value, for string it's the length. 
```
   assert: {
        arguments: {
            to: {
               min: integer, minimum value, inclusive,
               max: integer, maximum value, inclusive
           }
        }
    }
```

> - daily limit or similar

This is a range limitation. For integers it's the value, for string it's the length. How does this overlap with withdrawal_permission?
```
   assert: {
        arguments: {
            to: {
               limit: {
                  last: timestamp when counter was reset,
                  cumsum: integer, sum of all amounts that happened since last timestamp,
                  max: maximum allowed,
                  interval_in_sec: timeinterval, if (last == null || now >= last + interval_in_sec) set cumsum=0 and set last
           }
        }
    }
```

**So in total this would boil down to the following generic scheme of asserts:**
```
   assert: {
        arguments: {
            argument_name: {
               value: [ 'list', 'of', 'allowed', 'values'],
               min: integer, minimum value, inclusive,
               max: integer, maximum value, inclusive,
               limit: {
                  last: timestamp when counter was reset,
                  cumsum: integer, sum of all amounts that happened since last timestamp,
                  cumsum_max: maximum allowed,
                  interval_in_sec: timeinterval, if (last == null || now >= last + interval_in_sec) set cumsum=0 and set last
           }
        },
        frequency: {
           last: timestamp of last execution,
           interval_in_sec: integer, if (last == null || now >= last + interval_in_sec) execution is allowed, and set last
        }
    }
```

This would allow the most fine-grained control possible, but I am unsure as if can be implemented like this in an efficient manner. It would have my vote as I am in big favor of abstract concepts.

xeroc:
> The operation_permission_details field is operation-specific and can be very complex, for example, for a transfer, people may want to set limitations on:

Our idea was to actual keep it rather simple (if possible). I was thinking along the lines of
* having an regular operation structure as part of the custom operation structure
* each attribute of the operation can have a value or `NULL`
* if `NULL`, the attribute in an operation to be evaluated can have any value
* if not `NULL`, the value of the attribute **has to be exactly the same**.

This way we can grant access to specific operations to specific accounts and limit the attributes of an operation.

### Limitations
For operations that have *complex* attributes (like `new_options` for `account_update_operation`), we cannot limit values with in those complex attributes (think `votes` within `new_options`)

Thus, in order to grant access to a key to **only voting**, we would need to create a new operation that takes the votes directly instead of a nested options attribute.

bsip-0040.md
```diff
@@ -0,0 +1,87 @@
...
+BitShares Blockchain is bad publicity. The user needs to be educated in account security, and this BSIP
+aims to ensure all technical possibilities are met while being flexible to allow many use-cases.
+
+Examples:
+ - Witness Key: Only allows update signing key and publish price feed
+ - Trading Key: Only allows limit orders (arguments restricted to desired markets), update margin position and transfers (arguments restricted to certain accounts)
+ - Proposal Update Key: Approve proposals (2FA comes to mind)
+ - Faucet Key: Allow only to create accounts
+The above list of named keys is nothing that is known to the backend as the backend should have an abstract implementation.
+ The UI should provide a button "Create Trading Key" that properly configures the respective custom active permission entry.
```
pmconrad: UI changes are out of scope for a BSIP. Remove this line, or replace "should" with "could" to make it clear that this is not part of the specification.

bsip-0040.md
```diff
@@ -0,0 +1,87 @@
...
+      Stefan Schießl <https://github.com/sschiessl-bcp>
+    Status: Draft
+    Type: Protocol
+    Created: 2018-07-25
+    Discussion: https://github.com/bitshares/bitshares-core/issues/1061
+    Worker: <Id of worker proposal>
+
+# Abstract
+
+Strengthening user security is one of the main factors to elevate BitShares. Inlight of recent
```
pmconrad: "In light" - two words

bsip-0040.md
```diff
@@ -0,0 +1,87 @@
...
+
+The description here is more on a superficial level and no recommendation how it can best be implemented.
+Custom active permission is a list of custom active authorities. A `custom active authority` contains an `operation_id`, an `authority` (just like with active permission) and `assert`s than can be used to restricted arguments. When a transaction is signed with such an authority the backend checks if the contained operation has a corresponding custom active authority entry and if so acts as if the active authority of the corresponding account is given. It also checks if the arguments are in the allowed range.
+
+A Custom Active Permission looks like follows (in JSON for clarification, backend serializes and stores in a different way):
+```
+custom_active_permission = list of custom_active_authority items
+custom_active_authority = {
+    operationid,
+    auhtority,
```
pmconrad: typo

bsip-0040.md
```diff
@@ -0,0 +1,87 @@
...
+ - Trading Key: Only allows limit orders (arguments restricted to desired markets), update margin position and transfers (arguments restricted to certain accounts)
+ - Proposal Update Key: Approve proposals (2FA comes to mind)
+ - Faucet Key: Allow only to create accounts
+The above list of named keys is nothing that is known to the backend as the backend should have an abstract implementation.
+ The UI should provide a button "Create Trading Key" that properly configures the respective custom active permission entry.
+
+# Rational
+
+The description here is more on a superficial level and no recommendation how it can best be implemented.
+Custom active permission is a list of custom active authorities. A `custom active authority` contains an `operation_id`, an `authority` (just like with active permission) and `assert`s than can be used to restricted arguments. When a transaction is signed with such an authority the backend checks if the contained operation has a corresponding custom active authority entry and if so acts as if the active authority of the corresponding account is given. It also checks if the arguments are in the allowed range.
```
pmconrad: restrict~ed~

bsip-0040.md
```diff
@@ -0,0 +1,87 @@
...
+```
+That has the consquence now that a a transfer transaction sending funds away from A can be signed with key K as long as the receiver is B.
+
+# Specifications
+Requirements to modify the backend includes
+* Extend the account object to store custom active permission
+* Extend `account_update` or create a new operation to allow changing the custom active permission
+* Operation-specific authorities (if present) must be evaluated in incoming transactions
+* Additional committee parameters may be needed to limit the extend of usage of this feature
+
```
pmconrad: Please elaborate:
* can an account specify different authorities for the same operation but different asserts (i. e. require key 1 for transfer to Alice and key 2 for transfer to Bob)?
* list possible assertions for each operation
* specify assert logic, e. g. multiple fields/values, boolean operations, supported operators...

sschiessl-bcp: I have included a more detailed description, please see here

https://github.com/sschiessl-bcp/bsips/blob/patch-1/bsip-0040.md

sschiessl-bcp:
I have updated the specifications of the asserts, please have a look

@abitmore @pmconrad

clockworkgr:
Are asserts based on historical data possible seeing as witness nodes don't really carry history?

sschiessl-bcp:
> Are asserts based on historical data possible seeing as witness nodes don't really carry history?

No history is required, everything that the assert needs to know is stored in its state (`limit` would be a statefull assert)

clockworkgr:
Would it be possible to have a higher fee for using this feature? i.e. op_fee is 10% higher if op has custom permissions?

bsip-0040.md
```diff
@@ -0,0 +1,155 @@
...
+ - Cold Storage Key: Only allow to move funds to the Hot Wallet
+ 
+The above list of named keys is nothing that is known to the backend as the backend should have an abstract implementation.
+ The UI could provide a button "Create Trading Key" that properly configures the respective custom active permission entry.
+
+# Rational
+
+The description here is more on a superficial level and no recommendation how it can best be implemented.
+Custom active permission is a list of `custom active authorities`. A `custom active authority` contains an `operation_id`, an `authority` (just like with active permission) and `asserts` than can be used to restrict arguments and is only valid a certain time period (`valid_from` and `valid_to`).
+When a signed transaction arrives at the backend iterate the required accounts for each operations. If they have any one custom active authority that matches the given operation, signature and passes all asserts of the custom active authority, then the active authority of the required account is granted. This search can be stopped after finding one match.
```
pmconrad: This paragraph belongs in the specification section. The matching logic is important.

For example, what happens if you have a custom active authority that matches the operation but fails the assert, should the search continue? What if no full match is found, does it fall back to normal authority then?

bsip-0040.md
```diff
@@ -0,0 +1,155 @@
...
+       account_auth: []
+    },
+    asserts: [
+        {
+            argument: to
+            function: any,
+            data: [B]
+        }
+    ]
+}
```
pmconrad: Note: this is just an illustration of a possible serialization, not a specification of the serialized format.

bsip-0040.md
sschiessl-bcp: I changed it around according to your suggestion. Your question is answered in section 
   Outline of handling incoming transactions

bsip-0040.md
sschiessl-bcp: added as well

abitmore:
Using accounts in authority can be problematic due to increased complexity. For example, imagine this setup:

Account A
* custom authority 1
  * account B, can transfer 1 asset X to D
* custom authority 2
  * account C, can transfer 2 asset X to D

Since a transaction can only be signed with keys, we need to convert keys to accounts when checking authority. If both account B and C have same key in their active authority, even with different weights, it would be complicated to check.

IMHO, we can have a "signed_by" field in extensions of each operation, to indicate who is signing the transaction, to simply verification process as well as signing process.

sschiessl-bcp:
> Some thoughts here.
> 
>     Using authority adds flexibility as well as complexity
>         some use cases are to grant limited access to special keys, so we need to support keys;
>         one use case is to grant limited access to another account, so we need to support accounts;
>             what if A granted access to B and then B granted access to C? With current permission model, we check recursively, but it will be much more complex to recursively check limited / operation related permissions;
>         what if the authority is by itself complex, e.g. something like {"threshold":3,"account-a":{"weight":2},"key-b":{"weight":1},"account-c":{"weight":1}}?
> 
>     IMHO, on the first iteration, we can limit the authority to:
>         can only have one key or one account, and
>         we won't check recursively
> 
> Update: perhaps we can still check recursively, but with limitations, E.G. only check custom authorities for the first level. With this approach, perhaps we'll be able to deal with complex authority objects.

I"m not sure I can follow. Account A can grant B any operation that requires A as active authority. B can grant C any operation that requires B as active authority. I don't see how that would cascade.

bsip-0040.md
```diff
@@ -0,0 +1,170 @@
...
+| `limit` | [`max_cumsum`, `interval_in_sec`]      | [`current_cumsum`, `interval_began`] |
+| `limit_monthly` | [`max_cumsum`, `interval_in_months`]      | [`current_cumsum`, `interval_began`] |
+
+All asserts apply to `int`, `string` and `object_ids`. There is no type converstion, incompatible type means assert failure. Any more sophisticated data types are not included yet.
+
+#### `any`
+Stateless assert, all argument types. Value of argument must be equal to one of values in the list
+
+#### `range`
+Stateless assert, only `int` type arguments. Value must be inbetween `min` and `max`, inclusive. One of the bounds can be empty to indicate unlimited.
```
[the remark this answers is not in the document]
sschiessl-bcp: EQ is handled by `any` asset. For NE I have added `none` assert. For the rest I have split up `range` as your approach allows more fine-grained control.

bsip-0040.md
```diff
@@ -0,0 +1,170 @@
...
+```
+List of possible asserts are:
+
+| function        | data           | state  |
+| ------------- |:-------------:| -----:|
+| `any`      | [`list`, `of`, `allowed`, `values`] | stateless |
+| `range`      | [`min`, `max`] | stateless |
+| `length`      | [`min`, `max`] | stateless |
+| `limit` | [`max_cumsum`, `interval_in_sec`]      | [`current_cumsum`, `interval_began`] |
+| `limit_monthly` | [`max_cumsum`, `interval_in_months`]      | [`current_cumsum`, `interval_began`] |
```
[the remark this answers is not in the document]
sschiessl-bcp: I have added a `contains_only` assert, which flips your logic around. A custom active authority is more likely to allow just one specific attribute rather than everything except one attribute.

bsip-0040.md
```diff
@@ -0,0 +1,170 @@
...
+
+# Specification
+
+All descriptions in this section are on a pseudo/prosa level and no recommendation how it can best be implemented or serialized. They are meant to facilitate the understanding.
+
+### Custom active permission and custom active authority
+
+A `custom_active_permission` looks like follows (in JSON/pseudo for clarification):
+```
+custom_active_permission = list of custom_active_authority items
```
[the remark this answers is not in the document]
sschiessl-bcp: I tried to make that clearer in Rational and Outline of handling incoming transactions

sschiessl-bcp:
> Using accounts in authority can be problematic due to increased complexity. For example, imagine this setup:
> 
> Account A
> 
>     custom authority 1
>         account B, can transfer 1 asset X to D
>     custom authority 2
>         account C, can transfer 2 asset X to D
> 
> Since a transaction can only be signed with keys, we need to convert keys to accounts when checking authority. If both account B and C have same key in their active authority, even with different weights, it would be complicated to check.
> 
> IMHO, we can have a "signed_by" field in extensions of each operation, to indicate who is signing the transaction, to simplify verification process as well as signing process.

I tried describing the process a bit in "Outline of handling incoming transactions". Checking the custom active authorities is meant as a preprocessing step before the current authority logic is applied.

In your example that means: If the incoming transaction requires active authority of A, the backend would start iterating the matched custom active authorities. That means the backend would first consider custom authority 1 and check if active authority of account B is present through signatures. If so active authority of A is granted. If not, consider custom authority 2 and check if active authority of C is present through signatures, If so, active authority of A is granted. If not, do nothing and proceed with normal authority check.

I have added this as an example

sschiessl-bcp:
> I want to narrow down this BSIP to a MVP (minimum viable product), once implemented, add more features with new BSIPs.

I have defined milestone for the BSIP, which is split in MVP and more sophisticated asserts.

abitmore:
> I have defined milestone for the BSIP, which is split in MVP and more sophisticated asserts.

IMHO, don't define milestones inside a BSIP. A BSIP IS a milestone or a part of a milestone/release. According to BSIP guidelines, a BSIP can't have a status like "partly-approved" or "partly-implemented" or "partly-installed".

bsip-0040.md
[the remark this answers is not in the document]
abitmore: As I've stated, this part is hard.

1. fields need to be standardized, otherwise the back end won't understand. That said, we need to define mapping rules for fields of each operation, primarily due to existence of nested structures, e.g. `price` or `asset_options`.
* `price` is a 2-level struct with 4 data fields: `{"base":{"amount":1,"asset_id":"1.3.1"},"quote":{"amount":2,"asset_id":"1.3.2"}}`. 
  * by the way, usually a `price` need to be validated as a whole, e.g. we need to assert `base.amount/quote.amount < x`.
* `asset_options` is a nested struct as well, it even has a `price` object inside.

  Actually, we need to define format of "argument" as well. When the field type is `static_variant`, it would be hard to define.

2. An operation have to contain all the needed fields. My example above is that the values in the incoming transaction need to be same as current values on chain. But when creating the asserts (rules) we don't know the exact value or don't care what the value is when processing a transaction, so can't put a value there.

  A "contains_only" assert makes some sense, but it's operation-specific:
* for operations like "account_update" or "asset_update" we need to check whether the values are same as current on-chain values;
* for operations like "call_order_update" or "account_whitelist" we need to check whether the values are zero or empty, because they're deltas;
* for operations like "transfer" perhaps we need to check whether the memo is empty if it's not in the list;
* for operations like "limit_order_create" it seems there is nothing to check so far.

Last, not like other functions which each applies to an argument/field, this "contains_only" function applies to the whole operation?

bsip-0040.md
```diff
@@ -77,18 +77,28 @@ List of possible asserts are:
...
-Stateless assert, all argument types. Value of argument must be equal to one of values in the list
+Stateless assert, all argument types. `Argument value` must be equal to one of values in the data list
+
+#### `none`
+Stateless assert, all argument types. `Argument value` must NOT be equal to any of the values in the list.
 
-#### `range`
-Stateless assert, only `int` type arguments. Value must be inbetween `min` and `max`, inclusive. One of the bounds can be empty to indicate unlimited.
+#### `lt, le, gt, ge`
+Stateless assert, applies to `int` and `string` type arguments. If argument is `string` type, use its length as `argument value`.
```
abitmore: If use strings' length here, we don't need the "length" function below.

bsip-0040.md
```diff
@@ -101,7 +111,10 @@ The assert that needs to pass is now `current_cumsum + incoming value <= max_cum
 #### `limit_monthly`
 Statefull assert, only `int` type arguments. Analogue to `limit`, but `interval_began` is initially set to `month(valid_from)` and set to `month(now)` on update, additionally the time assert is `month(now) >= interval_began + interval_in_months` (include logic for month overflow when year changes).
 
-### Simple Example
+#### `contains_only`
+Stateless assert, only for dictionary type objects like `options`. The `argument` of the operation is a dictionary type objects and is only allowed to contain the attributes in the data list.
```
abitmore: Most of "options" are not "dictionary type objects" if I understood your definition correctly. In the case of `asset_update_operation`, all attributes/fields inside the "options" need to be explicitly specified.

There ARE some dictionary type objects, for example `extensions` field of `account_create_operation`, `account_update_operation` and `call_order_update_operation`, all fields in the `extension` are `optional` so can be omitted.

sschiessl-bcp:
> IMHO, don't define milestones inside a BSIP. A BSIP IS a milestone or a part of a milestone/release. According to BSIP guidelines, a BSIP can't have a status like "partly-approved" or "partly-implemented" or "partly-installed".

Should I split this up into two separate BSIPS or would it be acceptable to create the milestones to be voted on.

Meaning: Create BSIP-40 (MVP) and BSIP-41 (sophisiticated asserts), or BSIP-40-1 and BSIP-40-2 to be voted on-chain?

bsip-0040.md
sschiessl-bcp: For both cases it is dictionary type right?

Diference is that in one case everything needs to be filled (and if unchanged with the onchain value) and the other case allows to only set the desired change.

I have removed `contains only` and added `attribute_assert` which has reduced logic to what it can do. The logic is not supposed to know which arguments are reflected on-chain an thus cant check if a value is unchanged. See `attribute_assert`.

bsip-0040.md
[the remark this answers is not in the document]
sschiessl-bcp: 1. The `lt` asserts (and similar) have now an explicit type conversion for `price`, others may be added as well. This might be unnecessary considering the new `attribute_assert`. I have also included a line that `argument` type can be added to `assert objects` if necessary. I want to avoid operation specific logic in the `custom active authority` check at all costs.

2. On-chain state checking should not be possible (would mean that we need to know which operation store what and where on chain). In general: If there is no assert defined, any value is allowed.

3. contains_only does no longer exist

sschiessl-bcp:
I have revised the BSIP again to include all comments.

sschiessl-bcp:
@clockworkgr added Economics section to account for your comment about increasing transaction fees.

sschiessl-bcp:
> I'm trying to make a proof of concept (which may need some time). I tend to not make it too smart, as describe here: bitshares/bitshares-core#1182. That said, probably will do operation specific logic.

Proof of concept would be awesome! Can you elaborate where you see the need for operation-specific logic with the current set of asserts? 

In general the user configures restrictions, but has no control to do anything "smart". How does this relate to your mentioned issue?

bsip-0040.md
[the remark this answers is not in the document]
xeroc: I discussed this with Stefan already. Nested objects are probably out of scope here.
In case of voting (as an example), we would propose a new operation that can only vote.
I suggest we do the same with core exchange rate ..

abitmore:
> Can you elaborate where you see the need for operation-specific logic with the current set of asserts?
>
>In general the user configures restrictions, but has no control to do anything "smart". How does this relate to your mentioned issue?

I guess it's easier to implement.

abitmore:
> Nested objects are probably out of scope here.

IMHO we must support nested objects. E.G.
* `asset` type is `{share_type amount, asset_id_type asset_id}`
  * `transfer_operation::amount`
  * `limit_order_create_operation::amount_to_sale`
  * `limit_order_create_operation::min_to_receive`

sschiessl-bcp:
> IMHO we must support nested objects. E.G.
> 
> - asset type is {share_type amount, asset_id_type asset_id}
>   - transfer_operation::amount
>   - limit_order_create_operation::amount_to_sale
>   - limit_order_create_operation::min_to_receive

Thank you for that input! With the `attribute_assert` the corresponding assert would look like:
```
custom_active_permission = {
    account_id = account A,
    authorities = [
        {
            valid_from = 2018-07-30T00:00:00Z,
            valid_to = 2018-07-30T23:23:59Z,
            operationid = transfer_operation,
            authority = {
                threshold: 1
                key_auth: [account B, 1]
                account_auth: []
            },
            asserts: [ cut_out ]
        }
    ]
}

cut_out = (amount, [
    {
         function: attribute_assert,
         data: [
                (amount, [ {function: lt, data: 10000} ]),
                (asset_id, [ {function: any, data: [asset X]} ])
        ]
    }
])
```

What this accomplishes: Account B can sign a transfer from Account A to Account C as log as it transfers a maximum of 10000 of asset X.

I understand the structure is complex now, but it is consistent and completely flexible. If that is TOO flexible, I would rather restrict the number of nestings and restrict the length of the involved lists via on-chain parameter and keep the generic logic list based.

bsip-0040.md
```diff
@@ -94,15 +113,17 @@ Stateless assert, all argument types. `Argument value` must be equal to one of v
 Stateless assert, all argument types. `Argument value` must NOT be equal to any of the values in the list.
 
 #### `lt, le, gt, ge`
-Stateless assert, applies to `int` and `string` type arguments. If argument is `string` type, use its length as `argument value`. 
+Stateless assert. Allows explicit type converstion:
+- `int` type: use as is
+- `string` type: user `length` as `argument value`
+- `price` struct `{base = {amount, asset_id}, quote = {amount, asset_id}}`: convert to float price  `argument value = base.amount/quote.amount`
```
abitmore: Then the value need to be a float point. We don't support float points in consensus code. Perhaps you'll say use string? I would say no, it's inefficient.

Perhaps we can support some simple math/expressions, or predefined function?

abitmore:
> `], OR)`

I don't think it's good to have `OR` here. It greatly increases complexity and causes misunderstandings.

Imagine this setup:
```
(assert 1, AND)
(assert 2, OR)
(assert 3, AND)
(assert 4, OR)
(assert 5, AND)
(assert 6, OR)
```

How would you explain it?

We already have `authorities = list of custom_active_authority items` and `any` operator to achieve `OR` logic.

bsip-0040.md
sschiessl-bcp: I have removed implicit `price` converstion. If user wants a price restriction he can use `attribute_assert`.

sschiessl-bcp:
> I don't think it's good to have OR here. It greatly increases complexity and causes misunderstandings.
> 
> Imagine this setup:
> 
>     (assert 1, AND)
>     (assert 2, OR)
>     (assert 3, AND)
>     (assert 4, OR)
>     (assert 5, AND)
>     (assert 6, OR)
> How would you explain it?
> 
> We already have authorities = list of custom_active_authority items and any operator to achieve OR logic.

The generic structure is: `restriction = (argument_identifier, list of asserts, logical_link)`. In this restriction, either all asserts are `AND` or `OR`. A mix like you stated above would NOT be possible. What the user could do is:
```
[ 
    (argument 1, [ assert 1, assert 3, assert 5], AND),
    (argument 2, [ assert 2, assert 4, assert 6], OR)`
]
```
Those two restriction would be ALWAYS combined with `AND` logic.

The reasoning is as follows: I want to make it quite expensive for the user to create a `custom_active_authority`. If `OR` logic is not provided, the user needs to create two authorities for the above example and might feel ripped off.

TheTaconator:
If this BSIP is implemented, the CLI will need to be enhanced similar to [CLI: Implement update_authorities](https://github.com/bitshares/bitshares-core/pull/1080)

sschiessl-bcp:
> Then if a user want to limit transfer to account A by 100 BTS, but account B by 200 BTS, she still has to use two authorities.

You are completely right. I have re-thought my approach and moved `OR` logic outside of Milestone 1 and created `logical_or` function for Milestone 3 (may never be needed or implemented). Milestone 1 only includes `any`, `none` and `lt, le, gt, ge` and `attribute_assert` restrictions.

xeroc:
> If this BSIP is implemented, the CLI will need to be enhanced similar to CLI: Implement update_authorities

I guess the best approach to tackle this is to offer *common* custom authority templates that can be used/installed by a simple cli-wallet command - think: Give trading permission to key `K` but only for market pair `A1:A2`.

More complex stuff can still be created manually thru tx builder or other libraries

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+hacking and phishing attempts this becomes even more important. The need for a more sophisticated
+account security preceeded the idea for a finer-grained control of account permissions.
+We propose to add an additional authority to the account, called Custom Active (Permission). The
+permission contains a list of operationid-to-authority mappings that each grant access to the respective 
+operation as if it were the active permission of the account. Additionally, the arguments of said operation 
+can be restricted.
+
+# Motivation 
+
+Any successfull hacking or phishing attempt on any of the web wallets that are powered by the
```
pmconrad: Typo: successful~~l~~

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+ - Trading Key: Only allows limit orders (arguments restricted to desired markets), update margin position and transfers (arguments restricted to certain accounts)
+ - Proposal Update Key: Approve proposals (2FA comes to mind)
+ - Faucet Key: Allow only to create accounts
+ - Withdrawal Key: Allow another account to transfer funds to himself
+ - Cold Storage Key: Only allow to move funds to the Hot Wallet
+ 
+The above list of named keys is nothing that is known to the backend as the backend should have an abstract implementation.
+ The UI could provide a button "Create Trading Key" that properly configures the respective custom active permission entry.
+
+This BSIP will be split into two parts that will be voted on serparately (see Milestones section). All of the above keys are possible with Milestone 1. Milestone 2 allows statefull restrictions (e.g. allow market orders up to some amount for every month), Milestone 3 gives finer control of how to combine restrictions. Milestone 2 and 3 will be put up for voting if Milestone 1 proves successful.
```
pmconrad: Typo: se~~r~~parately

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+
+A `custom active permission` contains a list of `custom active authorities` and looks like follows (in JSON-like/pseudo for clarification):
+```
+custom_active_permission = {
+    account_id,    // account that is assigned to this permission
+    authorities = list of custom_active_authority objects
+}
+custom_active_authority = {
+    valid_from,    // timestamp when this is active, defaults to now
+    valid_to,      // timestamp when this is invalid, defaults to 1 year
```
pmconrad: Both fields should be required to avoid ambiguities wrt defaults

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+    operation_id,   // operation id of the target operation,
+    authority,     // same as for the existing authortities (multisig with weighted accounts or keys),
```
pmconrad: Typo: author~~t~~ities

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+| `any`      | [`list`, `of`, `allowed`, `values`] | stateless |
+| `none`     | [`none`, `of`, `these`, `values`] | stateless |
+| `lt, le, gt, ge`      | `comparative` | stateless |
+| `limit` | [`max_cumsum`, `interval_in_sec`]      | [`current_cumsum`, `interval_began`] |
+| `limit_monthly` | [`max_cumsum`, `interval_in_months`]      | [`current_cumsum`, `interval_began`] |
+| `attribute_assert` | list of restrictions | stateless |
+| `logical_or`     | list of restrictions lists | stateless |
+
+Following cases must hold for a restriction:
+- if there is no value given (e.g. an optional argument, or nested value not given), the restrictions is passed (no change, no violation)
```
pmconrad: "no value" does not always mean "no change", see discussion in https://github.com/bitshares/bitshares-core/pull/838

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+- if the expected type of the argument does not match the given type (no implicit type conversion), the restriction is violated
```
pmconrad: Does it make sense to allow such restrictions at all?

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+the argument of the operation specified by `argument` of a restriction. 
+
+#### `any`
+Stateless assert, all argument types. `Argument value` must be equal to one of values in the data list
+
+#### `none`
+Stateless assert, all argument types. `Argument value` must NOT be equal to any of the values in the list.
+
+#### `lt, le, gt, ge`
+Stateless assert. Allows explicit type converstion:
```
pmconrad: Typo: convers~~t~~ion

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+- `string` type: use `length` of string as `argument value`
+
+The different asserts read as:
+- `lt`: `Argument value` must be less than `comparative`
+- `le`: `Argument value` must be less than or equal to `comparative`
+- `gt`: `Argument value` must be greater than `comparative`
+- `ge`: `Argument value` must be greater than or equal to `comparative`
+
+#### `limit`
+Statefull assert, only `int` type arguments. When the authority is created, `interval_began` is set to `valid_from` from its custom active authority and `max_cumsum` to `0`. Incoming operations are first tried to match all stateless asserts,
```
pmconrad: Typo: stateful~~l~~

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+- `gt`: `Argument value` must be greater than `comparative`
+- `ge`: `Argument value` must be greater than or equal to `comparative`
+
+#### `limit`
+Statefull assert, only `int` type arguments. When the authority is created, `interval_began` is set to `valid_from` from its custom active authority and `max_cumsum` to `0`. Incoming operations are first tried to match all stateless asserts, 
+and if all passes continue with statefull asserts. If `now > interval_began + interval_in_sec`, then set `max_cumsum = 0` and set `interval_began = now`.
+The assert that needs to pass is now `current_cumsum + incoming value <= max_cumsum`. If all `asserts` of this `custom_active_authority` pass, update `current_cumsum = current_cumsum + incoming value`.
+
+#### `limit_monthly`
+Statefull assert, only `int` type arguments. Analogue to `limit`, but `interval_began` is initially set to `month(valid_from)` and set to `month(now)` on update, additionally the time assert is `month(now) >= interval_began + interval_in_months` (include logic for month overflow when year changes).
```
pmconrad: Typo: stateful~~l~~

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+- if the account has a `custom active authority` match for every operation in the transaction that requires it, then grant the `active authority` of said account. If no match is found, treat as if no authority was given
+
+Note:
+- A `custom_active_authority` can only grant the `active authority` of the corresponding account, nothing more
+
+### Economics
+
+Adding a custom active authority means increased effort for the backend, and with a stateful one also the need for more storage. Proposed transaction fees:
+- `install_custom_active_authority`: Tied to the duration of the custom active authority. 
+Normal accounts can only create custom active authoritites with a duration of maximum 1 year. LTM can do any duration and also unlimited, but the transaction fee is capped at duration of 2 years.
```
pmconrad: Too complicated IMO. Better use fixed fee and duration limit (duration limit configurable through committee?).

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+- `update_custom_active_authority`: Base fee similar to `account_update` plus dynamic one for any duration changes
+- `delete_custom_active_authority`: Cheap similar to `limit_order_cancel`
+
+### Modification to the backend
+
+* Add a new index or extend the account object to store custom active permission are assigned to an account and contain a list of custom active authorities. Multiple custom active authority entries are possible for one operation
+* If the active authority of the account is updated, all custom active authorities need to be confirmed in the update. Every unconfirmed one is deleted otherwise
+* Provide operations: `install_custom_active_authority`, `update_custom_active_authority`, `delete_custom_active_authority` to allow changing the custom active permission (3 operation to allow custom transaction fees and avoid having to send the complete list of all authorities for every update)
```
pmconrad: Please clarify: is it possible to set custom active permissions in an account_update operation? Or must an account_update operation be bundled together with an install_custom_active_authority operation in order to preserve the existing custom active authorities?

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+* Operation-specific authorities (if present) must be evaluated in incoming transactions
+* Additional committee parameters may be needed to limit the extend of usage of this feature
+
+Notes: The implementation must not differentiate on which operation the custom active authority is applied, all operations are treated in same fashion
```
pmconrad: Are you sure? What about proposal_update? :-)

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+The required accounts (meaning required active authority) for the transaction is Account A. 
+Backend would start considering `custom active authority 1` and check if active authority of account B is present through signatures. 
+It is not, thus continue by checking if authority of `custom active authority 2` is present, which it is. 
+Acive authority of Account A is granted and normal authority checks are continued. 
+Since the required accounts is Account A, and the given accounts is also Account A through `custom active authority 2`, 
+the transaction is executed.
+
+# Milestones
+
+We propose do split the implmentation into two milestones. Each milestone will be voted on as a separate BSIP:
```
pmconrad: You list three milestones, not two

bsip-0040.md
sschiessl-bcp: done

bsip-0040.md
sschiessl-bcp: done

bsip-0040.md
sschiessl-bcp: done, no defaults

bsip-0040.md
sschiessl-bcp: done

bsip-0040.md
sschiessl-bcp: I clarified the text. If missing attribute means delete, that is fine. Custom authority still passed.

bsip-0040.md
xeroc: The idea is to keep active authory as is, with power to do everything except chain owner key.
Additionally to that, additional custom_operation_permission can be installed that are linked to account-id in a `M to N` manner. All custom_operation_permissions would get their own "id" to be able to edit or delete them.

bsip-0040.md
sschiessl-bcp: Is possible to check if the given restrictions match the argument types in the target operation?

With variable types in operation header this might not be possible and must be enforced on runtime, right?

bsip-0040.md
sschiessl-bcp: done

bsip-0040.md
sschiessl-bcp: done

bsip-0040.md
xeroc: The only thing a custom_operation_permission for proposal_update could do is to add the `active` permission of an account to a pending proposal (or remove it).

This could serve as a true 2FA/Multisig setup where the other party has even less power over the funds than in a regular multisig setup.

bsip-0040.md
sschiessl-bcp: done

bsip-0040.md
sschiessl-bcp: The idea is that the user can change active key without having to pay the install fee again (because that should be somewhat expensive).

Best would be to keep the three new operations for custom active permission.

Cases:
1. bundle: account_update and update_custom_active in one transaction, where update_custom only says: be valid for this new active authority
2. implicit with account_update. Include a new array with reference to all custom_active_authorities that should remain
3. dont allow it at all. account_update deletes all custom_active_authorities, always

Which technical possibilities do we have roughly @pmconrad ?

bsip-0040.md
sschiessl-bcp: Hm. Idea is to have the user really think about how long they want it, as it increases RAM and computational effort server side.

I also added now that expired custom_active_authorities should be deleted on maintenance.

bsip-0040.md
sschiessl-bcp: done

bsip-0040.md
```diff
@@ -0,0 +1,322 @@
...
+`data = [ [restriction 1, restriction 2], [restriction 3, restriction 4], ... ]`. If one of the restrictions sub-lists in data passes as whole, this restriction passes. In above miniexample that would mean if `restriction 1` and `restriction 2` pass, the whole `logical_or` restriction is considered to be passed as well.
+
+### Outline of handling incoming transactions
+ 
+When a signed transaction arrives and before the backend evaluates if all necessary authorities are present through the signatures, do the following:
+- iterate over required accounts and for each account, iterate over all operations within the transactions that require the active authority of this account
+- iterate the `custom_active_authorities` of said account
+- if a `custom_active_authority` is found that matches , remember that and stop iterating the authorities and continue until all operations are checked
+- if the account has a `custom active authority` match for every operation in the transaction that requires it, then grant the `active authority` of said account. If no match is found, treat as if no authority was given
```
pmconrad: I think this is misleading, or perhaps I just don't get it.
Current validation logic for reference: https://github.com/bitshares/bitshares-core/blob/2.0.180612/libraries/chain/protocol/transaction.cpp#L249
This is how I understand your proposal:
- iterate over operations:
-- get required authority accounts for operation
-- for each active authority account:
--- if account has matching custom active authority for operation, add this to list of required custom authorities,
--- otherwise add account to list of active authority accounts
- Verify active/owner/other accounts as in current logic
- on top of that, verify new list of custom active authorities

This means for example that if Alice has active authority key K and a custom active authority that allows key K' to sign a transfer of up to 1 BTS to Bob, then 
- Alice can transfer 2 BTS to Bob using K
- Alice can transfer 1 BTS to Charlie using K
- Alice can transfer 1 BTC to Charlie using K
- Alice can sell 1 BTS for whatever using K
- Alice can transfer 1 BTS to Bob, but she **must** use K' for that (K cannot sign)

Note that this new logic also applies to proposals. Not sure about the consequences of this.

bsip-0040.md
pmconrad: My question above aims at a contradiction between the previous two lines: account_update must allow specifying custom active authorities because otherwise they'll be deleted, OTOH you introduce new operations for this because you want dedicated fees.

bsip-0040.md
sschiessl-bcp: I see. I will try to rewrite the formulation.

Anyways, this is the idea:
 - the existing authority checking is not altered. If active authority is is given and also required, the operation is executed
 - additionally, if the active authority of a required account A is [not found](https://github.com/bitshares/bitshares-core/blob/2.0.180612/libraries/chain/protocol/transaction.cpp#L283) we check if there are signatures in sigs that have a valid custom_active_authority of account A. If such a custom_active_authority is found for every operation that requires active authority of account A, the active authority of account A is [added to sigs](https://github.com/bitshares/bitshares-core/blob/2.0.180612/libraries/chain/protocol/transaction.cpp#L271)

For your example:

- Alice can transfer 2 BTS to Bob using K
- Alice can transfer 1 BTS to Charlie using K
- Alice can transfer 1 BTC to Charlie using K
- Alice can sell 1 BTS for whatever using K
- Alice can transfer up to 1 BTS to Bob using K
- Anyone can transfer up to 1 BTS from Alice to Bob, as long as transaction is sigend with K'

bsip-0040.md
sschiessl-bcp: Having thought about it I propose:

A custom active authority has a status that can be enabled or disabled. This status can be set by the `update_custom_active_authority`. Additionally, if active authority is changed, all `update_custom_active_authority`s become disabled ALWAYS. User can enable it again with seperate call afterwards.

Reasoning:
 - avoid any misuse when accounts are given to people. Example: Account A is created by Person A and given to Person B. Person B changes all keys in order to fully own the account. But Person A has put a custom active authority, which Person B doesnt know and thus gains access
 - user should not be forced to pay the fee of `install_custom_active_authority` (expensive) again if he updates his active authority
 - it is acceptable to force the user to enable the custom active authorities after changing active keys

abitmore: Technically, we can add a field in `extensions` of `account_update_operation`, to specify a list of custom authorities to keep or remove.

bsip-0040.md
abitmore: How about `fee = flat_fee + basic_fee * duration`?
For update, `fee = flat_fee + max(0, delta(basic_fee * duration))`.
For removal, `fee = flat_fee`.

abitmore: >I also added now that expired custom_active_authorities should be deleted on maintenance.

I'd recommend disable it rather than removing it, since it's easier for users to "renew" an existing custom authority than to reinstall a new one. We can remove expired authorities after a period, for example a month or 7 days, to save RAM.

bsip-0040.md
abitmore: The argument types of restrictions should be check when installing the custom authority, at the validation step or even deserialization step. In an operation, data types are predefined, so they should match anyways.

bsip-0040.md
sschiessl-bcp: That is true, but intentional. Such an "approval key" that allows `proposal_update` could approve anything, but can't create proposals. Like Fabian mentioned that enables generic 2FA, where I would have such a "approval key" e.g. on my cellphone.

bsip-0040.md
sschiessl-bcp: I included that option now

bsip-0040.md
sschiessl-bcp: > I'd recommend disable it rather than removing it, since it's easier for users to "renew" an  existing custom authority than to reinstall a new one. We can remove expired authorities after a period, for example a month or 7 days, to save RAM.

Good case, included as well.

sschiessl-bcp: > How about fee = flat_fee + basic_fee * duration?
> For update, fee = flat_fee + max(0, delta(basic_fee * duration)).
> For removal, fee = flat_fee.
> 
> Calculate basic_fee according to complexity, e.g. size of authority, number of restrictions and etc.

Personally, I like it. Not fixated on how this is implemented. Have included it for now, if anyone has other input?

bsip-0040.md
sschiessl-bcp: I have included that in Restrictions sections now

sschiessl-bcp:
All reviews up till now have been addressed (meaning they are resolved or I gave follow-up questions)

charyury:
Hi guys, it's Yury from OL team. Most of the concerns we had looking at the original version of the BSIP were sorted out during the discussion, so at the moment we are pretty happy with the approach proposed, it implements most of the security features we would like to see and allows users to configure their permissions in very flexible way. We do believe "device tagging" can add even more protection, but we've discovered some flaws in our current vision of this functionality, we are re-designing it at the moment. One of the concerns is user experience, although this change gives big flexibility, it is not easy to be used by a "typical" user, so big attention shall be paid by UI team while implementing this in the front end. A special desktop or mobile app for key management mentioned in our WP probably could also help users to maintain their security settings easily and safely.
Anyhow, great BSIP, we are looking forward to see it implemented! Would be happy to participate in the implementation!

ryanRfox:
Great draft @sschiessl-bcp very comprehensive spec. 

@charyury is there anything in the OL research to add this spec? I'm still not completely sure I understand the "device tagging" spec described elsewhere. 

It would be valuable to design this BSIP as flexible as possible without unduly increasing the complexity. I like the milestone approach to continually deliver increased functionality.

bsip-0040.md
abitmore: Please be aware that everyone can create proposals for you. So if that key is compromised, effectively you can lose all funds in your account. You can still change permissions with your owner authority though.

sschiessl-bcp:
>  A special desktop or mobile app for key management mentioned in our WP probably could also help users to maintain their security settings easily and safely. 

This would make most sense in combination with offline signing API. An easy to use key management could also be implemented in the wallet directly! Still I like the initiative for the desktop app

> Anyhow, great BSIP, we are looking forward to see it implemented! Would be happy to participate in the implementation!

> Great draft @sschiessl-bcp very comprehensive spec.

Thanks, also thanks to all contributors and reviewers! I will ask some non-dev people for feedback as well now.

charyury:
> @charyury is there anything in the OL research to add this spec? I'm still not completely sure I understand the "device tagging" spec described elsewhere.
@ryanRfox , it looks like at the moment we need to implement PoC to make sure our vision of this feature is feasible. We are not ready at the moment to come up with a clear design. I will keep you posted.

bsip-0040.md
```diff
@@ -0,0 +1,345 @@
...
+    ]
+}
+```
+Exemplatory outcomes:
+- Transfer asset X with amount Y from account A to account B, signed with Key K: Accepted
+- Transfer asset X with amount Y from account B to account A, signed with Key K: Denied
+- Transfer asset X with amount Y from account A to account C, signed with Key K: Denied
+- Transfer asset X with amount Y from account A to account B, signed with active authority of B: Denied
+- Transfer asset X with amount Y from account A to account B, signed with active authority of A: Accepted
+- Create a proposal that includes operation 'transfer asset X with amount Y from account A to account B', signed with Key K: Denied
```
pmconrad: Please remove that line. Authority for the creation of a proposal does not depend on the operations contained therein. Creating a proposal with key K is possible if K has authority to pay fees.

bsip-0040.md
pmconrad: I think it is a bad idea to add active authority to the sigs because that could have unforeseeable side effects (for example, account A could be part of a multisig authority of account B - adding active authority A might authorize an operation on account B that would otherwise not have been granted).

bsip-0040.md
sschiessl-bcp: That was a moment of forgetting all my BitShares knowledge ... fixed now

bsip-0040.md
sschiessl-bcp: Custom active authority is designed to grant active authority. So if a key from a active authority is used on a matching operation it should grant the active authority. If it is a multi-sig is irrelevant.

**Example**

Account A has Account B and C as multisig. A has custom active authority that grants key K transfer priviliges. B has a custom active authority that grants key L transfer priviliges.

The transaction contains a transfer from A to D.
- signed by B and C: Accepted
- signed by L and C: Denied. The custom active authority of B does not match, only applies for transfers with B as sender 
- signed by K: Accepted, basically bypassing multisig. This is intended, as the multisig needs to approve the installment of the custom active authority in the first place

sschiessl-bcp: > - signed by L and C: Denied. The custom active authority of B does not match, only applies for transfers with B as sender

I actually made a mistake here. The requirement is not that B is the sender, but that B is a required account. In the backend, in above example. Is A the only required account in a multi-sig setup? Or are B and C also required accounts?

pmconrad: Good question. IIRC A is in required active, and the signature checking logic will recurse and check for presence of B and C authorities.

I had a slightly different scenario in mind: suppose Alice has a custom active authority with key K for transfers to Charlie. Bob has Alice as his active authority account.
A proposal containing two operations
1. transfer 1 BTS from Alice to Charlie
2. transfer 1000 BTS from Bob to owner of K
Proposal is approved by K. According to your pseudocode this means that active authority for Alice will be granted (operation 1 requires active authority of Alice, which is fulfilled by custom active authority). Presence of Alice's active authority is sufficient to authorize operation 2, because Alice is authoritative for Bob. Unintended consequence is that owner of K has control over Bob's account. That's why I think "granting active authority" is a bad idea. Instead, K should only be sufficient to authorize operation 1, and Alice's immediate active authority should be required for operation 2.

sschiessl-bcp: I see. That case should be excluded, as the custom active authority *should* only work directly for required_active accounts, no recursion.

In your example that means the active authority of A is granted *only* to match against required_active, but not to check if A may grant any other active authority as well.

sschiessl-bcp: I changed the wording from "granting" to "behave as if present for matched operation"

abitmore:
About fee. I'm thinking about a time-based fee mechanism. Say, when setting up a custom authority, the user deposit an amount of CORE token to a virtual pool. When the authority is in effect, as time goes by, token in the pool is deducted slowly (can be implemented as lazily deducting on expiration / update / removal). If the authority is disabled/removed at some time, refund unused fee. Thoughts?

bsip-0040.md
pmconrad: +1

pmconrad:
IMO this only complicates things, without significant benefit (possibly depending on actual fee amount and use-case). I'd leave it out of this proposal for now. Perhaps discuss in a more general fee-related BSIP.

IMO this BSIP is now good enough for merging and discussing in a broader audience.

bsip-0040.md
```diff
@@ -0,0 +1,379 @@
...
+```
+restriction = {
+   function,     // argument_identifier
+   argument,     // pointer to a dynamic value (argument of the operation, or attribute when nested)
+   data,         // data specific to the function
+}
+```
+is called a restriction, that can assert itself positively (passed) or negatively (violated). All restrictions are evaluated per default with `AND` logic to determine if the whole list has asserted positively. 
+
+List of possible restrictions are:
```
abitmore: I think it's better to expand this list a bit.
* Good to be able to use a function on the member, E.G. "length of" a string or a list, "size of" an object, and etc.
* There are `optional` fields in operations, which could have a special `is_valid` function, and perhaps a `is_not_valid` function. E.G. a custom authority can not use memo when transferring.
* There are vectors / lists, which don't fit `attribute_asset`, but could have `contains` / `does_not_contain` function .
* Good to have `equal` and `not_equal` functions to make restrictions simpler.

Examples:

| member | function | operator | argument | description |
| ------------- | ------------- | ----- | ---| --- |
| to | self | eq | "1.2.3" | "to" should only be "1.2.3" |
| memo | size | le | 100 | total memo size should be less than 100 bytes |
| active | self | not_valid | | "active" field should be not valid (when updating an account) |
| new_feed_producers | length | eq | 7 | feed producer list should have 7 items |
| new_feed_producers | self | contains | "1.2.4" | must have "1.2.4" as a feed producer |

Advanced expressions, E.G. for `limit_order_create_operation`:
```
{
      asset           fee;
      account_id_type seller;
      asset           amount_to_sell;
      asset           min_to_receive;
}
```
users may want to restrict `min_to_receive / amount_to_sell > X`, how to define it?

sschiessl-bcp:
> About fee. I'm thinking about a time-based fee mechanism. Say, when setting up a custom authority, the user deposit an amount of CORE token to a virtual pool. When the authority is in effect, as time goes by, token in the pool is deducted slowly (can be implemented as lazily deducting on expiration / update / removal). If the authority is disabled/removed at some time, refund unused fee. Thoughts?

The benefit that this approach brings is that a user does not need to think how long he actually wants to use it. The current logic forces the user to think about it and enforces to put it rather too short than too long. If a custom active was too short, extending it is easily doable. After the period, the custom active becomes disabled and can be enabled again with a new period, paying only for that new period.

I rewrote economics to clarify

bsip-0040.md
sschiessl-bcp: > - Good to be able to use a modifier on the member, E.G. "length of" a string or a list, "size of" an object, and etc.

I have added `object` and `list` to the implicit type conversions of `lt` et al. Also I added `eq` as operator that also does the implicit number conversion.

> - There are optional fields in operations, which could have a special is_valid function, and perhaps a is_not_valid function. E.G. a custom authority can not use memo when transferring.

Can you elaborate what they are checking? To check if a value is NOT set we can use `any` with empty list value. It will pass if not given and fail if any value is given.

> - There are vectors / lists, which don't fit attribute_asset, but could have contains / does_not_contain function .

Added

> - Good to have equal and not_equal functions to make restrictions simpler.

I don't think that is necessary, use `any` and `none` with one-item list.

abitmore: TBH I don't like implicit conversions which will increase complexity and are error prone. 

For example, a `memo` is
```
{
  public_key    from,
  public_key    to,
  int           nonce,
  vector<char>  message
}
```
and in transfers we actually use `optional<memo>`. It doesn't make sense to specially convert it to a number for comparing to another number. It doesn't make sense as well to use "any" or "none" since it's a struct but not a simple data type so unable to specify any value in the list. "`any` function with an empty list value" is hard to understand, IMHO it means "this field should be any of nothing" thus should always fail.

Since we have `price_per_kbyte` in fee schedule for many operations, it makes sense to have a `pack_size` function to calculate size of a member attribute after serialized. `pack_size` and `size` and `length` are different metrics. IMHO, better explicitly give the tools to users rather than converting implicitly.

abitmore: More thoughts:

for `optional` type, we can define a special value `null` (which is also used in jsonification) for users to use.

* ` A eq null` means A can only be null
* `A any [ null, 1 ]` means A can be null or 1
* `A ge null` is invalid
* `A gt 1` means A must be a number, and is valid (not null), and greater than 1

Actually this is still too complex, it would require much efforts to implement an array type which can contain a `null`.

sschiessl-bcp: > "any function with an empty list value" is hard to understand, IMHO it means "this field should be any of nothing" thus should always fail.

To be precise, it means "if set, this field should be any of nothing".

I will think about it and adjust the text. I will probably aim to add flexibility in the meaning of "this list of restrictions may be redefined to improve clarity,performance or to reduce complexity of implementation while maintaining the intended functionality".

bsip-0040.md
```diff
@@ -102,7 +102,8 @@ List of possible restrictions are:
 | ------------- |:-------------:| -----:|
 | `any`      | [`list`, `of`, `allowed`, `values`] | stateless |
 | `none`     | [`none`, `of`, `these`, `values`] | stateless |
-| `lt, le, gt, ge`      | `comparative` | stateless |
+| `lt, le, gt, ge, eq`      | `comparative` | stateless |
```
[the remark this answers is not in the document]
sschiessl-bcp: added neq

bsip-0040.md
```diff
@@ -171,9 +180,13 @@ Adding a custom active authority means increased effort for the backend, and wit
 - `install_custom_active_authority`: Normal accounts can only create custom active authoritites with a duration of maximum 1 year, LTM can do any duration. Two options come to mind:
   - A fixed high fee independent of authorities content
   - Tied to the duration and complexity of the custom active authority. Transaction fee is then `fee = flat_fee + basic_fee * duration` where `basic_fee` is calculated according to complexity (e.g. size of authority, number of restrictions and etc.). Fee is capped at 1 year for LTM.
-- `update_custom_active_authority`: Base fee similar to `account_update` plus dynamic one for any duration changes
+- `update_custom_active_authority`: Base fee similar to `account_update` plus dynamic one for any duration changes, no payback if duration in decreased.
 - `delete_custom_active_authority`: Cheap similar to `limit_order_cancel`
 
+This logic forces the user to think about the desired duration and enforces to put it rather too short than too long. 
+If a custom active is expired, or considered to be too short, extending it is easily doable.
```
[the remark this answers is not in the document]
sschiessl-bcp: Ah I see. It was meant like `duration` refers to the time that is still left, not the actual duration (i.e. `date_to - max(now, date_from)`).

In your example, the reference base duration when updating id `2018-08-20 - 2018-09-01`. So updating to `2018-08-15 - 2018-09-15`  means adding 14 days.

bsip-0040.md
```diff
@@ -130,12 +131,20 @@ Stateless assert, all argument types. `Argument value` must NOT be equal to any
...
 - `lt`: `Argument value` must be less than `comparative`
 - `le`: `Argument value` must be less than or equal to `comparative`
 - `gt`: `Argument value` must be greater than `comparative`
 - `ge`: `Argument value` must be greater than or equal to `comparative`
+- `eq`: `Argument value` must equal to `comparative`
+
+#### `contains, not_contains`
+Stateless assert, for `list` type arguments.
+- `contains`: The `argument value` must contain all items specified by `data`, but can contain more
+- `not_contains`: The `argument value` must NOT contain any of the items specified by `data`, but can contain others
```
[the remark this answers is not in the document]
sschiessl-bcp: I have changed the wording to `contains_all` and `contains_none`. Single object contains is a special case of list-based contains, I added that that can be added if beneficial for performance.

sschiessl-bcp:
Added Milestone 4: Number of executions which allows to limit number of executions

bsip-0040.md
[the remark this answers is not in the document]
abitmore: `now` is chain state which can't be used when calculating fee with current design principle.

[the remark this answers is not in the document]
sschiessl-bcp: I see, good point. The fee would then depend on execution time and not creation time.

Let's simply do it that all time period increase counts, whereas decrease is ignored.

In your example:
The change `2018-08-01` to `2018-08-15` decreases the period, ignore.
THe change `2018-09-01` to `2018-09-15` increases the period, add to fee.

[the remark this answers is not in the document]
abitmore: Unfortunately the value before change is in chain state so can't be used as well.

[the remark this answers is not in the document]
sschiessl-bcp: If fee logic can't contain chain state, then we need something like suppose_now, don't we?

Like either that logic with now, or the update must contain old_from and old_to which are used to calculate and must validate to.be equal to on-chain values, otherwise fail. 

Other ideas?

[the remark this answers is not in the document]
abitmore: Yes, that's the idea I mentioned above.

BTW, I made an implementation about creation fee:

```
billable fields in a custom authority = 
   { time_from, time_to, authority, a list of restrictions }

a restriction = { member, modifier, function, argument }

struct fee_parameters_type
{
   uint64_t basic_fee        ;
   uint32_t price_per_k_unit ;
};

total units of a custom authority =
   (time_to - time_from) in seconds * items in authority * total units of restrictions

units of a restriction = units of argument

total_fee_required = basic_fee + price_per_k_unit * total_units / 1000;
```

Here is a table for calculation units of different argument types:

Type | Unit | Description
--- | --- | ---
public_key_type | 4 | it uses around 32 bytes to store, 4 = 32 / 8
fc::sha256 | 4 | SHA256 hash, used in blind transfers
string | size() / 8 | length of string in 64-bit blocks, round up
list of strings | sum(units(string)) | total units of strings in list
other simple types | 1 | account id, timestamp, integer and etc
list of non-string types | list.size() * units(type) | fixed length
attribute restriction | units(restriction) | calculate recursively

Note:
* list of non-simple types is not allowed in argument.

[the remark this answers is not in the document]
abitmore: Then the fee of updating should take argument size and authority size into account, but not only duration difference. Ideally, 
```
update_fee = basic_fee + max( 0, total_new_fee - unused_old_fee )
           = basic_fee + max( 0, 
                new_fee_rate * new_duration * new_auth_size * new_argument_size
              - old_fee_rate * max(old_time_to-now, 0) * old_auth_size * old_argument_size )
```
Since all old data is in chain state, to calculate the difference, we need to put some "supposed" data in the operation, to be simpler, we can directly put the supposed delta in the operation and calculate in background.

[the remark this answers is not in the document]
sschiessl-bcp: > Note: I changed format of restriction a bit (mentioned in #86 (comment)),

You basically added a modifier field, which is used to extract the implicit type conversions into explicit ones. Which is totally fine and is covered by my note in the BSIP
"This list of restrictions may be redefined to improve clarity, performance or to reduce complexity of implementation while maintaining the intended functionality"

> Since all old data is in chain state, to calculate the difference, we need to put some "supposed" data in the operation, to be simpler, we can directly put the supposed delta in the operation and calculate in background.

What is best here for implementation I leave to you. One question I have: If the user overpays the transaction fee, will he (or can he) get the rest back?

sschiessl-bcp:
Anything we can learn for implementation here from EOS?

Saw this here:
https://steemit.com/eos/@ottomagic/how-to-create-a-separate-eos-permission-for-voting-using-cleos

abitmore:
@sschiessl-bcp according to the post, it seems EOS implemented per-action custom permission. I didn't look into the code, so don't know whether can set extra restrictions there. Actions in EOS are like operations in BitShares, the difference is in EOS actions belong to contracts and contracts can be upgraded with transactions, while in BitShares changing logic of operations requires hard-forks.

sschiessl-bcp:
> @sschiessl-bcp according to the post, it seems EOS implemented per-action custom permission. I didn't look into the code, so don't know whether can set extra restrictions there. Actions in EOS are like operations in BitShares, the difference is in EOS actions belong to contracts and contracts can be upgraded with transactions, while in BitShares changing logic of operations requires hard-forks.

Ok thanks, I never looked into EOS that much yet. It is similar, but has no restrictions. Just wanted to point it out in case there is something in their implementation we can utilize

sschiessl-bcp:
Merge this now, we will?

abitmore:
About implementation, EOS separated the feature into 2 actions (yes I guess it's actions of a system contract), one for adding custom authorities with a key (and perhaps can add advanced authority), the other for adding a restriction (permitted action) to specified authority.

sschiessl-bcp:
> About implementation, EOS separated the feature into 2 actions (yes I guess it's actions of a system contract), one for adding custom authorities with a key (and perhaps can add advanced authority), the other for adding a restriction (permitted action) to specified authority.

Ah, I see. The equivalent to our approach would be to but the same authority in multiple custom active authorities. Should we consider splitting it up as well?

bsip-0040.md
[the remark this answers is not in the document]
abitmore: > One question I have: If the user overpays the transaction fee, will he (or can he) get the rest back?

In BitShares fees can always be overpaid, and the system never refund overpaid fees except the "refund order creation fee when order is cancelled" feature.

abitmore:
> Should we consider splitting it up as well?

I'll think about it when implementing transaction verification logic, to see which one is better for us. Currently I'm trying the non-split way which is better for us due to the static billing (fees) mechanism. EOS can do dynamic billing (on CPU time and etc) on contract execution, so they can implement either way.

bsip-0040.md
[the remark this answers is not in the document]
abitmore: BTW I mentioned fee pooling mechanism earlier but Peter didn't like it at this early stage.

bsip-0040.md
abitmore: Not done.

abitmore:
> Merge this now, we will?

Will.

bsip-0040.md
sschiessl-bcp: Now done

bsip-0040.md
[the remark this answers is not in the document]
sschiessl-bcp: I saw that, but that takes it to a new level in terms of complexity. I think user can be expected to think before they create one.

PR #95, opened 2018-08-15
Initial draft of recurring monthly withdrawals

TheTaconator:
Draft of recurring monthly withdrawals in response [request for BSIP](https://github.com/bitshares/bitshares-core/issues/1208#issuecomment-408862157)

### DIFF
--- a/bsip-recurringmonthly.md
+++ b/bsip-recurringmonthly.md
@@ -7,7 +7,7 @@
     Discussion: https://github.com/bitshares/bitshares-core/issues/1208
 
 # Abstract
-Recurring withdrawals can currently be defined in BitShares core with respect to any periodic interval greater than the minimum block interval (currently 3 seconds). However those intervals are not intuitive for many use cases such as intervals defined in terms of Gregorian calendar months. This BitShares Improvement Proposal describes describes possible defintions for the periodic intervals and possible implementations such that they are compatible with the existing recurring withdrawal capability. The adoption of of this proposal might increase the goods and services that are accessible through BitShares by making it easier for merchants to accept payment on an intuitive montly basis.
+Recurring withdrawals can currently be defined in BitShares core with respect to any periodic interval greater than the minimum block interval (currently 3 seconds). However those intervals are not intuitive for many use cases such as intervals defined in terms of Gregorian calendar months. This BitShares Improvement Proposal describes possible defintions for the periodic intervals and possible implementations such that they are compatible with the existing recurring withdrawal capability. The adoption of of this proposal will likely increase the goods and services that are accessible through BitShares by making it easier for merchants to accept payment on an intuitive montly basis.
 
 # Motivation
 Recurring withdrawals can currently be defined with respect to any periodic interval greater than the minimum block interval (currently 3 seconds). However those intervals are not intuitive for many use cases such as a desire for a monthly interval. An approximation to a monthly interval might be 30 days; with that approximation and with a starting date-time of January 1 (in the Gregorian Calendar), the next period will begin on January 31, and the subsequent period will begin on March 1 or March 2 depending on whether it is a leap year.
@@ -47,11 +47,11 @@ As in the existing definition, it is possible for the last period to be a fracti
 
 ## Definition B: Elements of the Additional Recurring Definition
 
-The proposed recurring withdrawal periods are defined by three elements:
+The proposed recurring withdrawal periods are defined by five elements:
 
 - a starting date-time in UTC,
 - a starting day of month defined as a number between 1-28,
-- a starting hour of the day in UTC defined by a number between 1 and 23 (inclusive),
+- a starting hour of the day in UTC defined by a number between 0 and 23 (inclusive),
 - a periodic interval defined in terms of calendar units, and
 - an expiration date-time in UTC.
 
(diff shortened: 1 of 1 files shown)


bsip-recurringmonthly.md
```diff
@@ -0,0 +1,153 @@
+    BSIP: <BSIP number>
+    Title: Recurring Withdrawals Defined in Monthly Intervals
+    Authors: Taconator
+    Status: Draft
+    Type: Protocol
+    Created: 2018-08-15
+    Discussion: https://github.com/bitshares/bitshares-core/issues/1208
+
+# Abstract
+Recurring withdrawals can currently be defined in BitShares core with respect to any periodic interval greater than the minimum block interval (currently 3 seconds). However those intervals are not intuitive for many use cases such as intervals defined in terms of Gregorian calendar months. This BitShares Improvement Proposal describes describes possible defintions for the periodic intervals and possible implementations such that they are compatible with the existing recurring withdrawal capability. The adoption of of this proposal might increase the goods and services that are accessible through BitShares by making it easier for merchants to accept payment on an intuitive montly basis.
```
sschiessl-bcp: "might increase" -> "will increase"

monthly is simply what the old economy uses

bsip-recurringmonthly.md
```diff
@@ -0,0 +1,153 @@
...
+- an expiration date-time in UTC.
+
+The difference in definition is how the periodic interval is defined in terms of calendar units such as Gregorian calendar months. If the calendar unit is a Gregorian month, the starting day of the month will be the day of the month in UTC **derived from the starting-date-time**.
+
+### Caution: Start of Recurring Month
+
+The specified starting date-time will serve two purposes: (1) it will define the start of the recurring authorization, and (2) will implicitly define the starting day of the month **and** the starting time in UTC within that day.
+
+If the starting day of the month is derived to be on either Day 29, 30, or 31 of a month, caution must be taken because not all months in a Gregorian calendar contain these days. One simple manner of addressing this is to prohibit the selection of these days as the start of a recurring month. Another manner is to interpret such a day as beginning on the day requested **if** that Gregorian month contains that day, and otherwise to specify the prior day in the month that is present during that calendar month (for example, a start day of 31 will become 30 when the month is April).
```
sschiessl-bcp: I find that most other monthly periods define "end of the month" to be the 28th, which I think is sufficient.

bsip-recurringmonthly.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Caution: Possibility of Fractional Last Period
+
+As in the existing definition, it is possible for the last period to be a fraction of a periodic interval due to the open-ended definition of the expiration date-time.
```
sschiessl-bcp: My suggestion would be to neglect fractional last periods. Only the last fully included period counts (imagine taking out your full monthly salary because someone did not look at the hour of the day).

One could think of implementing a fractional withdrawal for fractional periods, but I believe that is unnecessary as its merely a question of proper UI.

bsip-recurringmonthly.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Definition B: Elements of the Additional Recurring Definition
+
+The proposed recurring withdrawal periods are defined by three elements:
+
+- a starting date-time in UTC,
+- a starting day of month defined as a number between 1-28,
+- a starting hour of the day in UTC defined by a number between 1 and 23 (inclusive),
```
sschiessl-bcp: Why not 0?

bsip-recurringmonthly.md
```diff
@@ -0,0 +1,153 @@
...
+- will need to be referenced in [database_api_impl::get_full_accounts](https://github.com/bitshares/bitshares-core/blob/58969c2a0307e32bbdee731d1f3f9a193b0d1b3f/libraries/app/database_api.cpp#L722-L727)
+- will need to be referenced from [database_api_impl::get_withdraw_permissions_by_giver](https://github.com/bitshares/bitshares-core/blob/58969c2a0307e32bbdee731d1f3f9a193b0d1b3f/libraries/app/database_api.cpp#L2116)
+- will need to be referenced from [database_api_impl::get_withdraw_permissions_by_recipient](https://github.com/bitshares/bitshares-core/blob/58969c2a0307e32bbdee731d1f3f9a193b0d1b3f/libraries/app/database_api.cpp#L2137)
+- will need to be referenced from [database::update_withdraw_permissions()]((https://github.com/bitshares/bitshares-core/blob/58969c2a0307e32bbdee731d1f3f9a193b0d1b3f/libraries/chain/db_update.cpp#L516))
+
+
+# Discussion
+
+A design question (in contrast to the existing definition) is whether an exact quantity of periods should be explicitly defined which will result in an implied expiration date. This is related to _"Specification of Intervals"_ and _"Specification of Period Start Date-Times"_.
```
sschiessl-bcp: If the last fractional period is neglected, this is merely a personal choice. As for usability, anyone using the CLI will be able to handle both, for the UI it is a question of proper UX.

bsip-recurringmonthly.md
```diff
@@ -0,0 +1,153 @@
...
+
+An implementation question is whether this secular period definitions should enhance the existing withdrawal objects and operations, or should define new objects and operations.  See _"Approach A"_ and _"Approach B"_.
+
+
+# Summary for Stakeholders
+
+The ability to define recurring withdrawals in commonly recognized intervals, such as "monthly", might increase the goods and services that are accessible through BitShares by making it easier for merchants to accept payment on an intuitive montly basis. The adoption of such a feature might also increase the usage of various BitAssets including bitCNY and bitUSD.
```
sschiessl-bcp: will :)

bsip-recurringmonthly.md
TheTaconator: FYI the existing implementation of recurring withdrawals allows fractional last periods.

Limitations in UI are reasonable. Should we maybe also restrict the Core from allowing fractional?

bsip-recurringmonthly.md
TheTaconator: Depending on how strictly the Core handles fractional periods, this specification in the UI can make a difference. My impression from your comments is that Core should be made strict to prevent fractional periods.

sschiessl-bcp:
It popped up in Telegram that including an auto-payment feature would be nice to have for withdrawal feature. Meaning: Automatically transfer the maximum amount for each period to the configured receiver.

What do you think?

pmconrad:
This would require additional processing on-chain. I don't see a significant advantage for users - recipients have to check for incoming funds anyway.

Can someone check how often withdrawals have been used yet? We should keep an eye on the cost/benefit ratio of these things.

TheTaconator:
@pmconrad One review of the blockchain operations through June 2018 indicates

- 17 withdraw permission create (Operation 26)
- 0 withdraw permission update (Operation 27)
- 0 withdraw permission claim (Operation 28)
- 0 withdraw permission delete (Operation 29)

xeroc:
> This would require additional processing on-chain. I don't see a significant advantage for users - recipients have to check for incoming funds anyway.

I agree with this. Since the software does not have an architecture to process "stuff" on a timed basis (except for maintenance interval), I would also prefer not to *automate* things too much.

One of the design principles, originally, was to have actions explicitly executed.

Hence, we could as well do similar to vesting balances in that balances accumulate according to some rule (e.g. unit-step function with monthly reset) and then let users *explicitly* claim.

pmconrad:
Is anybody still working on this?

Issue #97, opened 2018-08-17
BSIP41: Reduce MSSR of bitCNY from 1.1 to 1.05


bitcrab:
BSIP: 
Title: Reduce MSSR of bitCNY from 1.1 to 1.05
Authors: Jerry Liu <bitcrab@qq.com>
Status: Draft
Type: Protocol
Created: 2018-08-17
Discussion: https://bitsharestalk.org/index.php?topic=26928.0
Worker: no

### **Abstract**
We here propose to reduce the Maximum short squeeze ratio (MSSR) of bitCNY from 1.1 to 1.05, to lower the shorting incentive.

### **Motivation**
Bitshares is a financial system with leverage, here in the traders there are BTS longers and shorters, surely the traders are finding their chance to make profit in trading, if shorters can always find chance to make big profit by shorting BTS, that’s not a good news to BTS ecosystem, from the perspective of the whole BTS ecosystem consisted of DEX and smartcoin, we need to review the rules continually, while guaranteeing fairness and the risks always be released in time, we also need to try optimize the rules to encourage BTS longers, to provide enough incentive to make BTS ecosystem to grow up.

  
### **Rationale**
Currently there are 2 key smartcoin parameters which are maintained by witnesses, they are MCR(Maintenance collateral ratio ) and MSSR(Maximum short squeeze ratio) , now MCR=1.75 and MSSR=1.1.

When a debt position is margin called, the placed sell orders is at price feed price/MSSR, so MSSR determine the discount of collateral force selling the higher MSSR, the bigger the discount.

So high MSSR benefits the filling of margin call sell orders and help to release risks rapidly, however at the same time, high MSSR means buying margin call orders is with big discount, so it encourage BTS shorters to accumulate smartcoins and wait or even try to make margin call happen and get cheaper collateral. This does not do good to the ecosystem grow up.

We need to release risks rapidly while margin call happen, we also need to try to lower the shorting incentive, so we need a balance, what a MSSR is appropriate depend on the liquidity of the market and also some other factors.

bitCNY have a very good liquidity, BTS/bitCNY pair in DEX is the pair with highest BTS volume on the earth, after in 19th July the hard fork activated the target CR feature, the sell pressure comes from margin called orders decreased greatly,  we are sure that the for bitCNY black swan has very very little chance to happen, it’s the time to reduce MSSR to push bitCNY ecosysm move forward. 

### **Specification**
witnesses adjust the MSSR from 1.1 to 1.05 while publishing feed price.
[...]

PR #98, opened 2018-08-21
BSIP41: Reduce MSSR of bitCNY from 1.1 to 1.05

abitmore:
PR for #97.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,3 +48,4 @@ Number             | Title                                                    | 
 [38](bsip-0038.md) | Add target collateral ratio option to short positions                       | Abit More | Protocol | Installed
 [39](bsip-0039.md) | Automatically approve proposals by the proposer                             | Fabian Schuh | Protocol | Draft
 [40](bsip-0040.md) | Custom active permission                             | Stefan Schießl | Protocol | Draft
+[41](bsip-0041.md) | Reduce MSSR of bitCNY from 1.1 to 1.05               | Jerry Liu      | Protocol | Draft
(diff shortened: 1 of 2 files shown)


xeroc:
I would like the summary to include more of the concerns raised in the forums. Specifically:

* https://bitsharestalk.org/index.php?topic=26928.msg321086#msg321086

> MSSR has one specific purpose: to incentivize shorters to avoid being margin called (or rather to punish them if they are). Margin calls are bad for the health of our smartcoins and should be avoided. Shorters are expected to actively manage their positions so they are *not* margin called.
>
> It is clear, IMO, that bitCNY shorters are doing a very bad job here. As can be seen here (https://wallet.bitshares.org/#/asset/CNY ), most shorters make use of the recently introduced target_collateral_ratio to keep their collateral at the bare minimum. Obviously, they *prefer* being margin called. At the same time, they are *rewarded* from the settlement offset of 5% everytime someone settles a long position!
>
> The logical conclusion is that the MSSR must be increased so its original purpose is fulfilled.

* https://bitsharestalk.org/index.php?topic=26928.msg321102#msg321102

> I agree that the margin call as such is neither good nor bad. It is, however, an indication of a bad (read: dangerous) situation, i. e. short positions without sufficient collateral, which increases the likelyhood of a black swan.
> 
> The health of the system depends on shorters providing sufficient collateral. In bitCNY, shorters don't do that. They prefer being margin called. They even deliberately try to keep the collateral as low as possible, perhaps because only the least collateralized positions can profit from the 5% settlement offset.
>
> IMO this is an indication of a very unhealthy situation. The settlement offset is an incentive for shorters to keep their collateral as low as possible, and the current MSSR isn't sufficient incentive for providing enough collateral. Your suggestion increases the likelyhood of a black swan, as you correctly pointed out above.


Additionally, I would like one thing to be added and that is a **time limitation** for this expermient after which another vote will be required to keep this BSIP active. The purpose is to put the community at ease with this experiment and not have an open-ended experiment like the 5% offset we have on bitCNY.

PR #101, opened 2018-08-23
Update worker ID for BSIP 42

abitmore:


pmconrad:
Perhaps add which one is pro and which one is con.

Issue #103, opened 2018-08-28
New BSIP: Hashed Time-Locked Contract


ryanRfox:
Please provide comments within PR #104  

Current version of BSIP44 draft:
Revised: 29 SEP

BSIP: 0044\
Title: Hashed Time-Locked Contract\
Authors: Ryan R. Fox, John M. Jones, taconator\
Status: Draft\
Type: Protocol\
Created: 2018-08-22\
Discussion: https://github.com/bitshares/bsips/pull/104

# **Abstract**

This BSIP describes an implementation of a Hashed Time-Locked Contract (HTLC) operation.

# **Motivation**

The ability to securely hold tokenized assets within a hashed time-locked contract on the BitShares blockchain is a desirable feature that could be used by many persons, services, and businesses to mitigate risks between participants during asset transfer. HTLC implement conditional transfers, whereby a designated party (the "recipient") will reveal the preimage of a hash in order to execute the asset transfers from a second party (the "depositor"), else after time lock expiry "depositor" may retrieve their assets. No third-party escrow agent is required, rather the HTLC operation enforces conditions, evaluations and transfers through the BitShares consensus protocol.

# **Rational**

## **Elements of a Hashed Time-Locked Contract (HTLC)**

An HTLC is defined to have the following components:

* Parties to the HTLC

    * The depositor

    * The recipient

* Secured Asset

    * Symbol
    
    * Quantity

* Conditions

    * Hash lock

        * Preimage (the secret)

        * Preimage hash (hash of the preimage)

        * Preimage length

    * Time lock

        * Timeout threshold (expiry)

* Condition Evaluators

* Fees

  * Prepare operation fee
  
  * Prepare duration fee

  * Redeem operation fee

### **Parties**

Two parties must be defined within each HTLC: the `depositor` and the `recipient`. The `depositor` will secure their assets within the HTLC and designate the `recipient` to receive them. Note that a proposal transaction may be used for tasks such as multi-signature, but the end result at approval remains a single `depositor` and a single `recipient`.

### **Secured Asset**

An HTLC involves a conditional transfer of the defined `asset symbol` in the amount of `assets quantity` from the `depositor` to the `recipient`. The HTLC holds these designated `secured assets` from `depositor` on the blockchain and will continue to enforce the specified `conditions` until one is satisfied.

### **Conditions**

There are two competing conditions within an HTLC, the `hash lock` and the `time lock`.

[...]

Issue #49, opened 2018-01-12
Port ACCT from Scorum


pmconrad:
Apparently Scorum have implemented Atomic Cross-Chain Trading in their graphene chain (apparently a steem fork). We should evaluate and port this to BitShares.
https://github.com/Scorum/Scorum/wiki/How+to+do+atomicswap
https://bitsharestalk.org/index.php/topic,25789.0.html

AlexShkor:
Hello from Scorum team! 

Let me share with you the Pull Request with ACCT implementation, feel free to review and comment. We are happy to assist you with an integration of our implementation of Atomic Swap into Bitshares.
https://github.com/scorum/scorum/pull/41

Cheers,
Alex

oxarbitrage:
do we have resources working on this feature already ?

pmconrad:
Closing this since BSIP-44 has been implemented.

PR #104, opened 2018-08-28
BSIP44: Hashed Timelock Contract

ryanRfox:
See #103 

Current version of BSIP44 draft:
Revised: 29 SEP

BSIP: 0044\
Title: Hashed Time-Locked Contract\
Authors: Ryan R. Fox, John M. Jones, taconator\
Status: Draft\
Type: Protocol\
Created: 2018-08-22\
Discussion: https://github.com/bitshares/bsips/pull/104

# **Abstract**

This BSIP describes an implementation of a Hashed Time-Locked Contract (HTLC) operation.

# **Motivation**

The ability to securely hold tokenized assets within a hashed time-locked contract on the BitShares blockchain is a desirable feature that could be used by many persons, services, and businesses to mitigate risks between participants during asset transfer. HTLC implement conditional transfers, whereby a designated party (the "recipient") will reveal the preimage of a hash in order to execute the asset transfers from a second party (the "depositor"), else after time lock expiry "depositor" may retrieve their assets. No third-party escrow agent is required, rather the HTLC operation enforces conditions, evaluations and transfers through the BitShares consensus protocol.

# **Rational**

## **Elements of a Hashed Time-Locked Contract (HTLC)**

An HTLC is defined to have the following components:

* Parties to the HTLC

    * The depositor

    * The recipient

* Secured Asset

    * Symbol
    
    * Quantity

* Conditions

    * Hash lock

        * Preimage (the secret)

        * Preimage hash (hash of the preimage)

        * Preimage length

    * Time lock

        * Timeout threshold (expiry)

* Condition Evaluators

* Fees

  * Prepare operation fee
  
  * Prepare duration fee

  * Redeem operation fee

### **Parties**

Two parties must be defined within each HTLC: the `depositor` and the `recipient`. The `depositor` will secure their assets within the HTLC and designate the `recipient` to receive them. Note that a proposal transaction may be used for tasks such as multi-signature, but the end result at approval remains a single `depositor` and a single `recipient`.

### **Secured Asset**

An HTLC involves a conditional transfer of the defined `asset symbol` in the amount of `assets quantity` from the `depositor` to the `recipient`. The HTLC holds these designated `secured assets` from `depositor` on the blockchain and will continue to enforce the specified `conditions` until one is satisfied.

### **Conditions**

There are two competing conditions within an HTLC, the `hash lock` and the `time lock`.

The HTLC contains a `hash lock` condition, which comprise both the `preimage hash` and `preimage length`, barring the transfer of held `secured assets` unless satisfied. If a `preimage` of requisite `length` is provided to the HTLC which generates a hash matching the `preimage hash`, the `preimage` is then stored within the blockchain, and the `secured assets` are transferred to the `recipient`.

If a satisfactory `preimage` is not provided to the HTLC before the stipulated `time lock` expires, the `depositor` may request the return of `secured assets`. The HTLC will only evaluate transfer request from `depositor` and after `timeout threshold`, then return `secured assets` to `depositor`.

**Note:** we recommend the Committee the set maximum allowable `preimage length` to ensure unreasonably large submissions are rejected. 

### **Condition Evaluators**

The `preimage` can be thought of a secret key, that will eventually be shared with the `recipient`. This can be a word, a phrase, or even a random series of bytes. The `length` of the `preimage` must be specified within the HTLC at creation.

Upon presentation of a `preimage`, the HTLC `condition evaluator` validates:

1. That the `timeout threshold` has not yet occurred.

2. That the length of the `preimage` matches the specified `preimage length`. 

3. That the hash of the `preimage` calculates to the specified `preimage hash`. 

If all evaluations succeed, the `secured assets` are transferred to the `recipient`. If any evaluation fails, nothing happens; the HTLC remains ready to evaluate the next `preimage`.

### **Timing of Condition Evaluation**

The `timeout threshold` of the contract is defined by `depositor` within the HTLC at creation. It can be any time in the future and should allow enough time for `recipient` to review the HTLC and provide the `preimage`. Further, it should not be set too far into the future to mitigate against an unresponsive `recipient` impacting `depositor`, as their `secured assets` will be locked until `timeout threshold` expiry. The accuracy is based on when the `condition evaluator` runs, and should be considered accurate ± 15 seconds.

**Note:** we recommend the Committee set the maximum value for `timeout threshold` to limit the amount of time a contract may consume memory of validation nodes.

### **Early Termination of an HTLC**

To protect the `recipient`, early termination of an HTLC is not allowed by any party. Placing a `timeout threshold` far into the future is valid, up to the maximum defined by the Committee. User protection from locking up funds for an extremely long period could be provided by the UI used to create the HTLC.

### **Automatic Transfers Upon Expiry**

Upon expiry of the `timeout threshold`, the `secured assets` held within the HTLC will be queued for return to `depositor`. From this time, the HTLC will no longer evaluate the `hash lock`, preventing `recipient` from receiving the `secured assets`. No action is required by the `depositor` to receive their "locked" funds back from the contract after expiry. 

### **Fees**

We propose three (3) operations (see Specification) to implement the HTLC feature, each requiring distinct fees. All fees will be set and maintained by the Committee. 

The "prepare" operation will store in-memory data on validation nodes until redeemed or expiry. We recommend the `htlc_preparation_fee` be comprised of two (2) components: `GRAPHENE_HTLC_PREPARE_FEE` which is flat and `GRAPHENE_HTLC_DAILY_FEE` which is variable based on the number of days until `timeout threshold`.

The "redeem" operation frees most of the memory from the validation nodes and adds the `preimage` data into blockchain storage when the transaction is validated. We recommend the `htlc_redemption_fee` be comprised of two (2) components: `GRAPHEN_HTLC_REDEEM_FEE` which is may be quite low and `GRAPHENE_HTLC_KB_FEE` which is variable based on the total number of kilobytes of data committed to the blockchain.

The "extend expiry" operation will update the `timeout_threshold` to a future date, extending in-memory resources on validation nodes. We recommend the `htlc_extend_expiry_fee` be comprised of two (2) components: `GRAPHENE_HTLC_EXTEND_EXPIRY_FEE` which is flat and `GRAPHENE_HTLC_DAILY_FEE` which is variable based on the number of additional days added to extend the `timeout_threshold` of the contract.

## **Existing Escrow Proposals**

This section describes various escrow concepts that have been proposed either for BitShares or for other blockchains or services in terms of the elements that have been defined above. This is intended to provide some background and comparison to the concepts that follow.

### **BitShares Escrow**

A separate BSIP [cite] is currently being discussed that provides a more traditional escrow service. This involves parties, agents, and a more complex evaluation. HTLC shares some similarities, and could be considered a thin subset of BitShares Escrow.

The smaller, well-defined nature of HTLC provides a major advantage for applications that want to attempt tasks such as cross chain atomic swaps.

### **Scorum Atomic Swap**
[cite]

### **BitShares Multi-Signature Account**

One of the existing features of BitShares is the ability to have an account that requires multiples signatures by differently authorized parties [cite] and even hierarchical authorizations. Using this mechanism as a form of escrow is possible. But there are many limitations. More information on escrow and multi-signatures can be found in the BitShares Escrow BSIP [cite].

### **BitShares Proposals**

One of the existing features of BitShares is the ability to have a proposal that is recorded on the blockchain and awaits the authorization of the requisite parties (e.g. M-of-N signatures) to execute. However, the proposal does not "lock" any assets, so the transfer will fail if the sending account lacks sufficient funds during validation. If the required authorizations are not given by proposal expiry, then no transfer will occur. This feature also contains many limitations when compared to HTLC.

## **Possible Concepts to Implement**

The following will describe possible concepts that could be implemented within the BitShares protocol.

### **Set-Price Swap**

Two parties may agree on a swap of two distinct `secured assets` at a set price (defined exchange ratio), without using an exchange such as the BitShares DEX. This will require two (2) HTLC contracts containing the identical `preimage hash` within each to "link" them together and facilitate the execution of an "atomic swap" of these "locked" `secured assets` between the party's accounts resulting in a trustless value exchange.

#### **Business Approach**

Alice begins by generating a distinct `preimage` of her choosing, notes the `preimage length` and calculates the `preimage hash`. She retains the `preimage` in secret, then creates a new HTLC stipulating that the `depositor` account "alice" will transfer `quantity` "100" "bitUSD" `asset` into the `recipient` account "bob" if a `preimage` is presented matching the `preimage hash` before the `timelock threshold` of 10AM tomorrow. Upon consensus validation of the HTLC, the 100 bitUSD `secured assets` are transferred from Alice's `depositor` account into the HTLC where they remain locked by the `preimage hash` and `timelock threshold`. She then shares the resulting `contract identifier` with Bob.

Bob queries the blockchain for the `contract identifier` Alice provided. He examines to ensure it contains his desired `recipient` account, `asset` symbol, asset `quantity`, `preimage length`, and `timelock threshold`. Bob now creates his own HTLC that will deposit `quantity` "10,000" "BTS" `symbol` into the `recipient` account "alice" from `depositor` account "bob", if a `preimage` that generates the `preimage hash` Bob copied from Alice's HTLC before the `timelock threshold` of 5pm today. Upon consensus validation of Bob's HTLC, his 10,000 BTS `secured assets` are transferred from his `depositor` account and "locked" into the contract. He then shares the resulting `contract identifier` with Alice. Notice Bob specified a `timelock threshold` much shorter than Alice defined in her contract. This ensures Bob will have enough time to observe and use the `preimage` Alice will publish to the blockchain next. 

Alice now examines the HTLC Bob created, ensuring the `preimage hash` and `preimage length` both match the original values she used within her contract. She also verifies her desired `recipient` account "alice", the `quantity`, `symbol`, and the `timelock threshold` agree with her intentions. She now uses her `preimage` to "unlock" Bob's contract. Once consensus validation occurs, the HTLC will transfer the `secured assets` 10,000 BTS into her `recipient` account "alice". This reveals the `preimage` on the BitShares blockchain for Bob to use next. NOTE: She must do this before 5PM. Otherwise, Bob may (and should) reclaim the funds in the contract he created.

Bob can now observe the `preimage` Alice used to "unlock" his HTLC, and he will use it to "unlock" her HTLC to receive the 100 bitUSD `secured assets` into his `recipient` account "bob". NOTE: He must do this before 10AM tomorrow. Otherwise, Alice may (and should) reclaim the funds in the contract she created.

### **Cross-Chain Swap**

Similar to the set-price swap mentioned above, two parties may exchange tokens between distinct blockchains when both implement HTLC support. Bitcoin, Litecoin and many others support HTLC [cite].

#### **Business Approach**

Alice and Bob intend to swap BTC (bitcoin token) and BTS (BitShares token). This will require both parties to define both a BTC deposit address and BTS deposit account. These addresses/accounts will be exchanged between the parties.

Alice will initiate the first leg of the swap on the BitShares Network with her HTLC and Bob will follow up on the Bitcoin Network with his HTLC. Allice generates a distinct `preimage` of her choosing, notes the `preimage length` and calculates the `preimage hash`. She retains the `preimage` in secret, then creates a new HTLC stipulating that the `depositor` account "alice" will transfer `quantity` "10,000" "bitUSD" `asset` into the `recipient` account "bob" if a `preimage` is presented matching the `preimage hash` before the `timelock threshold` of 10AM tomorrow. Upon consensus validation of the HTLC on the BitShares Network, the 10,000 bitUSD `secured assets` are transferred from Alice's `depositor` account into the HTLC where they remain locked by the `preimage hash` and `timelock threshold`. She then shares the resulting `contract identifier` with Bob.

Bob queries the BitShares Network for the `contract identifier` Alice provided. He examines to ensure it contains his desired `recipient` account, `asset` symbol, asset `quantity`, `preimage length`, and `timelock threshold`. Bob now creates and funds his own HTLC on the Bitcoin Network that will spend the `UTXO` of this contract to the `recipient address` Alice provided during their setup phase, of `amount` 1 BTC if a `preimage` that generates the `preimage hash` Bob copied from Alice's HTLC before the `timelock threshold` of 5pm today. Upon consensus validation of Bob's HTLC on the Bitcoin Network, 1 BTC he controlled are spent into the contract and "locked". He then shares the resulting `contract identifier` with Alice. Notice Bob specified a `timelock threshold` much shorter than Alice defined in her contract. This ensures Bob will have enough time to observe and use the `preimage` Alice will publish to the blockchain next. 

Alice now examines the HTLC Bob created on the Bitcoin Network, ensuring the `preimage hash` and `preimage length` both match the original values she used within her contract. She also verifies her desired `recipient address`, `quantity`, and `timelock threshold` agree with her intentions. She now uses her `preimage` to "unlock" Bob's contract. Once consensus validation occurs on the Bitcoin Network, the HTLC will spend 1 BTC to Alice's `recipient address`. This reveals the `preimage` on the Bitcoin Network for Bob to use next. NOTE: She must do this before 5PM. Otherwise, Bob may (and should) reclaim the funds in the contract he created.

Bob has now observed the `preimage` Alice used to "unlock" his HTLC, and he will use it to "unlock" her HTLC to receive the 10,000 bitUSD `secured assets` into his `recipient` account "bob". NOTE: He must do this before 10AM tomorrow. Otherwise, Alice may (and should) reclaim the funds in the contract she created.

# **Specifications**

## **Objects**

```
      class htlc_object : public graphene::db::abstract_object<htlc_object> {
         public:
            static const uint8_t space_id = implementation_ids;
            static const uint8_t type_id  = impl_htlc_object_type;

            account_id_type depositor;
            account_id_type  recipient;
            asset amount;
            fc::time_point_sec expiration;
            asset pending_fee;
            vector<unsigned char> preimage_hash;
            uint16_t preimage_size;
            transaction_id_type preimage_tx_id;
      };
```

### **Prepare**

```
transaction_obj htlc_prepare(depositor, quantity, symbol, recipient, hash_algorithm, preimage_hash, preimage_length, timeout_threshold, htlc_preparation_fee)
  Validate: HTLC signed by requisite `authority` for `depositor` account 
  Validate: `depositor` account has requisite `quantity` of `symbol` asset for the `guarantee`
  Validate: `timeout_threshold` < now() + GRAPHENE_HTLC_MAXIMUM_DURRATION
  Calculate: `required_fee` = GRAPHENE_HTLC_OPERATION_FEE + GRAPHENE_HTLC_DAILY_FEE * count((`timeout_threshold` - now()), days)
  Validate: `depositor` account has requisite `quantity` of BTS for `required_fee`
  Validate: `recipient` account exists
  Validate: `preimage_length` does not exceed GRAPHENE_HTLC_MAXIMUM_PREIMAGE_LENGTH
  Validate: `preimage_hash` well formed
  Update: BTS balance of `depositor` based on `required_fee`)
  contract = new htlc_obj
  Set: `contract.depositor` = `depositor`
  Set: `contract.recipient` = `recipient`
  Set: `contract.hash_algorithm` = `hash_algorithm`
  Set: `contract.preimage_hash` = `preimage_hash`
  Set: `contract.preimage_length` = `preimage_length`
  Set: `contract.timeout_treshold` = `timeout_threshold`
  Transfer: from `depositor` account to `contract.quantity` of `contract.symbol`
  return results
```

### **Redeem**

```
transaction_obj htlc_redeem(fee_paying_account, id, preimage, htlc_redemption_fee)
  Validate: transaction signed by requisite `authority` for `fee_paying_account` // any account may attempt to redeem
  Get: get_htlc(id)
  Validate: `fee_paying_account` account has requisite `quantity` of BTS for `htlc_redeem_fee` and `htlc_kb_fee`
  Update: balance of `fee_paying_account` based on total fees
  // Evaluate: timelock
     if now() < `timeout_threshold` then return error // "timeout exceeded"
  // Evaluate: hashlock
     if length(preimage) != `id.preimage_length` then return error // "preimage length mismatch"
     Calculate: `preimage_hash` = hash(preimage)
     if `preimage_hash` != `id.preimage_hash` then return error // "invalid preimage submitted"
     Update: balance of `id.recipient` add asset `id.symbol` of quantity `id.quantity`
     Add: transaction to mempool 
     Set: `id.preimage_tx_id` = `transaction_id` 
     Cleanup: memory allocated to this htlc
     Virtual Operation: Update account history for `depositor` to reflect redemption as by default the above operation will only appear for `redeemer`

     return: results
```

### **Extend Expiry**

```
transaction_obj htlc_extend_expiry(depositor, id, timeout_threshold, htlc_extention_fee)
  Validate: 'depositor' = get_htlc(id).depositor 
  Validate: `timeout_threshold` < now() + GRAPHENE_HTLC_MAXIMUM_DURRATION
  Calculate: `required_fee` = GRAPHENE_HTLC_DAILY_FEE * count((`timeout_threshold` - now()), days)
  Validate: `depositor` account has requisite `quantity` of BTS for `required_fee`
  Update: BTS balance of `depositor` based on `required_fee`)
  Set: `contract.timeout_treshold` = `timeout_threshold`
  return results
```

### **At Expiry** (evaluated at each block commit)

```
  Get: get_htlc(id)
  Update: balance of `depositor` add asset `id.symbol` of quantity `id.quantity`
  Cleanup: memory allocated to this htlc
  Virtual Operation: Update account history for `depositor` to reflect expiry without redemption.
```

### **get_htlcs_for_account**

# **Discussion**

https://github.com/bitshares/bsips/pull/104

# **Summary for Tokenholders**

Hashed Timelock Contracts (HTLCs) enable conditional transfers, whereby distinct account holders may transfer tokens from one account (`sender`) to a second account (`receiver`) before a defined expiry (`timelock`), only if the `preimage` (a.k.a. password) is revealed (`hashlock`) on the blockchain. If the `hashlock` condition is not satisfied prior to the `timelock` the tokens are return to the `sender`.

A typical scenario involves “Alice” and “Bob” each having accounts on the BitShares Network and addresses on the Bitcoin Network willing to trade their tokens. Alice will begin by creating an HTLC on BitShares to transfer BTS tokens from account `alice` to account `bob` with conditions set for hash of preimage (`hashlock`) and contract expiry (`timelock`). Bob will review her HTLC, if acceptable he will create an HTLC on the Bitcoin Network to transfer BTC from `his address` to `her address` with conditions set to the same `hashlock` value and a `timelock` value approximately half that specified by Alice. Next, Alice will review Bob’s HTLC for correctness and if acceptable, will redeem the BTC therein by publishing her `preimage` to satisfy the `hashlock` prior to the `timelock` expiry. Finally, Bob will observe the revealed `preimage` and use it to redeem Alice’s HTLC on the BitShares Network resulting in the BTS transferring to his account. Alice and Bob successfully exchanged native BTS and BTC at their agreed to ratio without any intermediaries.

# **Copyright**

# **See Also**

A description of [Hashed Timelock Contracts](https://en.bitcoinwiki.org/wiki/Hashed_Timelock_Contracts)


ryanRfox:
Please discuss this BSIP below. I will incorporate suggested modifications into the Description above so that it reflects the current draft.

xeroc:
> The fee for each operation is based on the standards set for blocks, and is similar to costs of other items stored on-chain.

Fees are defined by committee and follow the committees desire. Please do not make any promises in the BSIP. That will make things more difficult for committee.

> BitShares Proposals

A major difference to proposals is also that they do not actually **lock** funds away. If at the time of the proposal expiration/execution, the funds are not their, the proposal will simply fail to execute.

Else, i like the BSIP.

bsip-0044.md
```diff
@@ -0,0 +1,173 @@
...
+
+### **Early Termination of an Escrow Contract**
+
+To protect the `recipient`, early termination of an HTLC is not allowed by any party. Placing a `timeout threshold` far into the future is valid. User protection from locking up funds for an extremely long period could be provided by the UI used to create the HTLC.
+
+### **Automatic Transfers Upon Expiry**
+
+Upon expiry of the `timeout threshold`, the `escrow assets` held within the HTLC will **not** automatically be returned to `depositor`. A separate transaction must be initiated by the `depositor` to request `escrow assets` be transferred to `depositor`. 
+
+Code _could_ be added to automate the return of funds. This could be part of block production or maintenance. However, this will cause extra load on those processes. Hence, it is felt that requiring action by the `depositor` is the best option.
```
pmconrad: I disagree. *Not* returning the funds automatically will unnecessarily lock up memory resources for an indefinite period of time. Note that, typically, returning of funds will be requested sooner or later, so not returning them immediately doesn't save resources in the long run. Quite the opposite, since processing the return request transaction (including signature validation) will consume a lot more resources than the return operation itself.

bsip-0044.md
```diff
@@ -0,0 +1,173 @@
...
+
+3. That the `hash of the preimage` matches the specified `preimage hash`. 
+
+If all evaluations succeed, the `escrow assets` are transferred to the `recipient`. If any evaluation fails, nothing happens; the HTLC remains ready to evaluate the next `preimage`.
+
+Upon presentation of a request from the `depositor` for return of `escrow assets`, the HTLC `condition evaluator` validates that the `timeout threshold` is now in the past. It then returns the `escrow assts` to `depositor`.
+
+### **Timing of Condition Evaluation**
+
+The `timeout threshold` of the contract is defined by `depositor` within the HTLC at creation. It can be any time in the future and should allow enough time for `recipient` to review the HTLC and provide the `preimage`. Further, it should not be set too far into the future to mitigate against an unresponsive `recipient` impacting `depositor`, as their `escrow assets` will be locked until `timeout threshold` expiry. The accuracy is based on when the `condition evaluator` runs, and should be considered accurate ± 15 seconds.
```
pmconrad: I'd suggest that a committee-configurable maximum period is implemented. The lockup potentially consumes memory resources until the timeout, therefore we should put a limit upon it.

bsip-0044.md
```diff
@@ -0,0 +1,173 @@
...
+
+An HTLC is defined to have the following components:
+
+* Parties to the HTLC
+
+    * The depositor
+
+    * The recipient
+
+* Escrow Asset
```
pmconrad: I'd avoid the term "Escrow" here, to avoid confusion with #76 .

bsip-0044.md
```diff
@@ -0,0 +1,173 @@
...
+
+### **Automatic Transfers Upon Expiry**
+
+Upon expiry of the `timeout threshold`, the `escrow assets` held within the HTLC will **not** automatically be returned to `depositor`. A separate transaction must be initiated by the `depositor` to request `escrow assets` be transferred to `depositor`. 
+
+Code _could_ be added to automate the return of funds. This could be part of block production or maintenance. However, this will cause extra load on those processes. Hence, it is felt that requiring action by the `depositor` is the best option.
+
+### **Fee**
+
+Creating and fulfillment are two operations that add data to the blockchain. The `fee` for each operation is based on the standards set for blocks, and is similar to costs of other items stored on-chain.
```
pmconrad: Please be more specific. There's a difference between creating and fulfilling.
* Creating consumes a more or less fixed amount of memory for a significant time period. The is comparable to creating limit orders (but requires more memory, depending on hash length). It should therefore come with a significant fixed fee greater than that of a limit order.
* Fulfilling frees up memory, but a long pre-image consumes space in a block and network bandwidth. It should therefore come with a small fixed fee, plus a per-kb fee.

bsip-0044.md
```diff
@@ -0,0 +1,173 @@
...
+
+One of the existing features of BitShares is the ability to have an account that requires multiples signatures by differently authorized parties [cite] and even hierarchical authorizations. Using this mechanism as a form of escrow is possible. But there are many limitations. More information on escrow and multi-signatures can be found in the BitShares Escrow BSIP.
+
+### **BitShares Proposals**
+
+One of the existing features of BitShares is the ability to have a proposal that is recorded on the blockchain and awaits the authorization of the requisite parties (e.g. M-of-N signatures). If the required authorizations are not given by the proposal expiry then no transfer will occur. This feature also contains many limitations when compared to HTLC.
+
+# **Specifications**
+
+## **Possible Concepts to Implement**
```
pmconrad: These concepts are possible applications of the proposed functionality and belong into the "Motivation" section.

The "Specification"  section should contain a detailed description of the proposed protocol changes, i. e. operations to be added, evaluation logic to be added, other supporting changes (like adding new fee types and extending global parameters accordingly).

bsip-0044.md
```diff
@@ -0,0 +1,173 @@
...
+Code _could_ be added to automate the return of funds. This could be part of block production or maintenance. However, this will cause extra load on those processes. Hence, it is felt that requiring action by the `depositor` is the best option.
+
+### **Fee**
+
+Creating and fulfillment are two operations that add data to the blockchain. The `fee` for each operation is based on the standards set for blocks, and is similar to costs of other items stored on-chain.
+
+## **Existing Escrow Proposals**
+
+This section describes various escrow concepts that have been proposed either for BitShares or for other blockchains or services in terms of the elements that have been defined above. This is intended to provide some background and comparison to the concepts that follow.
+
```
pmconrad: Please add comparison with ACCT as implemented in Scorum. #49

bsip-0044.md
ryanRfox: I agree that returning allocated memory is beneficial for validation nodes. I support automation when possible. It's unclear to me how/when the HTLC should be triggered for the automated return of assets to depositor. I feel it should not execute until LIB advances beyond contract expiry. Would appreciate some insights toward a spec on that. @jmjatlanta considered during maintenance.

bsip-0044.md
ryanRfox: I will agree with the value in allowing the Committee to set and maintain the maximum timeout threshold. Consideration should be made regarding the timeout to ensure the needs of those offering "layer 2" services can maintain "payment channels" of reasonable length. Perhaps the timeout on an existing contract could be "extended" by paying an additional fee in a update operation, rather than having to establish a new HTLC.

bsip-0044.md
ryanRfox: Consider the following replacements for "Escrow Asset":
- "Guarantee"
- "Collateral"
- "Deposit Asset"
- "Secured Asset"
- "Bonded Asset"
(I listed them in my preferred order)

bsip-0044.md
ryanRfox: Consider the following as a replacement:
### **Fees**

We propose two (2) operations (see Specification) to implement the HTLC feature, each requiring distinct fees. All fees will be set and maintained by the Committee. 

The "prepare" operation will store in-memory data on validation nodes until redeemed or expires. We recommend this operation have two (2) fees: `prepare_fee` which is flat and `prepare_durration_day_fee` which is calculated by based on the number of days until `timeout threshold`. 

The "redeem" operation frees the memory from the validation nodes and adds the `preimage` to blockchain data storage. We recommend this operation have two (2) fees: `redeem_fee` which is quite low and `redeem_preimage_kb_fee` which is based the amount of data committed to the blockchain.

ryanRfox: Also intend to add notes prior to the fee section:

**Note:** we recommend the Committee the set maximum allowable `preimage length` to ensure unreasonably large submissions are rejected. 

**Note:** we recommend the Committee set the maximum value for `timeout threshold` to limit the amount of time a contract may consume memory of validation nodes.

bsip-0044.md
ryanRfox: Got it. Moving Specifications after these examples and will add details thereto.

ryanRfox:
- Added some pseudocode within the Specifications section. I'll look to @jmjatlanta for some refinement
- Addressed the comments of @xeroc and @pmconrad regarding fees
- Stubbed out a heading for Scorum Atomic Swap; will finish in future revision 
- TODO: change the "Escrow Assets" term to something else. Please review my suggestions above in a reply to @pmconrad comment. Please reply there with thoughts and I will incorporate into a future revision.
- TODO: receive feedback about mardown formatting of code/variables.

bsip-0044.md
jmjatlanta: Thank you for you inputs. I was trying to prevent an extra step in maintenance, and was not looking at the bigger picture. I think the LIB argument is a good one, but would certainly like everyone else to weigh in.

bsip-0044.md
jmjatlanta: I had thought about a time extension during the initial stages of planning. I decided it was one more thing that would probably rarely be used and we would have to code for. I believe it is something we can add later. IMO: It would not be hard to do, I just do not know its value.

bsip-0044.md
jmjatlanta: Open to whatever the community thinks, but here are my opinions:
* I like escrow, as that is what it is. 
* Guarantee makes me think of some kind of written negotiated piece of paper that is distinct/abstract from the product being held. You can argue that we are doing exactly that, but I saying it just puts a different idea in my mind.
* Collateral is a good one. It would be my second choice.
* Asset is a word that is losing its meaning because of the many ways it is applied.

bsip-0044.md
jmjatlanta: Please clarify: If an automatic return of funds is implemented, when it is used, no fee is created. Correct? Having a fee on an "automatic" operation gets messy.

bsip-0044.md
abitmore: I don't think LIB makes sense here. If we release the fund immediately when got the input, the input and the release will be atomic. That said, if the chain reorganized and reverted the transaction with the input, the `release` action will be reverted automatically as well.

bsip-0044.md
pmconrad: I'd prefer one of "Secured/Bonded/Deposit Asset" (in that order).
Collateral has potential for confusion, just like Escrow.
Guarantee is too general IMO.

bsip-0044.md
pmconrad: Correct, automatic return of funds should be free of charge.
Fee per day should work too.

bsip-0044.md
```diff
@@ -86,41 +92,46 @@ Upon presentation of a request from the `depositor` for return of `escrow assets
...
 ### **Early Termination of an Escrow Contract**
 
-To protect the `recipient`, early termination of an HTLC is not allowed by any party. Placing a `timeout threshold` far into the future is valid. User protection from locking up funds for an extremely long period could be provided by the UI used to create the HTLC.
+To protect the `recipient`, early termination of an HTLC is not allowed by any party. Placing a `timeout threshold` far into the future is valid, up to the maximum defined by the Committee. User protection from locking up funds for an extremely long period could be provided by the UI used to create the HTLC.
 
 ### **Automatic Transfers Upon Expiry**
 
-Upon expiry of the `timeout threshold`, the `escrow assets` held within the HTLC will **not** automatically be returned to `depositor`. A separate transaction must be initiated by the `depositor` to request `escrow assets` be transferred to `depositor`. 
+Upon expiry of the `timeout threshold`, the `escrow assets` held within the HTLC will be queued for return to `depositor`. No action is required by the `depositor` to receive their "locked" funds back from the contract after expiry.
```
pmconrad: Please clarify: an expired HTLC that is "queued for return" can no longer be claimed by recipient, even if it hasn't been returned to depositor yet.

bsip-0044.md
```diff
@@ -156,9 +167,101 @@ Alice now examines the HTLC Bob created on the Bitcoin Network, ensuring the `pr
...
+  Set: `contract.timeout_treshold` = `timeout_threshold`
+  Transfer: from `depositor` account to `contract.quantity` of `contract.symbol`
+  return results
+```
+
+### **Redeem**
+
+```
+transaction_obj htlc_redeem(redeemer, id, preimage, htlc_redemption_fee)
+  Validate: transaction signed by requisite `authority` for `redeemer` // any account may attempt to redeem
```
pmconrad: Better call it fee_paying_account instead of redeemer

bsip-0044.md
```diff
@@ -156,9 +167,101 @@ Alice now examines the HTLC Bob created on the Bitcoin Network, ensuring the `pr
...
+```
+
+### **Extend Expiry**
+
+```
+transaction_obj htlc_extend_expiry(depositor, id, timeout_threshold, htlc_extention_fee)
+  Validate: transaction signed by requisite `authority` for `depositor` account 
+  Validate: `timeout_threshold` < now() + GRAPHENE_HTLC_MAXIMUM_DURRATION
+  Calculate: `required_fee` = GRAPHENE_HTLC_DAILY_FEE * count((`timeout_threshold` - now()), days)
+  Validate: `depositor` account has requisite `quantity` of BTS for `required_fee`
```
pmconrad: extend should only be allowed for original depositor.

bsip-0044.md
ryanRfox: Consider this insertion after the first sentence:
"From this time, the HTLC will no longer evaluate the `hash lock`, preventing `recipient` from receiving the `escrow assets`. "

bsip-0044.md
```diff
@@ -156,9 +167,101 @@ Alice now examines the HTLC Bob created on the Bitcoin Network, ensuring the `pr
...
+
+## **APIs**
+
+### **get_htlc**
+
+### **htlc_prepare**
+
+### **htlc_redeem**
+
+### **htlc_extend_expiry**
```
pmconrad: The latter three aren't API calls.

bsip-0044.md
```diff
@@ -156,9 +167,101 @@ Alice now examines the HTLC Bob created on the Bitcoin Network, ensuring the `pr
...
+transaction_obj htlc_extend_expiry(depositor, id, timeout_threshold, htlc_extention_fee)
+  Validate: transaction signed by requisite `authority` for `depositor` account 
+  Validate: `timeout_threshold` < now() + GRAPHENE_HTLC_MAXIMUM_DURRATION
+  Calculate: `required_fee` = GRAPHENE_HTLC_DAILY_FEE * count((`timeout_threshold` - now()), days)
+  Validate: `depositor` account has requisite `quantity` of BTS for `required_fee`
+  Update: BTS balance of `depositor` based on `required_fee`)
+  Set: `contract.timeout_treshold` = `timeout_threshold`
+  return results
+```
+
```
pmconrad: Please add a virtual operation for redemption containing (depositor, recipient, amount, symbol, preimage) and one for expiration containing (depositor, amount, symbol). (Perhaps include HTLC object id in both? Unsure.)

bsip-0044.md
```diff
@@ -86,41 +92,46 @@ Upon presentation of a request from the `depositor` for return of `escrow assets
...
 
-### **Fee**
+The "prepare" operation will store in-memory data on validation nodes until redeemed or expiry. We recommend the `htlc_preparation_fee` be comprised of two (2) components: `GRAPHENE_HTLC_PREPARE_FEE` which is flat and `GRAPHENE_HTLC_DAILY_FEE` which is variable based on the number of days until `timeout threshold`.
 
-Creating and fulfillment are two operations that add data to the blockchain. The `fee` for each operation is based on the standards set for blocks, and is similar to costs of other items stored on-chain.
+The "redeem" operation frees most of the memory from the validation nodes and adds the `preimage` data into blockchain storage when the transaction is validated. We recommend the `htlc_redemption_fee` be comprised of two (2) components: `GRAPHEN_HTLC_REDEEM_FEE` which is may be quite low and `GRAPHENE_HTLC_KB_FEE` which is variable based on the total number of kilobytes of data committed to the blockchain.
+
+The "extend expiry" operation will update the `timeout_threshold` to a future date. We recommend the `htlc_extend_expiry_fee` be calculated based on the number of additional days at the current `GRAPHENE_HTLC_DAILY_FEE` rate.
```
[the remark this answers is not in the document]
ryanRfox: Fair enough, I can add it in, as it's more consistent.

One thing I contemplated within the specification was the calculation to extend being from now() to GRAPHEN_HTLC_MAXIMUM_DURATION, so in effect the `depositor` would be paying for at minimum one additional day they previous paid for, thus compensating the blockchain for the operation fee. The specification could also be modified to collect fees for only difference between the new expiry and current contract expiry.

bsip-0044.md
```diff
@@ -156,9 +167,101 @@ Alice now examines the HTLC Bob created on the Bitcoin Network, ensuring the `pr
...
+
+  return contract
+```
+
+## **Operations**
+
+### **Prepare**
+
+```
+transaction_obj htlc_prepare(depositor, quantity, symbol, recipient, preimage_hash, preimage_length, timeout_threshold, htlc_preparation_fee)
```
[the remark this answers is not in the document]
ryanRfox: Good thought. I'll add `hash_algorithm // enum` to the HTLC fields.

bsip-0044.md
ryanRfox: done.

bsip-0044.md
ryanRfox: Respectfully, I'm not sure how this should be modified. My intention is to word it such that the `depositor` account may require threshold `authority`. I agree, only `depositor` account may extend the `timeout_threshold` (although the `authority` thereof may be different at validation).

bsip-0044.md
ryanRfox: Dang. I'll need some guidance with this section, as I'm not sure what to include. Perhaps @jmjatlanta can jump in.

Are there any example BSIPs I should review for these details?

bsip-0044.md
ryanRfox: @jmjatlanta you gotta help with virtual ops, as that is beyond my ability to spec out properly.

bsip-0044.md
ryanRfox: I'm going to change it to "Secured Asset" in the next draft. Let's see how it reads throughout.

bsip-0044.md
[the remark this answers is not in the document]
jmjatlanta: I too like that idea, if the list is limited to highly trustworthy algorithms. I believe providing popular options may help with some other blockchains. I believe Bitcoin can handle SHA256 and RIPEMD.

bsip-0044.md
```diff
@@ -0,0 +1,280 @@
...
+  preimage_length   // int
+  preimage_tx_id    // reference to transaction object containing the preimage used for htlc_redeem
+  timeout_threshold // datetime
+
+  // Constructor
+  Set: id = next available
+  Initialize other fields = NULL
+
+  return contract
+```
```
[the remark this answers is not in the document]
ryanRfox: Thanks for the proper code. I can add this in. 

As for the reference rather than the preimage data within the HTLC object: my desire is to store this (rather larger) data object only once in the blockchain. I’m not a real developer, so will defer to you guys on implementation. My thought is: the Redeem TX must contain the valid preimage to be validated, so that is when and only when it needs to be committed to the blockchain. Can have the operation update the object to reference the TX. Can that work?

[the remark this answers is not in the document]
jmjatlanta: I agree that the crux of the issue is storage on the blockchain. The potential size of the preimage means this should be carefully considered. The operation that creates the HTLC tells us the preimage size, but not the preimage. A later transaction (that may never happen) would contain the actual preimage.

The immutability of the chain means that a redeem operation cannot go back and "fill in" the preimage there. Hence, the redeem operation must store the preimage on its own. But I believe with account history, an htlc object should be modified to include the preimage. That allows for clients to request 1 object that has the complete picture. If it only included the transaction id of the preimage, a second call would be necessary.

Please shoot holes in my theory.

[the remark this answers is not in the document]
jmjatlanta: Thinking it over again with the help of @pmconrad I can see an advantage to only including a reference to the transaction id. Yes, it could potentially cause the client to do 2 lookups, but that also means that the preimage does not have to be stored in the live database, as it is on the chain.

Once again, please weigh in with your thoughts.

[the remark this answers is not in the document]
ryanRfox: I'm going to offer my thoughts, with the caveat these concepts are beyond my authoritative knowing, so happy to be corrected: It is my understanding that the `recipient` will likely "subscribe to the HTLC" created by `depositor` after they evaluate the contents match their desires. This will allow the `recipient` to receive notification once the `depositor` supplies the `preimage` in the Redeem operation. I don't know enough about which bits get stored where to shoot holes.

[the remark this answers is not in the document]
jmjatlanta: In the few other HTLC implementations I have read about, the depositor shares the preimage out-of-band. That gives the redeemer the option to execute the contract or not. If he decides to not "sign on the dotted line", he simply does not use the preimage, and the contract times out.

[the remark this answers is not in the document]
ryanRfox: This is my understanding of HTLC process flow:

Given: d depositor, r receiver, p preimage, h hash(p), l length(p), c HTLC, A Alice, B Bob
A: generate p, h, l
A: create A_c as A_d >>> B_r, h, length(p)
A: send A_c identifier to B
B: evaluate A_c
B: create B_c as B_d >>> A_r, h, length(p)
A: receive B_c identifier
A: evaluate B_c
A: redeem B_c with p
B: observe p and redeem A_c with p

A will reveal p only when redeem B_c. If B becomes aware of p, B may allow B_c to expire and redeem A_c prior to expiry without penalty.

bsip-0044.md
TheTaconator: @pmconrad Trying to understand your point. Is it that the latter three should be classified as CLI commands instead of API calls?

How are they different from `get_htlc`?

I am assuming that we are discussing CLI commands

bsip-0044.md
jmjatlanta: If I understand virtual operations correctly, we need a non-virtual operation for redemption, as the preimage must be stored on the chain. Expiration does not affect the chain (only account balances), so should be a virtual operation. Please correct me if I am wrong.

bsip-0044.md
pmconrad: Validate: depositor = get_htlc(id).depositor

bsip-0044.md
pmconrad: Yes, `get_htlc` should be a node API call, because otherwise clients have no means to query for existing htlc contracts.
The other three would be nice to have in cli_wallet (and/or wallet_api), but they aren't node API calls.

bsip-0044.md
pmconrad: You have already correctly specified a non-virtual operation for redemption. Per default, that operation would only turn up in the account history of the redeemer, not in that of the depositor. IMO the depositor should see the outcome of the htlc in his history, so we either need an extra virtual op for this, or have to make sure that the non-virtual redemption op turns up in the depositor's history (which is probably simpler).
Expiration should appear in the depositor's history and is not otherwise represented, therefore a virtual op is required.

pmconrad:
> Not sure if you want pseudocode or something more real.

Depends (as usual...).
When it comes to **protocol changes**, we should strive to be specific and unambiguous.
**Implementation details**, however, should be left to the developers. Being overly specific here could complicate things down the line, or restrict possible optimizations.

bsip-0044.md
ryanRfox: I've attempted to reflect the virtual operations @pmconrad mentioned with the addition of `html_at_expiry` and inline within `htlc_redeem`. However, I'll need some guidance on both logic and format. 

I also revised the API Section to reflect interfaces for `cli_wallet` and `witness_node`. Again, not sure if that is the intended format. Also, what code should be included there?

bsip-0044.md
ryanRfox: I've attempted to reflect the virtual operations @pmconrad mentioned with the addition of `html_at_expiry` and inline within `htlc_redeem`. However, I'll need some guidance on both logic and format.

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+
+1. That the `timeout threshold` has not yet occurred.
+
+2. That the length of the `preimage` matches the specified `preimage length`. 
+
+3. That the hash of the `preimage` calculates to the specified `preimage hash`. 
+
+If all evaluations succeed, the `secured assets` are transferred to the `recipient`. If any evaluation fails, nothing happens; the HTLC remains ready to evaluate the next `preimage`.
+
+Upon presentation of a request from the `depositor` for return of `secured assets`, the HTLC `condition evaluator` validates that the `timeout threshold` is now in the past. It then returns the `secured assets` to `depositor`.
```
xeroc: Why would we need this if the secured assets are transfered to depositor automatically (according to line 105)

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+
+### **Prepare**
+
+```
+transaction_obj htlc_prepare(depositor, quantity, symbol, recipient, hash_algorithm, preimage_hash, preimage_length, timeout_threshold, htlc_preparation_fee)
+  Validate: HTLC signed by requisite `authority` for `depositor` account 
+  Validate: `depositor` account has requisite `quantity` of `symbol` asset for the `guarantee`
+  Validate: `timeout_threshold` < now() + GRAPHENE_HTLC_MAXIMUM_DURRATION
+  Calculate: `required_fee` = GRAPHENE_HTLC_OPERATION_FEE + GRAPHENE_HTLC_DAILY_FEE * count((`timeout_threshold` - now()), days)
+  Validate: `depositor` account has requisite `quantity` of BTS for `required_fee`
```
xeroc: fee checking happens outside of validation framework - iirc. Also fee can be paid in non BTS tokens

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+  Validate: `recipient` account exists
+  Validate: `preimage_length` does not exceed GRAPHENE_HTLC_MAXIMUM_PREIMAGE_LENGTH
+  Validate: `preimage_hash` well formed
+  Update: BTS balance of `depositor` based on `required_fee`)
```
xeroc: Fees are dealt with separately, no existing operation does anything with fees (except for fee calculation!)

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+  contract = new htlc_obj
+  Set: `contract.depositor` = `depositor`
+  Set: `contract.recipient` = `recipient`
+  Set: `contract.hash_algorithm` = `hash_algorithm`
+  Set: `contract.preimage_hash` = `preimage_hash`
+  Set: `contract.preimage_length` = `preimage_length`
+  Set: `contract.timeout_treshold` = `timeout_threshold`
```
xeroc: Provided that operations should be easy to understand outside of context, I would request to put a PointInTime type for timeout/expiration instead of a timeout in seconds

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+  Validate: 'depositor' = get_htlc(id).depositor 
+  Validate: `timeout_threshold` < now() + GRAPHENE_HTLC_MAXIMUM_DURRATION
+  Calculate: `required_fee` = GRAPHENE_HTLC_DAILY_FEE * count((`timeout_threshold` - now()), days)
+  Validate: `depositor` account has requisite `quantity` of BTS for `required_fee`
+  Update: BTS balance of `depositor` based on `required_fee`)
+  Set: `contract.timeout_treshold` = `timeout_threshold`
+  return results
+```
+
+### **At Expiry**
```
xeroc: Do I assume right that "At Expiry" would be determined in the maintenance block?

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+
+### **htlc_prepare**
+
+### **htlc_redeem**
+
+### **htlc_extend_expiry**
+
+## **witness_node APIs**
+
+### **get_htlc**
```
xeroc: This call is crucial as you need to identify an HTLC in order to redeem it.
I guess, a proper naming for the call should be `get_htlcs_for_account`

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+
+The HTLC contains a `hash lock` condition, which comprise both the `preimage hash` and `preimage length`, barring the transfer of held `secured assets` unless satisfied. If a `preimage` of requisite `length` is provided to the HTLC which generates a hash matching the `preimage hash`, the `preimage` is then stored within the blockchain, and the `secured assets` are transferred to the `recipient`.
+
+If a satisfactory `preimage` is not provided to the HTLC before the stipulated `time lock` expires, the `depositor` may request the return of `secured assets`. The HTLC will only evaluate transfer request from `depositor` and after `timeout threshold`, then return `secured assets` to `depositor`.
+
+**Note:** we recommend the Committee the set maximum allowable `preimage length` to ensure unreasonably large submissions are rejected. 
+
+### **Condition Evaluators**
+
+The `preimage` can be thought of a secret key, that will eventually be shared with the `recipient`. This can be a word, a phrase, or even a random series of bytes. The `length` of the `preimage` must be specified within the HTLC at creation.
```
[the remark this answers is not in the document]
jmjatlanta: The length is the length in bytes. So multi-byte characters, even binary data can be used as a preimage.

bsip-0044.md
jmjatlanta: It was originally planned to have the depositor ask for a refund. That has since changed to be automatic. This paragraph should have been deleted.

bsip-0044.md
jmjatlanta: Yes. The timeout here will be a point in time. Of course, the implementation of a point in time is the number of seconds since an epoch. So it is actually both a quantity of seconds and a point in time. :-)

bsip-0044.md
jmjatlanta: I have assumed that we would use the database ID as the identifier. This has advantages and disadvantages. The advantage is it is guaranteed to be system-wide unique after LIB. The disadvantage is that it could change before LIB, possibly misleading users. I would like to hear everyone's input on how best to implement HTLC identifiers.

bsip-0044.md
jmjatlanta: I assume so. We should clarify this here. Are maintenance intervals frequent enough to provide timely refunds?

xeroc: That's why I thought claiming back funds after timeout would be a nice idea.
One original design decision was to have as much as possible happen *explicitly* (transfer) and as little as possible *implicitly* (think, fill order). That said, I personally would prefer a manual step, but could understand that - given the expiration being explicit - we could as well use maintenance interval. However, maintenance intervals can be changed by committee, and thus inconvenient.

bsip-0044.md
xeroc: To get the object behind an idea, you can simply use `get_objects([id])`. But I may want to find (among many), the HTLC object that links to my account (as depositor or recipient).

bsip-0044.md
pmconrad: IMO this can be handled in _apply_block just like expired proposals. No need to delay until maintenance.

bsip-0044.md
pmconrad: > The disadvantage is that it could change before then, possibly misleading user

This applies to other operations as well. Solution is to wait until LIB is past creation.
(For assets + accounts there is also assert_operation, but IMO it doesn't make sense here.)

bsip-0044.md
pmconrad: Technically correct, but where fee checking happens is an implementation detail. IMO the remark doesn't hurt here.

bsip-0044.md
pmconrad: See above.

bsip-0044.md
jmjatlanta: So I think I am hearing that the ability to retrieve all HTLCs for a particular account would be a helpful API addition. I believe that can be added without difficulty.

The harder question is if the database id is okay to use. Here is my thinking: Immediately after creation, the database id is available, but has a chance to be incorrect. I believe that the more popular HTLCs become, the greater the chance of incorrectness. Retrieving a more accurate id is a function of time. Eventually, the id is guaranteed.

Those with more experience than I would be better at guestimating when it is "mostly safe" to believe an ID. Does an ID become "mostly safe" within a reasonable period of time? I realize block times can be adjusted. I am asking for an "in practice" time period.

If we still want to use the database id, we could minimize the chance of a user looking at an incorrect contract by (as xeroc suggests) including the account name as part of the search criteria. Personally, I would rather have Bob explain to Alice that the id she shared is not valid, rather than Bob see a contract that has nothing to do with what he and Alice are negotiating.

We can walk down other schemes for ids. GUID, user assigned, hashes of a signed timestamp. But I still hope that the database id will work if it is not too inconvenient for the user.

ryanRfox: Is retrieving the id of the correct/safe contract an implementation detail for the client side rather than the server API side? Should we push the logic about what is valid to the client side so the end user can feel confident in the data used in their application? There may be value to querying for an contract prior to LIB.

pmconrad: The real danger in publishing a claim operation is not that the ID might be wrong, the real danger is that before LIB you cannot be sure that the chain doesn't switch to a fork where the HTLC does not exists. That means you have revealed your secret, but there is nothing you can claim with it. So Alice will have to wait until Bob's HTLC is beyond LIB anyway.
Database IDs are used as a reference throughout the protocol. This is part of the reason why our transactions are so lean and our chain so fast. I see no reason to deviate from the rule in this case.

jmjatlanta: My interpretation of @pmconrad's comment: Go with the database ID. Make sure that all developers and users of this feature are clear that a contract is not "set in stone" until the block that contains it is protected by an irreversible block.

My additional input: With the above stipulation, wallet and b2b developers that want to use this feature can add business rules and user prompting to prevent poor decisions by users / businesses.

Is that an ideal solution? I do not think so. But I doubt that an ideal solution could be worked out in a consensus system. Other chains suffer the same problem. The vast majority (all?) are worse off than Bitshares (require more time before something is considered "irreversible").

bsip-0044.md
jmjatlanta: Deleted paragraph

bsip-0044.md
jmjatlanta: Added more detail (evaluated at each block)

bsip-0044.md
jmjatlanta: Added method get_htlcs_for_account

ryanRfox:
With John's latest changes, I feel the draft is complete. I've updated the Issue and PR Descriptions to the latest text.

ryanRfox:
@xeroc Do you feel the questions you raised during review have been answered by @pmconrad and @jmjatlanta? I'm keen to get this merged and put before the community, so want to understand what is left to refine. Thanks

pmconrad:
@jmjatlanta @ryanRfox you have been overtaken by BSIP 45, please rebase to resolve conflict.

@xeroc last chance for a review. :-)

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+
+Alice now examines the HTLC Bob created, ensuring the `preimage hash` and `preimage length` both match the original values she used within her contract. She also verifies her desired `recipient` account "alice", the `quantity`, `symbol`, and the `timelock threshold` agree with her intentions. She now uses her `preimage` to "unlock" Bob's contract. Once consensus validation occurs, the HTLC will transfer the `secured assets` 10,000 BTS into her `recipient` account "alice". This reveals the `preimage` on the BitShares blockchain for Bob to use next. NOTE: She must do this before 5PM. Otherwise, Bob may (and should) reclaim the funds in the contract he created.
+
+Bob can now observe the `preimage` Alice used to "unlock" his HTLC, and he will use it to "unlock" her HTLC to receive the 100 bitUSD `secured assets` into his `recipient` account "bob". NOTE: He must do this before 10AM tomorrow. Otherwise, Alice may (and should) reclaim the funds in the contract she created.
+
```
abitmore: One risk I can see is that after Alice broadcast her preimage, Bob would know it immediately and can redeem funds in Alice's contract immediately. However, it's not guaranteed that Alice can redeem Bob's contract successfully, since it's possible that her transaction fails to get into a block E.G. due to chain reorganization (forking) or transaction censoring.

Actually IMHO proposals are more suitable for this use case, because it can guarantee that 2 transfers both succeed or both fail (see https://github.com/bitshares/bitshares-ui/issues/866).

When it's a cross-chain swap, the situation would be even more complicated. To be safe, time-lock of both contracts should be long enough.

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+A typical scenario involves “Alice” and “Bob” each having accounts on the BitShares Network and addresses on the Bitcoin Network willing to trade their tokens. Alice will begin by creating an HTLC on BitShares to transfer BTS tokens from account `alice` to account `bob` with conditions set for hash of preimage (`hashlock`) and contract expiry (`timelock`). Bob will review her HTLC, if acceptable he will create an HTLC on the Bitcoin Network to transfer BTC from `his address` to `her address` with conditions set to the same `hashlock` value and a `timelock` value approximately half that specified by Alice. Next, Alice will review Bob’s HTLC for correctness and if acceptable, will redeem the BTC therein by publishing her `preimage` to satisfy the `hashlock` prior to the `timelock` expiry. Finally, Bob will observe the revealed `preimage` and use it to redeem Alice’s HTLC on the BitShares Network resulting in the BTS transferring to his account. Alice and Bob successfully exchanged native BTS and BTC at their agreed to ratio without any intermediaries.
```
abitmore: IHMO this paragraph is redundant, and is too long as a summary.

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+
+An HTLC involves a conditional transfer of the defined `asset symbol` in the amount of `assets quantity` from the `depositor` to the `recipient`. The HTLC holds these designated `secured assets` from `depositor` on the blockchain and will continue to enforce the specified `conditions` until one is satisfied.
+
+### **Conditions**
+
+There are two competing conditions within an HTLC, the `hash lock` and the `time lock`.
+
+The HTLC contains a `hash lock` condition, which comprise both the `preimage hash` and `preimage length`, barring the transfer of held `secured assets` unless satisfied. If a `preimage` of requisite `length` is provided to the HTLC which generates a hash matching the `preimage hash`, the `preimage` is then stored within the blockchain, and the `secured assets` are transferred to the `recipient`.
+
+If a satisfactory `preimage` is not provided to the HTLC before the stipulated `time lock` expires, the `depositor` may request the return of `secured assets`. The HTLC will only evaluate transfer request from `depositor` and after `timeout threshold`, then return `secured assets` to `depositor`.
```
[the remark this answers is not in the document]
jmjatlanta: I thought I removed all of those references. I will fix it.

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+     Calculate: `preimage_hash` = hash(preimage)
+     if `preimage_hash` != `id.preimage_hash` then return error // "invalid preimage submitted"
+     Update: balance of `id.recipient` add asset `id.symbol` of quantity `id.quantity`
+     Add: transaction to mempool 
+     Set: `id.preimage_tx_id` = `transaction_id` 
+     Cleanup: memory allocated to this htlc
+     Virtual Operation: Update account history for `depositor` to reflect redemption as by default the above operation will only appear for `redeemer`
+
+     return: results
+```
```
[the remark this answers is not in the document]
jmjatlanta: Good question. My initial reaction is that the transaction should obey any white/blacklisting at time of contract creation, and ignore later changes. Without that, adjustment of such a mechanism could prevent the funds from even going back to the original depositor.

I welcome others thoughts.

bsip-0044.md
jmjatlanta: Agreed that timelock lengths need to be long. The safest way to handle a cross-chain swap from the user's standpoint is:

1) Alice figures out how long it will take for such a transaction to be irreversible on either chain, and use the longest as the baseline.

2) Alice makes her contract timeout the ``baseline*2`` plus a little extra.

3) Bob makes his contract timeout ``baseline``, plus a little extra

Now Alice should wait until Bob's contract is irreversible before exposing her key. This means it is very important that both parties review the contracts carefully, and know the implications of the parameters in each.

There is the risk that Alice exposes her key, and her transaction is rejected. She would have to get a subsequent transaction pushed through before the timeout of that contract. Hence, waiting until the last moment is not a good idea.

Now, there is always the risk of transaction censoring, but I don't think that is only an HTLC problem.

bsip-0044.md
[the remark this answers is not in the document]
abitmore: Makes sense.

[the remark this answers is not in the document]
pmconrad: agree

bsip-0044.md
```diff
@@ -0,0 +1,292 @@
...
+            static const uint8_t space_id = implementation_ids;
+            static const uint8_t type_id  = impl_htlc_object_type;
+
+            account_id_type depositor;
+            account_id_type  recipient;
+            asset amount;
+            fc::time_point_sec expiration;
+            asset pending_fee;
+            vector<unsigned char> preimage_hash;
+            uint16_t preimage_size;
```
pmconrad: Please add hash_algorithm

CryptoKong:
I wonder if a feature could be incorporated along side this where BTS is used as collateral to allow instant action of the contract. So users dont have to wait for say 6 confirmations but the collateral is held until then. If the transaction is somehow reversed or whatever, collateral is instead sent to the other party. This may be of interest to some users and potentially create more demand for BTS.

syalon:
preimage_size is very unfriendly to users who create contracts later, and there are potential security risks.

Let's analyze the elements that the latter creates.
* preimage_hash (value and type): If the user enters the wrong, it will cause the funds to freeze, waiting for the contract to expire.
* preimage_size: If the person who created the contract first is wrong. The person who created the contract later is correct. If the user forgets to check, the latter will fail to redeem the contract. Loss of funds. And many chains do not have this parameter, and users may ignore the importance.
* expiry: The latter is set to be greater than the validity period of the former creation, there may be financial risks.

Therefore, in order to quickly verify that the hash is valid and add a size judgment, the latter user has a great risk.

@ryanRfox @abitmore

abitmore:
@syalon please open a new issue.

PR #105, opened 2018-08-29
Completion of this BSIP from a formal standpoint 

sschiessl-bcp:
and inclusion of a clear accepted/rejected state


sschiessl-bcp:
#106 has been opened as a second PR. It adds the notion to start with bitCNY.

PR #106, opened 2018-08-29
add discussion and voting relevant content in bsip-0042.md

bitcrab:
add discussion and voting relevant content

### DIFF
--- a/bsip-0042.md
+++ b/bsip-0042.md
@@ -156,11 +156,13 @@ by bugs.
 
 # Discussion
 
-[To be added]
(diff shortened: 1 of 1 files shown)


xeroc:
I think I like #105 better.

ryanRfox:
It is my understand that this PR should be closed in favor of #105 after reading: https://bitsharestalk.org/index.php?topic=26966.msg321675#msg321675

May I request @bitcrab to close this if that is his intention to mitigate confusion?  Thanks much.

Issue #102, opened 2018-08-24
BSIP-0043:  Market fee sharing


OpenLedgerApp:
```
BSIP:
Title: Market fee sharing
Authors: OpenLedger
Status:
Type:
Created: 2018-08-23
Discussion: https://github.com/bitshares/bsips/issues/102
Worker:
```
 

# Abstract
 

When creating a new asset, the asset issuer is the only beneficiary of the market fees in the current implementation. And the only way to affect the asset community growth is the market fee percentage. For example, one can decrease the market fee and it will result in somewhat larger number of trades with this asset. In this way the asset issuer might get a bigger profit during to increasing the trade volume with this asset.
 

However, there might be another opportunity to promote the asset and stimulate the trading - use native Bitshares referral program. At this time unfortunately an assets owner is not able to share market fees with registrars and referrers to stimulate them to market the asset trading, so we suggest to add this possibility. Furthermore, enabling this feature for MPAs (e.g. bitCNY or bitUSD) can provide additional bounty for Bitshares registrars and referrals which can lead to more traders joining to the ecosystem.
The workflow may be as following:

An asset issuer defines the market_fee_reward_percent - what percentage of the market fee he wants to share with the registrar in asset options. Registrar defines reward_percent for the referrer for each user(using the already existing BitShares mechanism).
Market fee reward is accumulated on the user account. The user decides when they want to claim the market fee reward and move it to their active balance. There is another operation created asset_claim_reward_operation. Each user pays network fee to call this operation.
 

# Motivation
 

To make promoting BitShares and bringing new users much more attractive to registrars and referrers by sharing UIAs and MPAs market fees with them.
 

# Rationale
When 'fill_order_operation' executing at the moment of market fee calculation there will be calculate the reward for the registrar used the parameter market_fee_reward_percent. Then this market_fee_reward_percent of the market fee is split between the registrar and referrer according to the referrer_rewards_percentage, which is set up during the new account registration by registrar (please see the parameters for register_account).
 

# Specifications
 

## market_fee_reward_percent asset option
Percent of market fee that will be paid to buyer's registrar. Set by UIA issuer.
 

## market_sharing_whitelist asset option
 An optional list of accounts (configurable by the UIA Issuer) could provide increased control over who is eligible to get market rewards.
 

## New database and wallet api methods
**get_account_reward(account, asset_symbol)** - Returns information about the reward amount for specific asset.
**list_account_rewards(account)** - Returns information about the reward amount in various assets
[...]

PR #107, opened 2018-08-29
bsip#43 Market fee sharing

OpenLedgerDev:
PR for #102 

According to the comments: https://github.com/bitshares/bsips/issues/102#issuecomment-416565443

> What is the reasoning to have a seperate "asset_claim_reward_operation", and not funnel it through existing referral reward mechnism?

A separate account_reward_object and asset_claim_reward_operation are needed to track profits from "market fee sharing".
Vesting_balance_object receives fee from all transactions.

> The market_sharing_whitelist allows discrimination. Is there a viable use-case for this?

The market_sharing_whitelist is an additional functionality. Asset Issuer may not use this list.

> Agreed to the points above, making this a general dividend for holding the asset would be misleading the intent

We have read a few times and we're not sure we understand points 2 and 3 correctly. Could you elaborate a bit more please?

### DIFF
--- a/bsip-0043.md
+++ b/bsip-0043.md
@@ -1,0 +1,96 @@
+    BSIP: 0043
+    Title: Market fee sharing
+    Authors: OpenLedgerApp <https://github.com/OpenLedgerApp>
+    Status: Draft
+    Type: Protocol
+    Created: 2018-08-23
+    Discussion: https://github.com/bitshares/bsips/issues/102
+    Worker:
+
+
+# Abstract
+
+When creating a new asset, the asset issuer is the only beneficiary of the market fees in the current implementation. And the only way to affect the asset community growth is the market fee percentage. For example, one can decrease the market fee and it will result in somewhat larger number of trades with this asset. In this way the asset issuer might get a bigger profit during to increasing the trade volume with this asset.
+
+However, there might be another opportunity to promote the asset and stimulate the trading - use native Bitshares referral program. At this time unfortunately an assets owner is not able to share market fees with registrars and referrers to stimulate them to market the asset trading, so we suggest to add this possibility. Furthermore, enabling this feature for MPAs (e.g. bitCNY or bitUSD) can provide additional bounty for Bitshares registrars and referrals which can lead to more traders joining to the ecosystem.
+The workflow may be as following:
+
+An asset issuer defines the market_fee_reward_percent - what percentage of the market fee he wants to share with the registrar in asset options. Registrar defines reward_percent for the referrer for each user(using the already existing BitShares mechanism).
+Market fee reward is accumulated on the user account. The user decides when they want to claim the market fee reward and move it to their active balance. There is another operation created asset_claim_reward_operation. Each user pays network fee to call this operation.
+
+# Motivation
+
+To make promoting BitShares and bringing new users much more attractive to registrars and referrers by sharing UIAs and MPAs market fees with them.
+
+# Rationale
+When 'fill_order_operation' executing at the moment of market fee calculation there will be calculate the reward for the registrar used the parameter market_fee_reward_percent. Then this market_fee_reward_percent of the market fee is split between the registrar and referrer according to the referrer_rewards_percentage, which is set up during the new account registration by registrar (please see the parameters for register_account).
+
+# Specifications
+
+## market_fee_reward_percent asset option
+Percent of market fee that will be paid to buyer's registrar. Set by UIA issuer.
+
+## market_sharing_whitelist asset option
+An optional list of accounts (configurable by the UIA Issuer) could provide increased control over who is eligible to get market rewards.
+
+## New database and wallet api methods
+**get_account_reward(account, asset_symbol)** - Returns information about the reward amount for specific asset.
+**list_account_rewards(account)** - Returns information about the reward amount in various assets
+**claim_reward(account, asset_symbol, amount)** - Claim account's reward for the given asset.
+
+## account_reward_object
+A new BitShares object type in implementation space impl_account_reward_object_type = 2.18.X that tracks the rewards of a single account/asset pair
+```
+account_reward_object {
+owner_account,
+asset_type,
+reward_amount
+}
+```
+## account_reward_index
+A new index that stores objects of account_reward_object-type in graphene::database and allow random fast access to objects by given criteria
+```
+account_reward_index multi-index of account_reward_objects
+indexed_by
+[id]
+[owner_account, asset_type]
+[asset_type, reward_amount desc, owner_account]
+```
+## asset_claim_reward_operation
+A new operation used to transfer reward to the account's balance.
+```
+asset_claim_reward_operation {
+fee
+claiming_account
+amount_to_claim
+}
+```
+
+## graphene::chain::database new methods
+**get_reward(owner_account, asset_id)** - Retrieve a particular account's reward in a given asset
+**adjust_reward(account, delta)** - Adjust a particular account's reward in a given asset by a delta
+
+## graphene::chain::database modifications
+**pay_market_fees(seller, recv_asset_object, receives_amount)** - Split pay to asset issuer fee, registrar fee and referrer fee. If the registrar is not in the market_sharing_whitelist, split_pay will not happen and the entire fee goes to Asset issuer.
+**calculate_market_fee(trade_asset, trade_amount)** - Calculate value for previous function.
+**fill_limit_order(order, pays, receives, cull_if_small, fill_price, is_maker)** - Append hardfork (HARDFORK_REWARD_TIME) check. Use old or new version of pay_market_fees() function.
+
+## asset_create_evaluator, asset_update_evaluator modifications
+Append hardfork (HARDFORK_REWARD_TIME) check. Validate additional asset options. Apply additional asset options (market_fee_reward_percent, market_sharing_whitelist)
+
+## Unit tests
+
+Add: reward_tests.cpp (asset_rewards_test, asset_claim_reward_test)
+
+Modified: fee_tests.cpp, uia_tests.cpp
+
+# Description
+
+# Summary for Shareholders
+The new modification - market fee sharing - will allow to bring in new clients for BitShares by making it financially lucrative for registrars and referrers. This modification is interesting mostly so asset issuers and registrars/referrers.
+
+# Copyright
+This document is placed in the public domain.
+
+# See Also
+https://github.com/bitshares/bitshares-core/issues/1268
--- a/bsip-0043.md
+++ b/bsip-0043.md
@@ -4,7 +4,7 @@
     Status: Draft
     Type: Protocol
     Created: 2018-08-23
-    Updated: 2018-10-05
+    Updated: 2018-10-19
     Discussion: https://github.com/bitshares/bsips/issues/102
     Worker:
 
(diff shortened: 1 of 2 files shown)


xeroc:
Nice specs!

Instead of setting up a new object `account_reward_object`, I would also like to understand why not to reuse the infrastructure of vesting balances.

Also, I would actually think that there is more use for a **black** list than for a whitelist. Why not have both?

OpenLedgerApp:
>Instead of setting up a new object account_reward_object, I would also like to understand why not to reuse the infrastructure of vesting balances.

Vesting_balance_object  receives fees from all kinds of transactions.
In this case, a user can't find out how much profit is received from market fee sharing.
So it is better to create a new object account_reward_object in order to track profits from market fee sharing only.
>Also, I would actually think that there is more use for a black list than for a whitelist. Why not have both?

That's a very good point! However, considering there are only a few referrers, black list seems like excessive functionality. It's one or the other, there is no point using both. 
We think that instead of blacklisting, it's safer to whitelist registrars that we trust specifically.
@xeroc What do you think?

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+## market_sharing_whitelist asset option
+An optional list of accounts (configurable by the UIA Issuer) could provide increased control over who is eligible to get market rewards.
+
+## New database and wallet api methods
+**get_account_reward(account, asset_symbol)** - Returns information about the reward amount for specific asset.
+**list_account_rewards(account)** - Returns information about the reward amount in various assets
+**claim_reward(account, asset_symbol, amount)** - Claim account's reward for the given asset.
+
+## account_reward_object
+A new BitShares object type in implementation space impl_account_reward_object_type = 2.18.X that tracks the rewards of a single account/asset pair
```
pmconrad: Please consider using this mechanism for all kinds of rewards. Related: https://github.com/bitshares/bitshares-core/issues/1276
(Perhaps I misunderstand what you mean with "tracks" here - are you talking about rewards payment history?)

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+A new index that stores objects of account_reward_object-type in graphene::database and allow random fast access to objects by given criteria
+```
+account_reward_index multi-index of account_reward_objects
+indexed_by
+[id]
+[owner_account, asset_type]
+[asset_type, reward_amount desc, owner_account]
+```
+## asset_claim_reward_operation
+A new operation used to transfer reward to the account's balance.
```
pmconrad: Please explain why this is required.
It would be much simpler to either pay these rewards automatically, e. g. during maintenance like the current rewards system does, or to use the existing vesting balance mechanism.

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+
+To make promoting BitShares and bringing new users much more attractive to registrars and referrers by sharing UIAs and MPAs market fees with them.
+
+# Rationale
+When 'fill_order_operation' executing at the moment of market fee calculation there will be calculate the reward for the registrar used the parameter market_fee_reward_percent. Then this market_fee_reward_percent of the market fee is split between the registrar and referrer according to the referrer_rewards_percentage, which is set up during the new account registration by registrar (please see the parameters for register_account).
+
+# Specifications
+
+## market_fee_reward_percent asset option
+Percent of market fee that will be paid to buyer's registrar. Set by UIA issuer.
```
pmconrad: ~~UIA issuer~~ asset owner

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+    Status: Draft
+    Type: Protocol
+    Created: 2018-08-23
+    Discussion: https://github.com/bitshares/bsips/issues/102
+    Worker:
+
+
+# Abstract
+
+When creating a new asset, the asset issuer is the only beneficiary of the market fees in the current implementation. And the only way to affect the asset community growth is the market fee percentage. For example, one can decrease the market fee and it will result in somewhat larger number of trades with this asset. In this way the asset issuer might get a bigger profit during to increasing the trade volume with this asset.
```
pmconrad: Certainly not "the only way". I doubt that global BTC trade volume depends on market fees alone.

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+
+However, there might be another opportunity to promote the asset and stimulate the trading - use native Bitshares referral program. At this time unfortunately an assets owner is not able to share market fees with registrars and referrers to stimulate them to market the asset trading, so we suggest to add this possibility. Furthermore, enabling this feature for MPAs (e.g. bitCNY or bitUSD) can provide additional bounty for Bitshares registrars and referrals which can lead to more traders joining to the ecosystem.
+The workflow may be as following:
+
+An asset issuer defines the market_fee_reward_percent - what percentage of the market fee he wants to share with the registrar in asset options. Registrar defines reward_percent for the referrer for each user(using the already existing BitShares mechanism).
+Market fee reward is accumulated on the user account. The user decides when they want to claim the market fee reward and move it to their active balance. There is another operation created asset_claim_reward_operation. Each user pays network fee to call this operation.
```
pmconrad: Why so complicated?

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+# Rationale
+When 'fill_order_operation' executing at the moment of market fee calculation there will be calculate the reward for the registrar used the parameter market_fee_reward_percent. Then this market_fee_reward_percent of the market fee is split between the registrar and referrer according to the referrer_rewards_percentage, which is set up during the new account registration by registrar (please see the parameters for register_account).
+
+# Specifications
+
+## market_fee_reward_percent asset option
+Percent of market fee that will be paid to buyer's registrar. Set by UIA issuer.
+
+## market_sharing_whitelist asset option
+An optional list of accounts (configurable by the UIA Issuer) could provide increased control over who is eligible to get market rewards.
```
pmconrad: ~~UIA issuer~~ asset owner

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+
+## New database and wallet api methods
+**get_account_reward(account, asset_symbol)** - Returns information about the reward amount for specific asset.
+**list_account_rewards(account)** - Returns information about the reward amount in various assets
```
pmconrad: database api calls should have ``start`` and ``limit`` parameters

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+}
+```
+## account_reward_index
+A new index that stores objects of account_reward_object-type in graphene::database and allow random fast access to objects by given criteria
+```
+account_reward_index multi-index of account_reward_objects
+indexed_by
+[id]
+[owner_account, asset_type]
+[asset_type, reward_amount desc, owner_account]
```
pmconrad: Please consider moving indices that are not generally useful or consensus-related into optional plugins.

bsip-0043.md
```diff
@@ -0,0 +1,96 @@
...
+amount_to_claim
+}
+```
+
+## graphene::chain::database new methods
+**get_reward(owner_account, asset_id)** - Retrieve a particular account's reward in a given asset
+**adjust_reward(account, delta)** - Adjust a particular account's reward in a given asset by a delta
+
+## graphene::chain::database modifications
+**pay_market_fees(seller, recv_asset_object, receives_amount)** - Split pay to asset issuer fee, registrar fee and referrer fee. If the registrar is not in the market_sharing_whitelist, split_pay will not happen and the entire fee goes to Asset issuer.
```
pmconrad: Please specify behaviour if the new options are absent, or if the whitelist is empty.

pmconrad: Also, note that fee is not paid out the the asset issuer but collected in the asset's fee pool (current behaviour).

abitmore:
> Vesting_balance_object receives fees from all kinds of transactions.
In this case, a user can't find out how much profit is received from market fee sharing.
So it is better to create a new object account_reward_object in order to track profits from market fee sharing only.

I think the approach (adding new objects and operations for market fee sharing only) is over-complicated. In order to distinguish a deposit is a market fee sharing or another (including witness pay, worker pay, trx fee sharing and etc), we can add a field in `vesting_balance_object` to indicate its source, and/or add fields in the virtual deposit operations, which IMHO is better for expanding further in the future.

bsip-0043.md
OpenLedgerApp: concerning vesting balance mechanism, please see our response to xeroc above and to abit below.
And paying out directly during maintenance, if we understand correctly, there is no such mechanism right now. In maintenenance, only process_fees is called, that distributes accumulated fees to the network, registrar and rererrer using deposit_cashback (the vesting balance mechanism we discussed above).

abitmore: > see our response to xeroc above and to abit below.

Sorry I didn't see a new response so far.

> ... if we understand correctly, there is no such mechanism right now. In maintenenance, only process_fees is called ...

Since we're discussing how to implement a new feature, I think adding more tasks to be done in maintenance interval **IS** an option.

bsip-0043.md
OpenLedgerApp: Yes, you're right. this could be a potential solution! 
However, the related https://github.com/bitshares/bitshares-core/issues/1276  is not implemented yet. 
So one way would be to implement the way that is described in the proposed BSIP.
And the other way would be to Implement #1276 first. Then, we can use this mechanism to implement part of BSIP-43. 
Our team can definitely implement #1276  - providing that estimation and budget are approved first. 
If that's approved  - we are totally up to changing bsip to use the newly implementing #1276 features.

bsip-0043.md
OpenLedgerApp: You're right, it's not the only way, but one of the ways

bsip-0043.md
OpenLedgerApp: It's not more complicated than vesting balance mechanism. In fact, the idea is copied from the vesting balance mechanism and then we simplified to some extent.

bsip-0043.md
OpenLedgerApp: These terms are interchangeable in our understanding. 
For example if we look in the code: in the general case, issuer = owner. It is the same account. 
```
class asset_object : public graphene::db::abstract_object<asset_object>{
...
  /// ID of the account which issued this asset.
  account_id_type issuer;
  ...
 };
```
There does not seem to be owner field. It's called issuer. 
If the current widely accepted way would be to use the term "asset owner" - we can definitely change it everywhere in the document. 
Do you think we should change to *asset owner*

bsip-0043.md
OpenLedgerApp: We think that one user will generally have only a few assets with rewards.  So we don't think pagination is needed in this particular case. On one hand, it's definitely good to follow the common standards. 
On the other hand, in bitshares there are other well-known functions that do not have pagination, for example: 
**list_account_balances(account_id)**

Alternatively, we can change list_account_rewards to **get_account_rewards**. 
Or if following standards is becoming a trend in bitshares, we can add start and limit parameters as you suggested. 
What do you think?

bsip-0043.md
OpenLedgerApp: We do not think that these indices belong to a plugin, because it's definitely consensus-related.

bsip-0043.md
OpenLedgerApp: if market_fee_reward_percent = 0 or absent - the old mechanism is used. Market fees are accumulated in the asset and can be claimed.

If whitelist empty or absent - there is no filtering. That is like everyone is in the whitelist by default if it is empty.

OpenLedgerApp: >Also, note that fee is not paid out the the asset issuer but collected in the asset's fee pool (current behaviour).

Yes, this mechanism stays the same way. The fees are accumulated, then claimed, not transferred. However, now not the whole market fee goes to the asset fee pool, but only a part of it: ```market_fee*(1 - market_fee_reward_percent)```. And ```market_fee* market_fee_reward_percent``` will get to the registar's fee pool .

bsip-0043.md
OpenLedgerApp: >Sorry I didn't see a new response so far.

sorry, @abitmore we planned to respond to your commment below, but i think we covered most of it to the response to Peter's comment above :)
Denis

OpenLedgerApp:
@abitmore We discussed these issues in the first comment of Peter Conrad. 
We can definitely implement it using vesting balance mechanism (having done #1276 first).

bsip-0043.md
abitmore: I guess Peter's opinion was that it's inappropriate to use `UIA` here, since the option can be used on MPAs as well.

bsip-0043.md
abitmore: > why fees distribution happens through vesting balance.

1. cashback is designed to have a vesting period;
2. for better performance for non-maintenance blocks (only process fees on maintenance blocks). Updating one object v.s. updating 3 objects and related looking-ups for the objects. Arguable though.

bsip-0043.md
pmconrad: Makes sense. I don't particularly care, can leave it as is.

bsip-0043.md
pmconrad: Thanks, I assumed as much. But you have to update the BSIP accordingly. Comments here are not part of the specification. :-)

bsip-0043.md
OpenLedgerApp: I think we can switch to the **```asset owner```** term, no problem

bsip-0043.md
OpenLedgerApp: Guys, we had another discussion today about direct transfer.  So we think having direct transfer does not generate separate transactions and hence does not generate network fee. 
So we think that the best way would be to recreate  vesting-like mechanism - so that we utilize  a similar mechanism to the  current mechanism of claiming accumulated fees from assets  (but for referrers and registrars. ) Then they have to use claim  - to claim their accumulated fees and hence, generate network fees for bitshares.

Alternatively, as we mentioned, using current vesting mechanism would be a great idea, but we then need to discuss the estimation and priority for #1276. Should we start this discussion? 
What do you think?
Denis

abitmore: Sorry, I didn't get it. Which is better in your opinion, "generate network fees", or not? I foresee that reward claiming fee won't be any big part of overall transaction fee. It's even possible that the committee will set balance claiming fee to zero since it removes data from RAM.

From a technical point of view, having to claim means more signed transaction on the chain, which will lead to chain bloating to an extent (more permanent storage); having to maintain unclaimed balances means larger chain state (more RAM usage). On the other hand, seems it requires less computation (less CPU usage).

In regards to user experience (UX) and business development (BD), each approach has its own pros and cons.

It's all about compromise.

sschiessl-bcp:
Are the aboves reviews all included/handled in the PR?

OpenLedgerApp:
> Are the aboves reviews all included/handled in the PR?

Stefan, what's a PR?

OpenLedgerApp:
> Are the aboves reviews all included/handled in the PR?

I will add the discussed points.

sschiessl-bcp:
Ping me when dealt with past reviews.

OpenLedgerApp:
We Updated the BSIP-43 according to the discussions with community members

pmconrad:
Thanks, looks good to me.
But I would really prefer if https://github.com/bitshares/bitshares-core/issues/1276 was implemented first and the generic mechanism be used here as well.

bsip-0043.md
sschiessl-bcp: I want to add that it also makes a differences for taxes.

In germany, vesting balance is an "Option", whereas a direct payout would have direct tax implications.

bsip-0043.md
sschiessl-bcp: > TBH I think this is over-detailed to be in a BSIP document. It's implementation details, which can be decided by the developer.

Agreed

OpenLedgerApp: Perhaps you're right! We have already half-implemented this BSIP when we did a prototype as part of the investigation if it could be done at all. 
So this is already described and to take it out would mean to spend more time :)

sschiessl-bcp: The BSIP should not tie you down, and allow to switch if anything would need to be restructured.

Put an abstract representation of what you did in it, not the actual representation.

abitmore:
Looks good to me so far, and I'm inline with Peter's preference.

Need to resolve conflicts again.

xeroc:
> I think the approach (adding new objects and operations for market fee sharing only) is over-complicated. In order to distinguish a deposit is a market fee sharing or another (including witness pay, worker pay, trx fee sharing and etc), we can add a field in `vesting_balance_object` to indicate its source, and/or add fields in the virtual deposit operations, which IMHO is better for expanding further in the future.

For cashback, the account_object has an attribute that links the cashback-related vesting balance. Every account could get their own "market fee share cashback" attribute which links to the other vesting balance ... No one says we need to combine them into one vesting balance

OpenLedgerApp:
> For cashback, the account_object has an attribute that links the cashback-related vesting balance. Every account could get their own "market fee share cashback" attribute which links to the other vesting balance ... No one says we need to combine them into one vesting balance

We think this is a great idea. Then #1276 becomes not required for this approach.
We can update the document. 
@abitmore @pmconrad  please voice your opinions as soon as you can so that we can update the BSIP.

Also, guys, do we still need #1276? We're doing the estimations right now, so we need to know if we should proceed with estimations.

pmconrad:
Should work IMO.

abitmore:
> Every account could get their own "market fee share cashback" attribute which links to the other vesting balance ... No one says we need to combine them into one vesting balance

I disagree the approach of storing IDs of vesting balances in account object, because there would be many assets thus many vesting balance objects for each account that need to be linked. The performance would be impacted. IMHO it's better to query vesting balance index directly by new indices. An additional field indicating the source would be necessary for depositing.

IMHO https://github.com/bitshares/bitshares-core/issues/1276 is independent. This PR is about how to process funds. https://github.com/bitshares/bitshares-core/issues/1276 is about how to track history.

OpenLedgerApp:
> I disagree the approach of storing IDs of vesting balances in account object, because there would be many assets thus many vesting balance objects for each account that need to be linked. The performance would be impacted. IMHO it's better to query vesting balance index directly by new indices. An additional field indicating the source would be necessary for depositing.

 Ok, we can take this approach as well.  @pmconrad  @xeroc  what do you think of abit's idea?

pmconrad:
IMO that's an implementation detail, we don't need to agree on a final solution here and now.
Unsure which is better, both solutions have advantages and disadvantages.

What *should* be specified is when and how new vesting_balances are created, since vesting_balance_ids are relevant to the protocol (vesting_balance_withdraw requires it).

xeroc:
I agree with @abitmore . havent thought about market fee potentially being many different assets.

OpenLedgerApp:
Guys, based on the discussion above here’s the way we suggest to update BSIP:
 
There should be a field called Source added into the vesting_balance _object. This field should indicate what kind of vesting balance it is. 
There are the following balances in BitShares currently: worker, cashback. We will add another type: Market fee sharing .
Each asset will have its own vesting_balance.
This implementation will allow to track each asset balance while not affecting performance much.
 
It is important to note that the new field affects the database and we will need to do node replay.
Where as with the reward_object there will be no need to do node replay as we do not change the DB.
 
We assume that everyone agrees with this way of updating BSIP.
Please let us know  if we still need to discuss anything or if anyone still wants to add his opinion.
Thanks, Denis

bsip-0043.md
```diff
@@ -15,99 +15,42 @@ When creating a new asset, the asset owner is the only beneficiary of the market
 
 However, there might be another opportunity to promote the asset and stimulate the trading - use native Bitshares referral program. At this time unfortunately an assets owner is not able to share market fees with registrars and referrers to stimulate them to market the asset trading, so we suggest to add this possibility. Furthermore, enabling this feature for MPAs (e.g. bitCNY or bitUSD) can provide additional bounty for Bitshares registrars and referrals which can lead to more traders joining to the ecosystem.
 
-The workflow may be as follows. It is copied from vesting balance mechanism and got simplified.
+An asset owner defines the **market_fee_reward_percent**  in asset options - what percentage of the market fee he wants to share with the registrar.
+**market_fee** * **market_fee_reward_percent** goes the registar's fee pool.
+**market_fee** * (1 - **market_fee_reward_percent**) goes to the asset owner.
```
abitmore: 1. Although I've read below that referrers will get a cut of the reward, words here seems only registrar will get rewarded;
2. registrar is an account which doesn't have a fee pool. Asset has a fee pool.

bsip-0043.md
```diff
@@ -16,11 +16,11 @@ When creating a new asset, the asset owner is the only beneficiary of the market
 However, there might be another opportunity to promote the asset and stimulate the trading - use native Bitshares referral program. At this time unfortunately an assets owner is not able to share market fees with registrars and referrers to stimulate them to market the asset trading, so we suggest to add this possibility. Furthermore, enabling this feature for MPAs (e.g. bitCNY or bitUSD) can provide additional bounty for Bitshares registrars and referrals which can lead to more traders joining to the ecosystem.
 
 An asset owner defines the **market_fee_reward_percent**  in asset options - what percentage of the market fee he wants to share with the registrar.
-**market_fee** * **market_fee_reward_percent** goes the registar's fee pool.
+**market_fee** * **market_fee_reward_percent** goes the registrar.
```
abitmore: Goes to (I don't know if my grammar is correct though)?

PR #108, opened 2018-09-03
Escrow tech specs

jmjatlanta:
Changed filename, added some tech specs

- [x] Add parameters

### DIFF
--- a/bsip-1041.md
+++ b/bsip-1041.md
@@ -261,7 +261,17 @@ struct escrow_object {
 
 Create a new escrow contract
 ```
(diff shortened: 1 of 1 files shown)


abitmore:
Why 1041?

jmjatlanta:
> Why 1041?

Because it alerts me that it needs to change. I do not know when or how numbers are assigned. I'm sure someone will help me with that, and I will correct it.

jmjatlanta:
This PR is very outdated and no longer applies.

PR #112, opened 2018-10-12
grammar fixes BSIP 44 (HTLC)

jmjatlanta:
This PR fixes a few "bugs" in the HTLC BSIP. 

Comments from @abitmore that require more discussion: 

https://github.com/bitshares/bsips/pull/104#pullrequestreview-164386716

Specifically:
1) Whether (and when) HTLCs should obey whitelisting/blacklisting
2) Risk of exposing preimage if there is a badly-timed chain reorganization or transaction censoring
3) How the summary should be worded (shortened).

Please discuss the changes made, Abit's concerns, and anything else you see that should be fixed below.

### DIFF
--- a/bsip-0044.md
+++ b/bsip-0044.md
@@ -70,9 +70,9 @@ There are two competing conditions within an HTLC, the `hash lock` and the `time
 
 The HTLC contains a `hash lock` condition, which comprise both the `preimage hash` and `preimage length`, barring the transfer of held `secured assets` unless satisfied. If a `preimage` of requisite `length` is provided to the HTLC which generates a hash matching the `preimage hash`, the `preimage` is then stored within the blockchain, and the `secured assets` are transferred to the `recipient`.
 
-If a satisfactory `preimage` is not provided to the HTLC before the stipulated `time lock` expires, the `depositor` may request the return of `secured assets`. The HTLC will only evaluate transfer request from `depositor` and after `timeout threshold`, then return `secured assets` to `depositor`.
+If a satisfactory `preimage` is not provided to the HTLC before the stipulated `time lock` expires, the HTLC will then return `secured assets` to `depositor`.
 
-**Note:** we recommend the Committee the set maximum allowable `preimage length` to ensure unreasonably large submissions are rejected. 
+**Note:** we recommend the Committee set maximum allowable `preimage length` to ensure unreasonably large submissions are rejected. 
 
 ### **Condition Evaluators**
 
@@ -100,7 +100,7 @@ To protect the `recipient`, early termination of an HTLC is not allowed by any p
 
 ### **Automatic Transfers Upon Expiry**
 
-Upon expiry of the `timeout threshold`, the `secured assets` held within the HTLC will be queued for return to `depositor`. From this time, the HTLC will no longer evaluate the `hash lock`, preventing `recipient` from receiving the `secured assets`. No action is required by the `depositor` to receive their "locked" funds back from the contract after expiry. 
+Upon expiry of the `timeout threshold`, the `secured assets` held within the HTLC will be returned to `depositor`. From this time, the HTLC will no longer evaluate the `hash lock`, preventing `recipient` from receiving the `secured assets`. No action is required by the `depositor` to receive their "locked" funds back from the contract after expiry. 
 
 ### **Fees**
 
--- a/bsip-0044.md
+++ b/bsip-0044.md
@@ -190,9 +190,9 @@ Bob has now observed the `preimage` Alice used to "unlock" his HTLC, and he will
 
 Note 1: The preimage hash algorithm must be speceified as  either ``SHA256``, ``RIPEMD160``, or ``SHA1``. The use of ``SHA1`` is added for compatibility to other chains, but its use is discouraged.
 
-Note 2: The preimage size must be less than 65536 bytes.
+Note 2: The preimage size must be less than the maximum number of bytes as specified by the committee.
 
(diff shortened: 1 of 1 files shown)


pmconrad:
Please specify initially supported hash algorithms
Please add new chain parameter 'max_htlc_preimage_length' and default value to specification section

abitmore:
> Please add new chain parameter 'max_htlc_preimage_length' and default value to specification section

`preimage_size` is defined as `uint16_t`, which implies that the maximum length should be `65535` or less.

jmjatlanta:
> Please add new chain parameter 'max_htlc_preimage_length' and default value to specification section

Suggestions as to what that limit should be? I can increase the variable size, if needed. I think 65K is plenty big for now. If people desire to use big stuff for preimages (i.e. real images) it could be changed later. My guess is most will use random gobldygook generated by their software.

abitmore:
I think we'd better implement https://github.com/bitshares/bitshares-core/issues/167 before adding new parameters to global parameters (which needs a new BSIP as well), although it's possible to add new parameters now via extensions (see https://github.com/cryptonomex/graphene/pull/612).

jmjatlanta:
I went ahead and added some limits, just to see what you think. The BSIP still says that the committee should set the max timeout and max preimage size. So we need to pick one or the other.

I agree that bitshares/bitshares-core/#167 should be implemented. I would rather do that than shoe-horn in a parameter via an extension. 

So, I am looking for suggestions:
1) Hard-code values (at least for now), and not permit the committee to change it.
2) Add via extensions
3) Push a BSIP for #167 and get that in first.

pmconrad:
IMO it makes little sense to use a preimage that is longer than the hash size. I'd suggest a default maximum of 1024 bytes. Maybe even less.

I think HTLC is more imporant than https://github.com/bitshares/bitshares-core/issues/167, so we shouldn't create a dependency there. Still, the parameter should be a chain_parameter right from the start. I think adding the extension would be the best option.

jmjatlanta:
> IMO it makes little sense to use a preimage that is longer than the hash size. I'd suggest a default maximum of 1024 bytes. Maybe even less.
> 
> I think HTLC is more imporant than [bitshares/bitshares-core#167](https://github.com/bitshares/bitshares-core/issues/167), so we shouldn't create a dependency there. Still, the parameter should be a chain_parameter right from the start. I think adding the extension would be the best option.

All that makes sense. My desire to not do things twice was greater than what is logical. Updated spec to eliminate the hard-coding and went back to committee flexibility.

jmjatlanta:
I have modified the document to more closely match the implementation, and to make terminology consistent. If @abitmore or @pmconrad could please give it a second look to verify its accuracy. Thanks!

bsip-0044.md
```diff
@@ -177,13 +177,11 @@ Bob has now observed the `preimage` Alice used to "unlock" his HTLC, and he will
...
-         account_id_type         depositor;
-         account_id_type         recipient;
+         account_id_type         from;
+         account_id_type         to;
          asset                   amount;
          fc::time_point_sec      expiration;
-         asset                   pending_fee;
-         vector<unsigned char>   preimage_hash;
-         fc::enum_type<uint8_t, hash_algorithm> preimage_hash_algorithm;
+         htlc_hash               preimage_hash;
```
pmconrad: need to define htlc_hash type if you mention it here

bsip-0044.md
```diff
@@ -199,59 +197,57 @@ Note 3: The initial HTLC expiration can not be more than the maximum amount of t
...
-  Set: `contract.depositor` = `depositor`
-  Set: `contract.recipient` = `recipient`
+  Set: `contract.from` = `from`
+  Set: `contract.to` = `to`
   Set: `contract.hash_algorithm` = `hash_algorithm`
   Set: `contract.preimage_hash` = `preimage_hash`
   Set: `contract.preimage_length` = `preimage_length`
   Set: `contract.timeout_treshold` = `timeout_threshold`
-  Transfer: from `depositor` account to `contract.quantity` of `contract.symbol`
+  Transfer: remove `contract.quantity` of `contract.symbol` from from` account
```
pmconrad: Missing backtick before second "from"

bsip-0044.md
```diff
@@ -172,80 +172,82 @@ Bob has now observed the `preimage` Alice used to "unlock" his HTLC, and he will
...
+         account_id_type         from;
+         account_id_type         to;
+         asset                   amount;
+         fc::time_point_sec      expiration;
+         htlc_hash               preimage_hash;
+         uint16_t				      preimage_size;
+   };
 ```
 
+Note 1: The preimage hash algorithm must be speceified as  either ``SHA256``, ``RIPEMD160``, or ``SHA1``. The use of ``SHA1`` is added for compatibility to other chains, but its use is discouraged.
```
pmconrad: This note is no longer applicable.

pmconrad:
Generally, I think BSIPs should not be modified after approval. Modifying implementation details is probably ok to prevent future confusion.

This PR contains a little more than just "grammar fixes" and implementation details. I think these additional changes do not modify the original intent and are therefore probably ok too.

bsip-0044.md
```diff
@@ -172,6 +172,12 @@ Bob has now observed the `preimage` Alice used to "unlock" his HTLC, and he will
 ## **Objects**
 
 ```
+   typedef fc::static_variant<
+      htlc_algo_ripemd160,
+      htlc_algo_sha1,
+      htlc_algo_sha256
```
pmconrad: Now you should mention what these three new types are... :-)

jmjatlanta: Do you think it best to put it in a descriptive note below this section or a //comment on each line? I'll give //comment a try.

PR #114, opened 2018-10-15
BSIP47 - An explicit voting operation

xeroc:
This BSIP comes together with a pull request for a [**proposed** implementation](https://github.com/bitshares/bitshares-core/pull/1367).

It is sponsored by Blockchain Projects BV

### DIFF
--- a/bsip-0047.md
+++ b/bsip-0047.md
@@ -1,0 +1,99 @@
+    BSIP:       <BSIP number>
+    Title:      An explicit voting operation
+    Authors:    Fabian Schuh <https://github.com/xeroc>
+                Dmitrij Vinokour <https://github.com/Dimfred>
+    Status:     Draft
+    Type:       Protocol
+    Created:    2018-09-20
+    Discussion: <url>
+    Worker:     <Id of worker proposal> (optional)
+
+# Abstract
+
+On the BitShares Blockchain the `account_update_operation` allows to change account
+related settings. In the current implementation all possible settings have to be
+passed to the operation, to not override old settings. This is inconvenient
+because in certain situations there is no need in changing all settings.
+
+In this BSIP the `account_update_votes_operation` is introduced, which solves the
+described problem.
+
+# Motivation
+
+The current `account_update_operation` forces the user to pass all possible account
+settings, to not override existing ones. This is inconvinient and also requires a
+higher payload to be stored on the blockchain, thus results in a higher fee.
+
+To reduce the fee the `account_update_votes_operation` is introduced to update only
+a part of the `account data`.
+
+This new operation will be particularily useful in conjunction with
+[BSIP40](https://github.com/bitshares/bsips/blob/master/bsip-0040.md)
+which will allow to assign a specific key for this operation and
+simplifies life for voters, significantly.
+
+# Rational
+
+The `account_update_votes_operation` is an account related operation and allows to set
+variables in the `account_object`. The targeted variables of this operation are located
+in the `account_options field` of the `account_object`.
+
+- `voting_account`  (the voting account to use)
+- `num_witness`     (number of witness votes)
+- `num_committee`   (number of committee votes)
+- `votes`           (list of `vote_id_type`)
+
+As described before to not override existing settings, every other needed variable is fetched
+with the database, when the operation is evaluated.
+
+# Specifications
+
+This BSIP comes with a change to the protocol, which requires a propotocol upgrade.
+
+In the following an exemplary implementation is given.
+
+    struct account_update_votes_operation : public base_operation
+    {
+     asset fee;
+     /// The account to update
+     account_id_type account;
+     /// New account options
+     flat_set<vote_id_type>    votes_to_add;
+     flat_set<vote_id_type>    votes_to_remove;
+     // a new voting account to set
+     optional<account_id_type> voting_account;
+     // a new number of witness_votes to set
+     optional<uint16_t> num_witness;
+     // a new number of commitee_votes to set
+     optional<uint16_t> num_committee;
+     // for future extensions (see account_update_operation)
+     extensions_type extensions;
+    };
+
+
+We propose to allow adding and removing votes individually by means of
+defining a *list* of vote ids to be added or removed. This has multiple
+advantages:
+
+* Reduces the size of the operation compared to replacing the entire
+  slate
+* Allows to easily compare votes before and after the operation
+
+Setting the proxy is explicitly not part of this opertation.
+Authors of this BSIP came to the conclusion that proxy settings should
+require `active authority`. It is their opinion that setting proxy
+settings are distinct from voting.
+
+# Discussion
+
+See BSIP issues on github.
+
+# Summary for Shareholders
+
+This BSIP simplifies the way voting occurs on the blockchain and will
+later (together with BSIP40) simplify automated voting without
+compromising the account.
+
+# Copyright
+
+This document is placed in the public domain.
--- a/bsip-0047.md
+++ b/bsip-0047.md
@@ -126,16 +126,24 @@ Changes to direct votes can be specified by means of a *list* of vote identifier
 * `witness_voting_account`, if present, must be a valid account ID, must exist
 * `worker_voting_account`, if present, must be a valid account ID, must exist
 * Validate the existence of each vote to add and remove
-* `num_witness`, if present, must be greater than or equal to  &ge; 0
-* `num_committee`, if present, must be greater than or equal to  &ge; 0
+* `num_witness`, if present, must be greater than or equal to  (&ge;) 0 and must not be greater than the resulting list of witness votes
+* `num_witness`, if not present and if witness votes are removed, the resulting list of witness votes must not be smaller than the current `num_witness` setting of the account
+* `num_committee`, if present, must be greater than or equal to  (&ge;) 0 and must not be greater than the resulting list of committee votes
+* `num_committee`, if not present and if committee votes are removed, the resulting list of committee votes must not be smaller than the current `num_committee` setting of the account
 
 **Evaluation:**
 
 * Update the account's `account_options.committee_votes`, `account_options.witness_votes`, and `account_options.worker_votes` by filtering the `votes_to_add` and `votes_to_remove` field appropriately
-* Update the `account_options.num_witness` if `witness_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` and if `num_witness` is present
-* Update the `account_options.num_committee` if `committee_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` and if `num_committee` is present
+* If any of `committee_voting_account`, `witness_voting_account` or `worker_voting_account` is present
+    * Update the corresponding field in the account's `account_options`
+    * Set the account's `account_options.voting_account` to `GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT`
+* Update the `account_options.num_witness`
+    * If votes_to_add and votes_to_remove contain _witness_ votes that are accepted regardless of GRAPHENE_PROXY_TO_SELF_ACCOUNT
+    * if `witness_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` and if `num_witness` is present
+* Update the `account_options.num_committee`
+    * If votes_to_add and votes_to_remove contain _committee_ votes that are accepted regardless of GRAPHENE_PROXY_TO_SELF_ACCOUNT
+    * if `committee_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` and if `num_committee` is present
 * Update the account's last vote time to the head block time
-
 
 ### `account_create_operation` (Existing)
 
--- a/bsip-0047.md
+++ b/bsip-0047.md
@@ -1,99 +1,182 @@
-    BSIP:       <BSIP number>
-    Title:      An explicit voting operation
+    BSIP:       47
+    Title:      Vote Proxies for Different Referendum Categories
     Authors:    Fabian Schuh <https://github.com/xeroc>
                 Dmitrij Vinokour <https://github.com/Dimfred>
+                Alex Megalokonomos <https://github.com/clockworkgr>
+                Michel Santos <https://github.com/MichelSantos>
     Status:     Draft
     Type:       Protocol
     Created:    2018-09-20
-    Discussion: <url>
-    Worker:     <Id of worker proposal> (optional)
+    Discussion: https://github.com/bitshares/bsips/pull/114
+
 
 # Abstract
 
-On the BitShares Blockchain the `account_update_operation` allows to change account
-related settings. In the current implementation all possible settings have to be
-passed to the operation, to not override old settings. This is inconvenient
-because in certain situations there is no need in changing all settings.
+Voting power by core token holders may be assigned to different proxies in each of the BitShares referendum categories: witnesses, committee members, and worker proposals.
 
-In this BSIP the `account_update_votes_operation` is introduced, which solves the
-described problem.
 
 # Motivation
 
-The current `account_update_operation` forces the user to pass all possible account
-settings, to not override existing ones. This is inconvinient and also requires a
-higher payload to be stored on the blockchain, thus results in a higher fee.
+Some proxies can be more or less knowledgeable, wise, and trusted by token holders to vote on certain referendum categories than others.  It is therefore desirable to empower token holders to select different proxies for each of the referendum categories.
 
-To reduce the fee the `account_update_votes_operation` is introduced to update only
-a part of the `account data`.
+If the ability to select multiple proxies or to directly vote through a single new operation were made possible, this new operation be used in conjunction with [BSIP40](https://github.com/bitshares/bsips/blob/master/bsip-0040.md) which will allow an account holder to assign a specific key for voting.  This combination could simplify voting for many account holders.
 
-This new operation will be particularily useful in conjunction with
-[BSIP40](https://github.com/bitshares/bsips/blob/master/bsip-0040.md)
-which will allow to assign a specific key for this operation and
-simplifies life for voters, significantly.
 
-# Rational
+# Rationale
 
-The `account_update_votes_operation` is an account related operation and allows to set
-variables in the `account_object`. The targeted variables of this operation are located
-in the `account_options field` of the `account_object`.
+Holders of the core token of BitShares (BTS) have always had the ability to directly vote on three referendum categories: witnesses, committee members, and worker proposals.  They have also had the option of delegating their voting power to another account (a "proxy") to vote on any of these referendum categories.
 
-- `voting_account`  (the voting account to use)
-- `num_witness`     (number of witness votes)
-- `num_committee`   (number of committee votes)
-- `votes`           (list of `vote_id_type`)
+Rather than delegating voting power to a single proxy for each of three referendum, a core token holder might prefer to select different proxies for each of the referendum categories.
 
-As described before to not override existing settings, every other needed variable is fetched
-with the database, when the operation is evaluated.
+This more granular approach empowers core token holders with more options.  A BTS token holder may choose to directly vote on any referendum category including the option to abstain.  A BTS token holder may choose to select proxies for each of the referendum categories.  Each of the referendum categories may have distinct proxies or may have common proxies.  Some referendum categories may have assigned proxies while others may be unassigned.  _Any referendum category that does not have an assigned proxy permits the voter to vote directly on any referendum in that category._  _But if a **confused** core token holder votes for a referendum in a category **while also** delegating that category to a proxy, the token holder's direct vote shall be ignored._
+
+The introduction of a new operation (`account_update_votes_operation`) with additional changes to the core shall also enable an account holder to select proxies for different referendum categories, and to vote for referendums without needing to use the existing `account_update_operation`.  The existing operation requires the account holder to pass all possible account settings to avoid overriding existing settings.  This is inconvinient and requires a higher payload to be stored on the blockchain which results in the account holder paying a higher fee.  The proposed operation shall only update the portion of the `account data` that is related to voting and should also reduce the fee.
+
+The new operation only needs to include
+
+- The voting account
+- The desired proxy, if any, for the _committee_ referendum category
+- The desired proxy, if any, for the _witness_ referendum category
+- The desired proxy, if any, for the _worker_ referendum category
+- The quantity of desired witnesses
+- The quantity of desired committee members
+- The list of referendums that are supported
+
 
 # Specifications
 
-This BSIP comes with a change to the protocol, which requires a propotocol upgrade.
+Whenever a proxy for a category is _assigned_, any direct votes by a core token holder within that category shall be _ignored_.  Whenever a proxy for a category _is unassigned_, any direct votes by a core token holder for that category _shall be counted_ during the vote tallying process of the blockchain.  **This shall require an upgrade to the protocol.**
 
-In the following an exemplary implementation is given.
+### `GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT` (New)
 
-    struct account_update_votes_operation : public base_operation
-    {
-     asset fee;
-     /// The account to update
-     account_id_type account;
-     /// New account options
-     flat_set<vote_id_type>    votes_to_add;
-     flat_set<vote_id_type>    votes_to_remove;
-     // a new voting account to set
-     optional<account_id_type> voting_account;
-     // a new number of witness_votes to set
-     optional<uint16_t> num_witness;
-     // a new number of commitee_votes to set
-     optional<uint16_t> num_committee;
-     // for future extensions (see account_update_operation)
-     extensions_type extensions;
-    };
+A new special account `GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT` with ID `1.2.6` is defined and added to the database at protocol change activation time.
 
 
-We propose to allow adding and removing votes individually by means of
-defining a *list* of vote ids to be added or removed. This has multiple
-advantages:
+### `account_options` extension (Existing)
 
-* Reduces the size of the operation compared to replacing the entire
-  slate
+**Fields:**
+
+The `account_options` receives three new extensions:
+
+* `optional<account_id_type> committee_voting_account`
+* `optional<account_id_type> witness_voting_account`
+* `optional<account_id_type> worker_voting_account`
+
+**Validation:**
+
+* These extensions must not be present before the protocol change activation date, neither in transactions nor in proposals.
+* If any extension is present, it must contain the ID of an existing account.
+* If any extension is present, `account_options.voting_account == GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT`
+
+
+### `account_object`(Existing)
+
+**Note:** this is an implementation detail and meant to be a hint for later development. The actual implementation may differ.
+
+The account object shall receive these new fields:
+
+* `flat_set<vote_id_type> committee_votes`
+* `flat_set<vote_id_type> witness_votes`
+* `flat_set<vote_id_type> worker_votes`
+
+`account_create_operation` and `account_update_operation` shall be modified to create these sets by filtering the `account_options.votes` field appropriately. `account_update_votes_operation` shall also create these sets by filtering the `votes_to_add` and `votes_to_remove` fields appropriately.
+
+
+### `account_update_votes_operation` (New)
+
+The following is an example implementation.
+
+```
+struct account_update_votes_operation : public base_operation
+{
+	asset fee;
+
+	/// The account to update
+	account_id_type account;
+
+	/// New account options
+	flat_set<vote_id_type>    votes_to_add;
+	flat_set<vote_id_type>    votes_to_remove;
+
+	// A new voting account to set
+	optional<account_id_type> committee_voting_account
+	optional<account_id_type> witness_voting_account
+	optional<account_id_type> worker_voting_account
+
+	// A new number of witness_votes to set
+	optional<uint16_t> num_witness;
+
+	// A new number of committee_votes to set
+	optional<uint16_t> num_committee;
+
+	// For future extensions (see account_update_operation)
+	extensions_type extensions;
+};
+```
+
+Changes to direct votes can be specified by means of a *list* of vote identifiers to be added or removed. This has multiple advantages:
+
+* Reduces the size of the operation compared to replacing the entire slate
 * Allows to easily compare votes before and after the operation
 
-Setting the proxy is explicitly not part of this opertation.
-Authors of this BSIP came to the conclusion that proxy settings should
-require `active authority`. It is their opinion that setting proxy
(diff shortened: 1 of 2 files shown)


xeroc:
Questions:

* [ ] Should this operation fail if a proxy is defined?

cogutvalera:
> I think this should better be implemented as extensions on the existing account_update_operation.
> 
> Adding a new operation type for functionality that already exists in some other form only leads to confusion and clutters up the operation namespace.

Absolutely agree with Peter. IMHO extension makes more sense here for additional feature.

xeroc:
> Btw, to my knowledge number 47 (and possibly more) has already been assigned to the proposed STEALTH BSIPs. Please coordinate with core team. @ryanRfox ?

I will



> I think this should better be implemented as extensions on the existing account_update_operation.

The point here is that this operation separates *operational* permissions (voting) from *administrative* permissions (changing keys).
This would, together with BSIP40, even allow to automate voting (e.g. for witnesses) without risking access to the account, or the funds in it.
An extension could surely be used to allow adding/removing of votes without replacing the entire options attribute, but does not decouple operational from administrative permission.

pmconrad:
The distinction sounds artificial to me. I don't see why that would be an advantage.

BSIP-40 states that active authorities can not only be defined for specific operations, but also that the operation parameters can be restricted. If this is implemented properly, it will be possible to create a "voting key" using ``account_update_operation`` that cannot modify authorities.

sschiessl-bcp:
> The distinction sounds artificial to me. I don't see why that would be an advantage.

It is certainly a facilitation for the user since it allows delta voting (easier to check). The big advantage compared to `account_update` is the massive decrease in payload. Account update option are large and must always be fully given, hard to check and lots of payload.

I understand the argument to avoid duplicate logic, but then I'd rather cut the voting out of `account_update`.

pmconrad:
> Advantage compared to account_update is the massive decrease in payload.

Of course. My point is, the same can be achieved using extensions, without introducing a new operation.

> Is the account update options an optional field?

Yes.

xeroc:
I agree with Peter, we should migrate the code over to an extension field of `account_update_operation`. Let's not pollute the operations space

paging @dimfred

pmconrad:
@Dimfred @xeroc @sschiessl-bcp are you still working on this?

Dimfred:
Yes we were. We haven't reviewed it internally yet, but here would be my implementation for now.

https://github.com/blockchainprojects/bitshares-core/tree/feature/account-update-votes-ext

pmconrad:
I'm only talking about this BSIP here, not about the implementation.

If you update the PR please remove the change on README.md to avoid the conflict.

Dimfred:
Ah okay I see. Thought the BSIP is allready out there. Yes Sure we will.

nathanhourt:
Related/possibly redundant with https://github.com/bitshares/bsips/issues/159 (Which includes these changes, but proposes them as an extension to `account_update_operation`)

MichelSantos:
Q: @xeroc @dimfred I merged the two proposals in such a way that the motivation comes from the multiple proxy proposal (#177), and the technical specifications largely come from the explicit voting proposal.  This is one of many possible mergers so please feel free to reject or revise as necessary.

    BSIP:       47
    Title:      Vote Proxies for Different Referendum Categories
    Authors:    Fabian Schuh <https://github.com/xeroc>
                Dmitrij Vinokour <https://github.com/Dimfred>
                Alex Megalokonomos <https://github.com/clockworkgr>
                Michel Santos <https://github.com/MichelSantos>
                Peter Conrad <https://github.com/pmconrad>
    Status:     Draft
    Type:       Protocol
    Created:    2018-09-20
    Discussion: https://github.com/bitshares/bsips/pull/114


# Abstract

Voting power by core token holders may be assigned to different proxies in each of the BitShares referendum categories: witnesses, committee members, and worker proposals.


# Motivation

Some proxies can be more or less knowledgeable, wise, and trusted by token holders to vote on certain referendum categories than others.  It is therefore desirable to empower token holders to select different proxies for each of the referendum categories.

If the ability to select multiple proxies or to directly vote through a single new operation were made possible, this new operation be used in conjunction with [BSIP40](https://github.com/bitshares/bsips/blob/master/bsip-0040.md) which will allow an account holder to assign a specific key for voting.  This combination could simplify voting for many account holders.


# Rationale

Holders of the core token of BitShares (BTS) have always had the ability to directly vote on three referendum categories: witnesses, committee members, and worker proposals.  They have also had the option of delegating their voting power to another account (a "proxy") to vote on any of these referendum categories.

Rather than delegating voting power to a single proxy for each of three referendum, a core token holder might prefer to select different proxies for each of the referendum categories.

This more granular approach empowers core token holders with more options.  A BTS token holder may choose to directly vote on any referendum category including the option to abstain.  A BTS token holder may choose to select proxies for each of the referendum categories.  Each of the referendum categories may have distinct proxies or may have common proxies.  Some referendum categories may have assigned proxies while others may be unassigned.  _Any referendum category that does not have an assigned proxy permits the voter to vote directly on any referendum in that category._  _But if a **confused** core token holder votes for a referendum in a category **while also** delegating that category to a proxy, the token holder's direct vote shall be ignored._

The introduction of a new operation (`account_update_votes_operation`) with additional changes to the core shall also enable an account holder to select proxies for different referendum categories, and to vote for referendums without needing to use the existing `account_update_operation`.  The existing operation requires the account holder to pass all possible account settings to avoid overriding existing settings.  This is inconvinient and requires a higher payload to be stored on the blockchain which results in the account holder paying a higher fee.  The proposed operation shall only update the portion of the `account data` that is related to voting and should also reduce the fee.

The new operation only needs to include

- The voting account
- The desired proxy, if any, for the _committee_ referendum category
- The desired proxy, if any, for the _witness_ referendum category
- The desired proxy, if any, for the _worker_ referendum category
- The quantity of desired witnesses
- The quantity of desired committee members
- The list of referendums that are supported


# Specifications

Whenever a proxy for a category is _assigned_, any direct votes by a core token holder within that category shall be _ignored_.  Whenever a proxy for a category _is unassigned_, any direct votes by a core token holder for that category _shall be counted_ during the vote tallying process of the blockchain.  **This shall require an upgrade to the protocol.**

### `GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT` (New)

A new special account `GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT` with ID `1.2.6` is defined and added to the database at protocol change activation time.


### `account_options` extension (Existing)

**Fields:**

The `account_options` receives three new extensions:

* `optional<account_id_type> committee_voting_account`
* `optional<account_id_type> witness_voting_account`
* `optional<account_id_type> worker_voting_account`

**Validation:**

* These extensions must not be present before the protocol change activation date, neither in transactions nor in proposals.
* If any extension is present, it must contain the ID of an existing account.
* If any extension is present, `account_options.voting_account == GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT`


### `account_object`(Existing)

**Note:** this is an implementation detail and meant to be a hint for later development. The actual implementation may differ.

The account object shall receive these new fields:

* `flat_set<vote_id_type> committee_votes`
* `flat_set<vote_id_type> witness_votes`
* `flat_set<vote_id_type> worker_votes`

`account_create_operation` and `account_update_operation` shall be modified to create these sets by filtering the `account_options.votes` field appropriately. `account_update_votes_operation` shall also create these sets by filtering the `votes_to_add` and `votes_to_remove` fields appropriately.


### `account_update_votes_operation` (New)

The following is an example implementation.

```
struct account_update_votes_operation : public base_operation
{
	asset fee;

	/// The account to update
	account_id_type account;

	/// New account options
	flat_set<vote_id_type>    votes_to_add;
	flat_set<vote_id_type>    votes_to_remove;

	// A new voting account to set
	optional<account_id_type> committee_voting_account
	optional<account_id_type> witness_voting_account
	optional<account_id_type> worker_voting_account

	// A new number of witness_votes to set
	optional<uint16_t> num_witness;

	// A new number of commitee_votes to set
	optional<uint16_t> num_committee;

	// Renew existing voting slate and proxy
	optional<bool> renew_votes;

	// For future extensions (see account_update_operation)
	extensions_type extensions;
};
```

Changes to direct votes can be specified by means of a *list* of vote identifiers to be added or removed. This has multiple advantages:

* Reduces the size of the operation compared to replacing the entire slate
* Allows to easily compare votes before and after the operation

**Validation checks:**

* `account` must be a valid account ID, must exist, must authorize the operation, must pay fee
* `committee_voting_account`, if present, must be a valid account ID, must exist
* `witness_voting_account`, if present, must be a valid account ID, must exist
* `worker_voting_account`, if present, must be a valid account ID, must exist
* Validate the existence of each vote to add and remove
* num_witness &ge; 0
* num_committee &ge; 0
* If the `renew_votes` field is present the other optional fields should be absent _and_ the `votes_to_add` and `votes_to_remove` should be empty

**Evaluation:**

* Update the account's `account_options.committee_votes`, `account_options.witness_votes`, and `account_options.worker_votes` by filtering the `votes_to_add` and `votes_to_remove` field appropriately
* Update the `account_options.num_committee` if `account_options.committee_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT`
* Update the `account_options.num_witness` if `account_options.witness_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT`
* Update the account's last vote time to the head block time


### `account_create_operation` (Existing)

This operation shall also be updated for backwards compatibility.  If `account_options.voting_account == GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT`, then the account's per-category voting proxies are set to the specified extension if present, or to `GRAPHENE_PROXY_TO_SELF_ACCOUNT` if absent.
Otherwise, the voting proxy for all three categories is set to `account_options.voting_account`.


### `account_update_operation`(Existing)

This operation shall also be updated for backwards compatibility.  If `account_options.voting_account == GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT`, then the account's per-category voting proxies are set to the specified extension if present, or remain unchanged if absent.
Otherwise, the voting proxy for all three categories is set to `account_options.voting_account`.


### Vote Tallying

The vote tallying algorithm shall be modified to select the `opinion_account` per voting category, and use the current account's voting stake only for the opinion account's votes in the respective category.

Other BSIPs that may scale the voting weight of an account, such as by [vote decay](https://github.com/bitshares/bsips/bsip-0022.md) or by [vote staking](https://github.com/bitshares/bsips/bsip-0024.md), can remain compatible with this BSIP by assigning the scaled voting weight to the different voting proxies that may or may not be selected by an account.


# Discussion

### Performance Impact

There will be a performance impact by `account_create_operation`, `account_update_operation`, and `account_update_votes_operation` due to the filtering if implemented as suggested. This impact should be small overall, because these operations are comparatively rare.

There will also be some overhead for vote tallying. Vote tallying happens only during the maintenance interval so the operational impact should be negligible as well.

Note that the cumulative effect of both may be noticable during chain replay.

# Summary for Shareholders

This proposal introduces a means for token holders to select different proxies in each of the BitShares referendum categories: witnesses, committee members, and worker proposals.  The proposed mechanism in combination with [BSIP40](https://github.com/bitshares/bsips/blob/master/bsip-0040.md) will permit core token holders to simplify automated voting without compromising the account through the use of a new operation that can be used only for voting.


# Copyright

# See Also

- [Vote Proxies for Different Referendum Categories](https://github.com/bitshares/bsips/issues/177)
- [Proposed details for BSIP-22 specifications](https://github.com/bitshares/bsips/pull/153)

MichelSantos:
@xeroc @Dimfred Some of the merger merits your review.

1. I removed "Setting the proxy is explicitly not part of this opertation.  Authors of this BSIP came to the conclusion that proxy settings should require `active authority`. It is the opinion that setting proxy settings are distinct from voting."  in consideration of the merger.  Is this acceptable to both of you?

2. There is a new specification that might conflict with your anticipated uses for BSIP40 for automated voting?

> "`account` must be a valid account ID, must exist, must authorize the operation, must pay fee" 

This could maybe be relaxed or changed depending on your intended usage

MichelSantos:
@pmconrad Regarding

> "There will be a performance impact on `account_create`, `account_update`, and `account_update_votes` due to the filtering (if implemented as suggested). This impact should be small overall, because these operations are comparatively rare."

If this is a concern for the `account_update_votes` operation, we could change

```
/// New account options
flat_set<vote_id_type>    votes_to_add;
flat_set<vote_id_type>    votes_to_remove;
```

to be

```
/// New account options
flat_set<vote_id_type>    committee_votes_to_add;
flat_set<vote_id_type>    committee_votes_to_remove;
flat_set<vote_id_type>    witness_votes_to_add;
flat_set<vote_id_type>    witness_votes_to_remove;
flat_set<vote_id_type>    worker_votes_to_add;
flat_set<vote_id_type>    worker_votes_to_remove;
```

This will address the performance issue but will add a small amount of complexity to client implementations.

MichelSantos:
Another question is how should the renewal of a voting slate or the renewal of a proxy be handled?  This question is relevant to the [vote decay proposal](https://github.com/bitshares/bsips/pull/153) where the latest vote time must be updated.  The [current logic](https://github.com/bitshares/bitshares-core/blob/af8a01aa9a791072f7f9180b34dc69bd6acde114/libraries/chain/account_evaluator.cpp#L312-L314) only updates the time if the voting slate or the proxy **changes**.

I think that the best two options are:

- for a voting slate to be changed in a meaningless way in one operation, and  to then be reset in a second operation
- to include an explicit "renew" field in the `account_update_votes_operation`

I chose the latter option by adding a field called `renew_votes` into `account_update_votes_operation` which can be set to `true` for this purpose.

pmconrad:
> > This impact should be small overall, because these operations are comparatively rare.
> If this is a concern for the account_update_votes operation,

It is not, because the operation is rare.

> how should the renewal of a voting slate or the renewal of a proxy be handled?

The proposed "renew" field is redundant - `account_update_votes_operation` implies "renew", IMO. No actual changes need to be contained in the operation, its presence alone would be sufficient to renew the timer.

MichelSantos:
> > how should the renewal of a voting slate or the renewal of a proxy be handled?
> 
> The proposed "renew" field is redundant - `account_update_votes_operation` implies "renew", IMO. No actual changes need to be contained in the operation, its presence alone would be sufficient to renew the timer.

I understand now.  The new evaluator for the `account_update_votes_operation` will not contain the same filtering logic as is found in the [account_update_evaluator.do_apply()](https://github.com/bitshares/bitshares-core/blob/af8a01aa9a791072f7f9180b34dc69bd6acde114/libraries/chain/account_evaluator.cpp#L312-L314).

The proposed text now reflects the removal of this field.

xeroc:
I updated the pull request/commit to use @MichelSantos 's text. I like it!

pmconrad:
> I updated the pull request/commit to use @MichelSantos 's text. I like it!

@xeroc please double-check. I can't see your update to the proposal text.

xeroc:
cleaned up. sorry for the delay

bsip-0047.md
```diff
@@ -0,0 +1,182 @@
...
+* Allows to easily compare votes before and after the operation
+
+**Validation checks:**
+
+* `account` must be a valid account ID, must exist, must authorize the operation, must pay fee
+* `committee_voting_account`, if present, must be a valid account ID, must exist
+* `witness_voting_account`, if present, must be a valid account ID, must exist
+* `worker_voting_account`, if present, must be a valid account ID, must exist
+* Validate the existence of each vote to add and remove
+* `num_witness`, if present, must be greater than or equal to  &ge; 0
```
pmconrad: ...and must not be greater than the resulting list of witness votes.

bsip-0047.md
```diff
@@ -0,0 +1,182 @@
...
+* `num_committee`, if present, must be greater than or equal to  &ge; 0
```
pmconrad: ...and must not be greater than the resulting list of committee votes.

bsip-0047.md
```diff
@@ -0,0 +1,182 @@
...
+
+**Evaluation:**
+
+* Update the account's `account_options.committee_votes`, `account_options.witness_votes`, and `account_options.worker_votes` by filtering the `votes_to_add` and `votes_to_remove` field appropriately
+* Update the `account_options.num_witness` if `witness_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` and if `num_witness` is present
+* Update the `account_options.num_committee` if `committee_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` and if `num_committee` is present
+* Update the account's last vote time to the head block time
+
```
pmconrad: * If any of `committee_voting_account`, `witness_voting_account` or `worker_voting_account` is present
    * update the corresponding field in the account's `account_options`
    * set the account's `account_options.voting_account` to `GRAPHENE_PROXY_PER_CATEGORY_ACCOUNT`

bsip-0047.md
```diff
@@ -0,0 +1,182 @@
...
+* `worker_voting_account`, if present, must be a valid account ID, must exist
+* Validate the existence of each vote to add and remove
+* `num_witness`, if present, must be greater than or equal to  &ge; 0
+* `num_committee`, if present, must be greater than or equal to  &ge; 0
+
+**Evaluation:**
+
+* Update the account's `account_options.committee_votes`, `account_options.witness_votes`, and `account_options.worker_votes` by filtering the `votes_to_add` and `votes_to_remove` field appropriately
+* Update the `account_options.num_witness` if `witness_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` and if `num_witness` is present
+* Update the `account_options.num_committee` if `committee_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` and if `num_committee` is present
```
pmconrad: If `votes_to_add` and `votes_to_remove` are accepted regardless of `GRAPHENE_PROXY_TO_SELF_ACCOUNT`, then `num_witness` and `num_committee` should be, too.

bsip-0047.md
pmconrad: * If not present, and witness votes are removed, the resulting list of witness votes must not be smaller than the current `num_witness` setting of the account.

bsip-0047.md
pmconrad: * If not present, and committee votes are removed, the resulting list of committee votes must not be smaller than the current `num_committee` setting of the account.

bsip-0047.md
pmconrad: I think you misunderstood me. That was not meant to be a new rule in the list, it was a comment hinting at an inconsistency in the existing list. Sorry for being inconsistent with my comment style. :-)

What I meant was: 
In step 1 you update the account's `*_votes` fields unconditionally. In steps 2 and 3 you update `num_witness` and `num_committee` only "if `witness_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT`.
You should either apply the condition to all steps or to none (IMO none is better).

MichelSantos: @pmconrad I did misunderstand so I appreciate the clarification.

I do think that we do need some form of the following checks:

* Update committee votes if and only if `account_options.witness_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT` **after** the merging of this operation
	* `committee_voting_account` may or may not be specified in the operation
* Update witness votes if and only if `account_options.witness_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT` **after** the merging of this operation
	* `witness_voting_account` may or may not be specified in the operation
* Update worker votes if and only if `account_options.worker_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT` **after** the merging of this operation
	* `worker_voting_account` may or may not be specified in the operation
* Update the `account_options.num_witness` if `num_witness` is present and if `account_options.witness_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` **after** the merging of this operation
* Update the `account_options.num_committee` if `num_committee` is present and if `account_options.committee_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT`  **after** the merging of this operation

Take for example, 

* Update the `account_options.num_witness` if `num_witness` is present and if `account_options.witness_voting_account = GRAPHENE_PROXY_TO_SELF_ACCOUNT` **after** the effect of this operation

The trickier use cases are ones where one operation might update `num_witness` **or**  `witness_voting_account` **but not both** and then _merge_ with its counterpart that is missing in the operation but is already present in the account_options on the blockchain.  This type of check is essential but could be performed either in validation or evaluation depending on your recommendation.

pmconrad: Ok, then you should add these constraints to the validation section. I. e. operations with these inconsistencies are not allowed into the blockchain.

MichelSantos: The updated validation specs attempt to cover the proper merging of the changes with the current values on the blockchain.

The updated and previous validation also permit an operation that has nothing of substance (an "empty" update).  This might be appropriate if the extensions field might be used.  Such an "empty" operation will have the practical effect of updating the account's last vote time.  We should add an extra validation if "empty" updates are to be prohibited.

ryanRfox:
It feels like this is very close to "done". @pmconrad are you awaiting anything further from @MichelSantos prior to completing your review? @xeroc are you agreeing with the modifications made thus far?

bsip-0047.md
```diff
@@ -0,0 +1,230 @@
...
+_General validation checks_
+* `account` must be a valid account ID, must exist, must authorize the operation, must pay fee
+* `committee_voting_account`, if present, must be a valid account ID, must exist
+* `witness_voting_account`, if present, must be a valid account ID, must exist
+* `worker_voting_account`, if present, must be a valid account ID, must exist
+* Validate the existence of each vote to add and remove
+
+_Committee validation checks_
+* Let the new _committee_ slate consist of the current _commitee_ slate plus the _committee_ `votes_to_add` minus the _committee_ `votes_to_remove`
+* If `account_options.committee_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT`
```
pmconrad: I think what you mean is
* If committee_voting_account is present, let new_committee_account = committee_voting_account
* otherwise let new_committee_account = account_options.committee_voting_account

bsip-0047.md
```diff
@@ -0,0 +1,230 @@
...
+	* Let `new_committee_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`
+* otherwise
+	* Let `new_committee_account` = `account_options.committee_account`
+* If `num_committee` is present
+	* Let `new_num_committee` = `num_committee`
+* otherwise
+	* Let `new_num_committee` = `account_options.num_committee`
+* `num_committee`, if present, is valid:
+	* if `new_committee_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`, and
+	* if 0 &le; `num_committee` &le; `new_num_committee`
```
pmconrad: I think "size of new _committee_ slate" instead of "new_num_committee"

pmconrad: ...but that's redundant with the check below.

bsip-0047.md
```diff
@@ -0,0 +1,230 @@
...
+* The _committee_ `votes_to_add` and `votes_to_remove` are valid:
+	* if `new_committee_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`, and
+	* if 0 &le; `new_num_committee` &le; size of new _committee_ slate &le; `chain_params.maximum_committee_count`
+
+_Witness Validation Checks_
+* Let the new _witness_ slate consist of the current _witness_ slate plus the _witness_ `votes_to_add` minus the _witness_ `votes_to_remove`
+* If `account_options.witness_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT`
```
pmconrad: I think what you mean is
* If witness_voting_account is present, let new_witness_account = witness_voting_account
* otherwise let new_witness_account = account_options.witness_voting_account

bsip-0047.md
```diff
@@ -0,0 +1,230 @@
...
+	* Let `new_witness_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`
+* otherwise
+	* Let `new_witness_account` = `account_options.witness_account`
+* If `num_witness` is present
+	* Let `new_num_witness` = `num_witness`
+* otherwise
+	* Let `new_num_witness` = `account_options.num_witness`
+* `num_witness`, if present, is valid:
+	* if `new_witness_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`, and
+	* if 0 &le; `num_witness` &le; `new_num_witness`
```
pmconrad: This doesn't make sense. num_witness is either not present or equal to new_num_witness. Also see comments for similar line in committee validation.

bsip-0047.md
```diff
@@ -0,0 +1,230 @@
...
+* The _witness_ `votes_to_add` and `votes_to_remove` are valid:
+	* if `new_witness_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`, and
+	* if 0 &le; `new_num_witness` &le; size of new _witness_ slate &le; `chain_params.maximum_witness_count`
+
+_Worker Validation Checks_
+* If `account_options.worker_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT`
```
pmconrad: See above, I think you mean something else.

bsip-0047.md
```diff
@@ -0,0 +1,230 @@
...
+# Abstract
+
+Voting power by core token holders may be assigned to different proxies in each of the BitShares referendum categories: witnesses, committee members, and worker proposals.
+
+
+# Motivation
+
+Some proxies can be more or less knowledgeable, wise, and trusted by token holders to vote on certain referendum categories than others.  It is therefore desirable to empower token holders to select different proxies for each of the referendum categories.
+
+If the ability to select multiple proxies or to directly vote through a single new operation were made possible, this new operation be used in conjunction with [BSIP40](https://github.com/bitshares/bsips/blob/master/bsip-0040.md) which will allow an account holder to assign a specific key for voting.  This combination could simplify voting for many account holders.
```
pmconrad: Missing word: ...this new operation *could* be used...

bsip-0047.md
```diff
@@ -0,0 +1,230 @@
...
+
+
+# Rationale
+
+Holders of the core token of BitShares (BTS) have always had the ability to directly vote on three referendum categories: witnesses, committee members, and worker proposals.  They have also had the option of delegating their voting power to another account (a "proxy") to vote on any of these referendum categories.
+
+Rather than delegating voting power to a single proxy for each of three referendum, a core token holder might prefer to select different proxies for each of the referendum categories.
```
pmconrad: Perhaps "single proxy for *all* three referendum *categories*" ?

bsip-0047.md
```diff
@@ -0,0 +1,230 @@
...
+
+The introduction of a new operation (`account_update_votes_operation`) with additional changes to the core shall also enable an account holder to select proxies for different referendum categories, and to vote for referendums without needing to use the existing `account_update_operation`.  The existing operation requires the account holder to pass all possible account settings to avoid overriding existing settings.  This is inconvinient and requires a higher payload to be stored on the blockchain which results in the account holder paying a higher fee.  The proposed operation shall only update the portion of the `account data` that is related to voting and should also reduce the fee.
```
pmconrad: Typo: inconv*e*~~i~~nient

bsip-0047.md
MichelSantos: Indeed, thank you.  I'd like to say that I was thinking about some corner case but it just looks wrong now.  The text has been updated.

bsip-0047.md
MichelSantos: I see your point.  The updated text combines the validation of this field with the redundant validation below it.

bsip-0047.md
```diff
@@ -131,42 +131,36 @@ _General validation checks_
...
 
 _Committee validation checks_
-* Let the new _committee_ slate consist of the current _commitee_ slate plus the _committee_ `votes_to_add` minus the _committee_ `votes_to_remove`
-* If `account_options.committee_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT`
-	* Let `new_committee_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`
+* Let the new committee slate consist of the current commitee slate plus the committee `votes_to_add` minus the committee `votes_to_remove`
+* If `committee_voting_account` is present
+	* Let `new_committee_account` = `committee_voting_account`
 * otherwise
 	* Let `new_committee_account` = `account_options.committee_account`
```
pmconrad: It's `account_options.committee_voting_account` I think.

bsip-0047.md
```diff
@@ -131,42 +131,36 @@ _General validation checks_
...
 * If `num_committee` is present
 	* Let `new_num_committee` = `num_committee`
 * otherwise
 	* Let `new_num_committee` = `account_options.num_committee`
-* `num_committee`, if present, is valid:
+* The _committee_ `votes_to_add`,`votes_to_remove`, and `num_committee`, if present, are valid:
```
pmconrad: I suppose you mean "valid if present" in the sense of "only allowed when ..."?

bsip-0047.md
```diff
@@ -131,42 +131,36 @@ _General validation checks_
...
 
 _Witness Validation Checks_
-* Let the new _witness_ slate consist of the current _witness_ slate plus the _witness_ `votes_to_add` minus the _witness_ `votes_to_remove`
-* If `account_options.witness_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT`
-	* Let `new_witness_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`
+* Let the new witness slate consist of the current witness slate plus the witness `votes_to_add` minus the witness `votes_to_remove`
+* If `witness_voting_account` is present
+	* Let `new_witness_account` = `witness_voting_account`
 * otherwise
 	* Let `new_witness_account` = `account_options.witness_account`
```
pmconrad: `account_options.witness_voting_account`

bsip-0047.md
```diff
@@ -131,42 +131,36 @@ _General validation checks_
...
-	* if 0 &le; `new_num_witness` &le; size of new _witness_ slate &le; `chain_params.maximum_witness_count`
+	* if 0 &le; `new_num_witness` &le; size of new witness slate &le; `chain_params.maximum_witness_count`
 
 _Worker Validation Checks_
-* If `account_options.worker_voting_account == GRAPHENE_PROXY_TO_SELF_ACCOUNT`
-	* Let `new_worker_account` = `GRAPHENE_PROXY_TO_SELF_ACCOUNT`
+* If `worker_voting_account` is present
+	* Let `new_worker_account` = `worker_voting_account`
 * otherwise
 	* Let `new_worker_account` = `account_options.witness_account`
```
pmconrad: `account_options.worker_voting_account` (note worker not witness!)

bsip-0047.md
MichelSantos: I agree that "only allowed" is clearer.  I have updated the text to reflect that `num_X` and `votes_to_add` and `votes_to_remove` have their own separate validations in an attempt to be clear to the reader especially because `num_X` _may or may not_ be present, and `votes_to_add` and `votes_to_remove` _may or may not_ be non-empty.

The entire validation code can be written in a more computational efficient manner to reflect this intent.

sschiessl-bcp:
Adjusted the Readme, and added the explicit voting operation in title. Please re-approve.

PR #115, opened 2018-10-15
BSIP 48

xeroc:
This is a new proposal that adds a flag to the asset permissions.
This flag allows to opt out of "changing the max supply".

### DIFF
--- a/bsip-0048.md
+++ b/bsip-0048.md
@@ -1,0 +1,55 @@
+    BSIP:       <BSIP number>
+    Title:      Add flag to prevent asset's max supply from being changed
+    Authors:    Fabian Schuh <https://github.com/xeroc>
+                Dmitrij Vinokour <https://github.com/Dimfred>
+    Status:     Draft
+    Type:       Protocol
+    Created:    2018-10-14
+    Discussion: <url>
+    Worker:     <Id of worker proposal> (optional)
+
+# Abstract
+
+The possibility of creating assets on the BitShares Blockchain is a core feature
+of our platform. In the case of an ico, the ico-holders can create an asset with
+a fixed maximum supply. However this maximum supply can be changed by the
+asset creator (issuer) anytime. This exposes participators ico to the risk,
+of loosing value of their bought asset.
+
+With this BSIP a flag will be introduced which forbids the modification of the maximum
+supply of an issuer created asset.
+
+# Motivation
+
+The implementation of a flag, to disable the modification of the maximum supply for 
+issuer created asset, has effects for both sides. On the one side the security for the
+participators or non-issuer users increases, because the value of their assets (tokens)
+can't be manipulated by exposing the market to more created tokens. And on the other side
+it is good marketing for the ico-hoders, because they can advertise with their fixed
+maximum supply.
+
+# Rational
+
+The `disable_modify_max_supply` is a new created flag in the `asset_issuer_permissions`.
+The flag can be activated to forbid the modification of the `max_supply` (`asset_object.options`).
+Once the flag is activated, it can't be deactivated.
+
+# Specifications
+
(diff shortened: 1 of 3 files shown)


bsip-0048.md
```diff
@@ -0,0 +1,55 @@
...
+participators or non-issuer users increases, because the value of their assets (tokens)
+can't be manipulated by exposing the market to more created tokens. And on the other side
+it is good marketing for the ico-hoders, because they can advertise with their fixed
+maximum supply.
+
+# Rational
+
+The `disable_modify_max_supply` is a new created flag in the `asset_issuer_permissions`.
+The flag can be activated to forbid the modification of the `max_supply` (`asset_object.options`).
+Once the flag is activated, it can't be deactivated.
```
abitmore: If it can't be deactivated, why not set it as an issuer_permission?

bsip-0048.md
```diff
@@ -0,0 +1,55 @@
...
+# Specifications
+
+This BSIP comes with a change to the protocol, which requires a propotocol upgrade.
+However the changes to the code are minor.
+
+1. Add the `disable_modify_max_supply` flag to the issuer permissions
+2. Add a `can_modify_max_supply()` the `asset_object`
+3. Change the `asset_update_evaluator::evaluate` method to assert false for
+    1. the deactivation of the `disable_modify_max_supply` flag
+    2. the modification of the `max_supply` when the `disable_modify_max_supply` is set
```
abitmore: How about existing assets?

xeroc: Good call, I would recommend the permission is enabled, as is the flag.
That way, behavior isn't changed.
The feature is supposed to be opt-out afterall.

What needs to be done to ensure that for existing assets?

bsip-0048.md
pmconrad: Note: permissions can be modified when current_supply = 0. Makes sense also in this case IMO.

bsip-0048.md
```diff
@@ -0,0 +1,55 @@
...
+issuer created asset, has effects for both sides. On the one side the security for the
+participators or non-issuer users increases, because the value of their assets (tokens)
+can't be manipulated by exposing the market to more created tokens. And on the other side
+it is good marketing for the ico-hoders, because they can advertise with their fixed
+maximum supply.
+
+# Rational
+
+The `disable_modify_max_supply` is a new created flag in the `asset_issuer_permissions`.
+The flag can be activated to forbid the modification of the `max_supply` (`asset_object.options`).
```
pmconrad: IMO the flag is easier to understand if it is called `allow_modify_max_supply`, i. e. the meaning is inverted.

pmconrad:
Btw, number 48 may have already been assigned to STEALTH, please coordinate. @ryanRfox ?

bsip-0048.md
Dimfred: @pmconrad Indeed, I will add that behaviour.

bsip-0048.md
Dimfred: @pmconrad That was our first thought. But every other ìssuer_permissions` are also defined as disabled. Also if you for example create a new asset and forget to set the flag, the asset will be created with fixed supply by default.  But in this case there won't be any `current_supply` if understand it right. So with the above mentioned logic the issuer would still be able to modify  the `max_supply` but not be able to reset the flag (or in the `allow_modify_max_supply`case to enable it. Maybe it  would also make sense to be able to reset the `disable_modify_max_supply` , when there is no `current_supply`.

ryanRfox:
Any progress toward advancing this?

bsip-0048.md
abitmore: Discussion in https://github.com/bitshares/bitshares-core/pull/1375#pullrequestreview-164580457 suggests the opposite.

Issue #123, opened 2018-11-09
BSIP - Market Fee Backed Asset (MFBA)


OpenLedgerApp:
## Abstract
Market Fee Backed Asset (MFBA)  is a SmartCoin introduced by this BSIP. 
The main idea of a MFBA is to provide opportunity for the Assets Owners to automatically share revenue received from trading operations (Market Fees) to the holders of the MFBA in proportion of the owned amount of the MFBA.

## Motivation
By implementing this proposals, we let Asset Owners an opportunity to use their assets as profit-sharing assets.
Distribution of fees will be conducted automatically (in a pre-defined period of time). Such functionality is currently not available in the blockchain.

## Specifications
**Market Fee Backed Asset  Definition**
New attribute of the SmartCoin - "MFBA" attribute becomes available for the users. Asset for which "MFBA" option is available will be called MFBA Asset. 

**Tied assets**
Issuer of the MFBA Asset will be able to tie his/her other assets (one or many) to the MFBA Asset. Tied assets cannot become MFBA.

**Dividends**
Market Fee, received from the trading operations of the tied assets will be distributed between holders of the MFBA Asset according to the stake of the MFBA Asset that holders have. 
Issuer of the MFBA Asset must be able to set share of the Market Fee, that will be distributed to the holders of the MFBA Asset as a parameter (MFBA share, %).
Whatever is left after distribution to the MFBA holders will be left for the Asset Owner. 
Funds currently in the open orders will not be included in the MFBA share.

**Blacklist and Whitelist**
Whitelist and blacklist (configurable by the UIA Issuer) could provide control over who is eligible to earn dividends.

**Dividends Distribution Process**
Distribution will be performed in a set period of time. Period of time must be set by the MFBA Issuer (minimal available interval must exist).
Market Fee of the tied assets is accumulated during this period. At the end of the period fee is distributed to the MFBA holders' "special balances" and to the MFBA Issuer "special balance". 
After that MFBA holders and MFBA Issuer can claim the fees.

**Dividends Distribution Process Restrictions**
Distribution must be performed starting from the user, that has the biggest share of the MFBA Asset.
Users for which amount to be transferred is less, than should be accrued, will not receive their share of Market Fee.

## CORE TEAM TASK LIST
- [ ] Evaluate / Prioritize Feature Request
- [ ] Refine User Stories / Requirements
- [ ] Define Test Cases
- [ ] Design / Develop Solution
- [ ] Perform QA/Testing
- [ ] Update Documentation

[...]

PR #132, opened 2018-12-17
draft

OpenLedgerApp:
Dear BitShares Community,

We would like to introduce the **sharedrop operation BSIP**. 
The purpose is to share potential profit/money/other assets with the holders of the specific asset according with their stake. 

Please find more details here: 
https://github.com/openledger/bsips/blob/bsip-sharedropoperation/bsip-00XX%20Sharedrop%20Operation.md

Sincerely,
Denis Sokolov
OpenLedger

### DIFF
--- a/bsip-00XX Sharedrop Operation.md
+++ b/bsip-00XX Sharedrop Operation.md
@@ -1,0 +1,153 @@
+    BSIP: TBD
+    Title: Sharedrop operation
+    Authors: OpenLedgerApp <https://github.com/OpenLedgerApp>
+    Status: Draft
+    Type: Protocol
+    Created: 2018-12-17
+    Updated: 2018-12-17
+    Discussion: 
+    Worker: TBD
+ 
+A sharedrop is a giveaway of some tokens to a certain group of people, for example to all BTS holders in proportion to their holdings.
+
+# Abstract
+We should add the ability to distribute arbitrary amount of some token (SHAREDROP token) to holders of a specific other token (STAKE token)  according to the share(stake) of such tokens.
+The importance of a similar solution was raised in BSIPs-19, 20,  however never implemented. (link below).
+
+
+# Motivation
+By implementing this proposal, we let Asset Owners or other businesses an opportunity to distribute their profit to the holders of a specific asset in proportion of the owned amount of this asset.
+
+Also, usual token drops can be done via the sharedrop operation.
+
+# Rationale
+
+- Other cryptocurrency platforms offer profit-sharing/dividends functionality, such as Peerplays/NXT/CounterParty/DigixDAO/LBRY/Waves/Dash.
+- Most (if not all) Bitcoin forks (1000's of them) have 'sendmany' functionality which enables the mass distribution of their tokens against a large quantity of users (addresses) in a single transaction; this is currently not possible within BitShares.
+
+To make it more difficult to abuse the system dropping multiple useless tokens, there will be a vesting stake for the sharedrop operation. For example, vest 5000 bitUSD (in BTS) for 2 month to be able to do sharedrop operation. 
+
+Sharedrop operation should also be quite expensive as well. 
+However, to make such an operation more lucrative for BtiShares - the sharedrop operation will send the money to each stakeholder's vesting balance. Thus making them withdraw the drop from the vesting balances.
+
+ 
+# Specifications
+## General workflow
+First of all, there is a need to calculate transaction fee, so **dry_run_sharedrop** operation should be called on a local node to do calculations.  THe results will be provided (displayed) to a user. 
+
+Then, knowing that user has enough BTS for transaction fees and vesting money, he should call Sharedrop operation itself. Vesting specifics will be covered below. 
+
+Sharedrop operation will drop the Sharedrop asset in proportion to the stake that stakeholders have to their respective vesting balances.
+
+## dry_run_sharedrop operation 
+This is a locally done operation with the following parameters: 
+
+- Sharedrop asset
+- Sharedrop amount
+- Stake asset
+- Minimum stake asset amount to participate in airdrop (**minimum_amount_for_drop**)
+dry_run_sharedrop operation returns transaction fee for the sharedrop operation as well as if a user has enough money in his sharedrop vesting or if he should prepare additonal funds.
+
+This will be approximate calculation as the final calculation will be determined only during the sharedrop operation.
+
+The following parameters are calculated by a node and presented to the user:  
+
+- Sharedrop operation price
+- Sharedrop vesting stake 
+- Sharedrop vesting period
+
+## Sharedrop operation
+Sharedrop operation should be available with the following parameters:
+
+- Sharedrop asset
+- Sharedrop amount
+- Stake asset
+Minimum stake asset amount to participate in airdrop (**minimum_amount_for_drop**)
+Sharedrop operation will drop the Sharedrop asset in proportion to the stake that stakeholders have to their respective vesting balances.
+
+Each stakeholder will then need to retrieve the sharedrop token from the vesting balance in a separate operation.
+
+If the amount to distribute for stakeholders is too small and cannot be distributed among stakehoders or part of stakeholders then it accumulates and is returned to the account that originated sharedrop operation.
+
+## Sharedrop Parameters 
+
+### Minimum amount to participate in sharedrop.
+If an asset holder has too little amount of the stake token it becomes expensive for the blockchain to do sharedrop for such a user.
+So a sharedrop operation should have **minimum_amount_for_drop** property. 
+
+### Sharedrop price
+sharedrop price should be determined based on the number of accounts (stakeholders) to split the value for, as in `stakeholder_number * sharedrop_price_multiplier`, where **sharedrop_price_multiplier** is defined by the committee.
+
+## Sharedrop vesting stake 
+Sharedrop vesting stake will be a flat fee determined by the committee. It will be charged in BTS.
+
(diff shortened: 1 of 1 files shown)


bsip-00XX Sharedrop Operation.md
```diff
@@ -0,0 +1,153 @@
...
+    Created: 2018-12-17
+    Updated: 2018-12-17
+    Discussion: 
+    Worker: TBD
+ 
+A sharedrop is a giveaway of some tokens to a certain group of people, for example to all BTS holders in proportion to their holdings.
+
+# Abstract
+We should add the ability to distribute arbitrary amount of some token (SHAREDROP token) to holders of a specific other token (STAKE token)  according to the share(stake) of such tokens.
+The importance of a similar solution was raised in BSIPs-19, 20,  however never implemented. (link below).
```
xeroc: This is pretty much the same thing as paying dividends.
The dividends feature has been implemented already for the PeerPlays blockchain.
See here: https://github.com/bitshares/bitshares-core/issues/695

bsip-00XX Sharedrop Operation.md
```diff
@@ -0,0 +1,153 @@
...
+
+# Motivation
+By implementing this proposal, we let Asset Owners or other businesses an opportunity to distribute their profit to the holders of a specific asset in proportion of the owned amount of this asset.
+
+Also, usual token drops can be done via the sharedrop operation.
+
+# Rationale
+
+- Other cryptocurrency platforms offer profit-sharing/dividends functionality, such as Peerplays/NXT/CounterParty/DigixDAO/LBRY/Waves/Dash.
+- Most (if not all) Bitcoin forks (1000's of them) have 'sendmany' functionality which enables the mass distribution of their tokens against a large quantity of users (addresses) in a single transaction; this is currently not possible within BitShares.
```
xeroc: BitShares has a "sendmany" too. It is just a transaction with "many" transfer operations. Same thing that bitcoin forks do.

bsip-00XX Sharedrop Operation.md
```diff
@@ -0,0 +1,153 @@
...
+
+To make it more difficult to abuse the system dropping multiple useless tokens, there will be a vesting stake for the sharedrop operation. For example, vest 5000 bitUSD (in BTS) for 2 month to be able to do sharedrop operation.
```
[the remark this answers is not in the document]
OpenLedgerApp: That's just an idea. It's up for discussion. Our take was that if we make it let's say 50 bitUSD, then it's still open for shitcoins to send airdrops.

bsip-00XX Sharedrop Operation.md
OpenLedgerApp: We have looked at the way "sendmany" is implemented and it does not exactly cover our needs. It is like many different transfers.. Block can only have limited amount of transfers. So if there are too many, then we will have several transactions and it will not be atomic.

bsip-00XX Sharedrop Operation.md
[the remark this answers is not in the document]
OpenLedgerApp: Actually, we thought some more about it.. And we can definitely scrap the idea of vesting stake.
AND/OR  
We might change the transaction fee for the sharedrop not to depend on the number of users. This way we would not need to calculate the fee. And it will be much simpler and easier to implement.  
LET'S HAVE A DISCUSSION HERE GUYS! 
Or we can have it in the forum
https://bitsharestalk.org/index.php?topic=27588.0

bsip-00XX Sharedrop Operation.md
```diff
@@ -0,0 +1,153 @@
...
+
+To make it more difficult to abuse the system dropping multiple useless tokens, there will be a vesting stake for the sharedrop operation. For example, vest 5000 bitUSD (in BTS) for 2 month to be able to do sharedrop operation. 
+
+Sharedrop operation should also be quite expensive as well. 
+However, to make such an operation more lucrative for BtiShares - the sharedrop operation will send the money to each stakeholder's vesting balance. Thus making them withdraw the drop from the vesting balances.
+
+ 
+# Specifications
+## General workflow
+First of all, there is a need to calculate transaction fee, so **dry_run_sharedrop** operation should be called on a local node to do calculations.  THe results will be provided (displayed) to a user.
```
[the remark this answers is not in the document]
OpenLedgerApp: pls see below

bsip-00XX Sharedrop Operation.md
```diff
@@ -0,0 +1,153 @@
...
+ 
+# Specifications
+## General workflow
+First of all, there is a need to calculate transaction fee, so **dry_run_sharedrop** operation should be called on a local node to do calculations.  THe results will be provided (displayed) to a user. 
+
+Then, knowing that user has enough BTS for transaction fees and vesting money, he should call Sharedrop operation itself. Vesting specifics will be covered below. 
+
+Sharedrop operation will drop the Sharedrop asset in proportion to the stake that stakeholders have to their respective vesting balances.
+
+## dry_run_sharedrop operation
```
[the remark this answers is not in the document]
OpenLedgerApp: Please see above. If we change the fee to the FIXED fee then we do not need it at all. 
if we still have the fee depend on the number of users, then we still need to calculate it. 
In this case, we can definitely extend the ``` get_required_fees ``` for this operation and implement our calculations there.
Thank you for the idea.
Again, it depends on the discussion about the fees and vesting stake.

bsip-00XX Sharedrop Operation.md
OpenLedgerApp: As I mentioned, the idea is very similar to bsip-20 (and 19).  
We have spent quite a lot of time looking into the PeerPlay code. And at this point of time (2 years after the code has been written) we have concluded that it is easier, safer and more maintainable in the future to write our own code. 
Please see my extensive reply here: 
https://github.com/bitshares/bsips/issues/123#issuecomment-444119802
Thanks,
Denis

OpenLedgerApp: We have just posted related BSIP  - Market Fee Backed Asset (MFBA) - #134 
and here:
https://bitsharestalk.org/index.php?topic=27601.0
The idea is quite similar, but it's more automated with MFBA. 
And it's more flexible with Sharedrop. 
We can have one or the other or both. 
Code base, when it comes to splitting share with stakeholders to their vesting balances would be quite similar.

ryanRfox:

bsip-00XX Sharedrop Operation.md
sschiessl-bcp: The same will happen with an actual `send_many` operation. The thing that will make the transaction too big will be the `account_id` to `amount` list, and not the header of the serialization.

bsip-00XX Sharedrop Operation.md
sschiessl-bcp: What is the difference to BSIP 19/20 in essence?

bsip-00XX Sharedrop Operation.md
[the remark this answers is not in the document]
sschiessl-bcp: What is the idea which vesting balance object to use?

If the user does not have a vesting balance object, it will need to be created inducing, i.e. more fees. IMO it can't be a fixed fee in such a case

bsip-00XX Sharedrop Operation.md
```diff
@@ -0,0 +1,153 @@
...
+
+2. One of the ideas is instead of making flat fee vesting for all sharedrop transactions, for each sharedrop transaction we put into vesting an amount based on the stakeholder number. And we have to put a new amount for every sharedrop transaction.
+In this case, it might quickly become too expensive for some users.
+
+
+# See Also
+
+https://github.com/bitshares/bsips/blob/master/bsip-0020.md
+
+## Copyright
```
sschiessl-bcp: Please add a easy-to-understand Summary for the Shareholders

bsip-00XX Sharedrop Operation.md
pmconrad: A serialized `transfer_operation` in BitShares is much smaller than a bitcoin address. A BitShares transaction with many transfers will be significantly smaller than a bitcoin transaction with as many recipients.
For a large number of recipients you'll hit the bitcoin block size limit much faster than the BitShares block size limit. Not to mention the fees.

pmconrad:
A sharedrop on BTS could easily lead to more than 1 million write operations in the database. This cannot be done during normal block processing. Please modify the spec so that sharedrops are executed during maintenance.

bsip-00XX Sharedrop Operation.md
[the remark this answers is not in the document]
grctest: It shouldn't be the same price to sharedrop on 2 as it is 2 million, the more users the more resources it'll consume so the higher the fee required, no?

bsip-00XX Sharedrop Operation.md
grctest: Addressing why BSIP 20 was deferred is relevant too, I believe it's because the BSIP terminology was about dividends/profit-sharing of fees instead of sharedropping as a form of issuance? If the reason for deference was related to regulations then perhaps BTS and committee owned assets could be excluded?

OpenLedgerApp:
We will add sharedrops operations in our BSIP Market Fee Based Asset -https://github.com/bitshares/bsips/pull/134

We will modify the BSIP MFBA so that sharedrops are executed during maintenance.

sschiessl-bcp:
> We will add sharedrops operations in our BSIP Market Fee Based Asset -#134
> 
> We will modify the BSIP MFBA so that sharedrops are executed during maintenance.

I would recommend to keep them separate as they are different design. Consider that each BSIP requires voter approval. Additionally, it makes sense to have a generic sharedrop operation that can be used for MFBA and also for UIA and SmartCoins.

OpenLedgerApp:
We agree with @pmconrad 
So we will modify the BSIP MFBA so that sharedrops are executed during maintenance.

abitmore:

PR #133, opened 2018-12-17
Bsip dynamicmarketfees

OpenLedgerApp:
Dear BitShares Community,

We would like to introduce the Dynamic Market Fees BSIP.
The purpose is to support high-volume trading and market makers. Thus making it more profitable for them.  We believe it will bring more people to BitShares.

As per BitShares Core Team request we have spent some time drafting this BSIP.
And here's the resulting BSIP:
https://github.com/openledger/bsips/blob/bsip-dynamicmarketfees/bsip-00XX%20Dynamic%20market%20fees.md

The pull request is here:
#133
Forum thread is here:

Please share your opinion.
If you think it might help BitShares, please voice your opinion and vote for it, when the worker is created.

Sincerely,
Denis Sokolov
OpenLedger


bsip-0043.md
```diff
@@ -61,4 +61,6 @@ The new modification - market fee sharing - will allow to bring in new clients f
 This document is placed in the public domain.
 
 # See Also
-https://github.com/bitshares/bitshares-core/issues/1268
\ No newline at end of file
+https://github.com/bitshares/bitshares-core/issues/1268
+
+https://github.com/bitshares/bitshares-core/pull/1419
```
xeroc: Please make a separate pull request for this change

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+    Worker: TBD
+    
+# Abstract
+UIA owners do not have many marketing tools to increase trade volume of their assets.
+
+One of the ways to increase the trade volume can be Dynamic Market Fees that will provide additional discounts for active traders.
+For example, to set lower fees for trading larger volumes.
+
+So, instead of market fee, an asset owner can optionally set up dynamic_market_fee_percent property that may look like a following example table.
+**Please refer to the general notes below for terms (maker, taker) explanation.**
```
xeroc: Please make clear that the below is an example for **USD:X** markets (since the volume is declared in USD)

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+30-day volume 
+This mehod gives enough information to figure out last 30-days traders habits with much lower memory requirements but slightly less precision than the 30-day average sliding window.
+Average trading volume over the last 365 days (a year) is a variation of this method.
+
+- **30-day volume method**
+This is quite similar to the 30-day average method and shares it's advantages and drawbacks.
+
+# Technical Specifications
+
+We suggest using average total volume for a particular asset. However, it's open for the discussion.
```
xeroc: Obtaining the *average* is not trivial and adds quite some extra requirements on CPU and RAM to the backend.

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+
+## Statistics Storage
+
+There are two options for storing account statistics for trading:
+
+1. expand class account_statistics_object by adding a map of trade_statistics_object for each asset
+2. create a new type trade_statistics_object and a new index
```
xeroc: You'll need a statistic for every *base* asset (think USD) that you have traded.
Also stats should be reset after x weeks (4)

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+
+This way we accumulate total bought volume for each user for each asset and average out daily volumes for his entire term of trading of this asset (per day).
+
+Volume discounts are estimated based on the average daily `Trader_Activity  * 30 days.`
+
+_Trade statistics object_
+
+
+```
+class trade_statistics_object
```
xeroc: Need many of those per user given that I could trade in USD and CNY ...

xeroc: This will eat up a lot of RAM and, given that it is required for consensus, this also affects block producers

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+timestamp first_trade_date; //first trade date for this asset
+share_type total_volume; //total bought asset volume
+};
+```
+
+### 30-day sliding window average 
+
+In **trade_statistics_object** we need to store the queue of volumes aggregated by 1 day. In other words, for each asset a user should have a queue of 30  daily volumes.
+
+During the first maintenance interval of the day, yesterday's volume will be added to the queue and the  oldest volume will be deleted (for each asset even for assets where there was no trade done).
```
xeroc: a.k.a. 30 day FIFO (first-in first-out) buffer

bsip-00XX Dynamic market fees.md
OpenLedgerApp: Of course, there will be many of such objects per user (for each asset).
trade_statistics_objects will be stored in trade_statistics_index ordered by statistics object owner (user), asset.

bsip-00XX Dynamic market fees.md
OpenLedgerApp: pls see comment above.

bsip-00XX Dynamic market fees.md
OpenLedgerApp: queue = FIFO

bsip-00XX Dynamic market fees.md
OpenLedgerApp: thanks, noted

bsip-0043.md
OpenLedgerApp: done

ryanRfox:

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+
+The maker-taker model encourages market liquidity by rewarding the makers of that liquidity with a fee discount. It also results in a tighter market spread due to the increased incentive for makers to outbid each other. The higher fee that the taker pays is usually offset by the better prices this tighter spread provides.
+
+**2) Supporting heavy volume traders.**
+Fees are charged and deduced on a per-trade basis. The bigger “trading volume” users have traded, the lower fee they get on subsequent trades.
+
+# General notes
+Each asset will have either usual market fee or a table with dynamic market fees that will represent the table above. 
+
+A switch **market_fee_type (REGULAR, DYNAMIC)** should clearly indicate, which market fee type is used.
```
sschiessl-bcp: Is a switch here necessary?

All assets can support this market fee. The default is how it is now, and if the dynamic one is set, the dynamic one will be used.

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+
+**Makers** are the ones who put up their orders first.
+
+And **takers** are the ones to put up their orders later in time than makers.
```
sschiessl-bcp: I's suggest to enhance this BSIP by allowing maker/taker differentiation also for normal market fee (default is only one value for market fee, which is then used for maker and taker)

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+- **Average trading volume over the last 30 days - sliding window**.
+This way it is a bit clearer how to calculate. 
+However there might be extreme but real cases where this kind of calculation might not be fair: 
+a. when a person who usually trades high volumes throughout the year goes to vacation for 30 days. Suddenly all his discount is gone when he comes back
+b.   When another person comes in and makes 1 or 2 high volume trades while trading really low volumes before that. Suddenly this person might have a big discount. 
+30-day volume 
+This mehod gives enough information to figure out last 30-days traders habits with much lower memory requirements but slightly less precision than the 30-day average sliding window.
+Average trading volume over the last 365 days (a year) is a variation of this method.
+
+- **30-day volume method**
```
sschiessl-bcp: What is the difference of "Average trading volume over the last 30 days" to "30-day volume".

avg(30-day-volume) vs. sum(30-day-volume) as the metric to calculate the discount?

bsip-00XX Dynamic market fees.md
sschiessl-bcp: This can clutter the up the current state database significantly. Please indicate how garbage collection will work

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+This is similar to the 30-day sliding window average except the memory requirements for the stored values are SIGNIFICANTLY larger.
+
+## Potential issues
+We need to make sure that the suggested calculations will not lead to the significantly wrong results or the blockchain abuse. 
+We need to consider border conditions: For example, when there is no statistics at all, unusual statistics of many very small orders, very large order at the beginning of the average window and then no activity registered, etc. 
+
+ 
+## Technical challenges
+Some problems include performance as an additional search and update new index are required.
+
```
sschiessl-bcp: Please add a easy-to-understand Summary for the Shareholders

pmconrad:
Please squash this PR and use a sensible commit message.

bsip-00XX Dynamic market fees.md
```diff
@@ -0,0 +1,183 @@
...
+    Discussion: 
+    Worker: TBD
+    
+# Abstract
+UIA owners do not have many marketing tools to increase trade volume of their assets.
+
+One of the ways to increase the trade volume can be Dynamic Market Fees that will provide additional discounts for active traders.
+For example, to set lower fees for trading larger volumes.
+
+So, instead of market fee, an asset owner can optionally set up dynamic_market_fee_percent property that may look like a following example table.
```
pmconrad: Please elaborate. Since the given table is only meant as an example, how is that property to be set up and evaluated, and which degrees of freedom does this allow?

pmconrad:
This proposal introduces a significant cost in terms of CPU and RAM for all node operators. IMO this needs to be balanced against the potential benefits.
I agree that this might incentivize market makers, but it is unclear if this can justify the significant cost.

bsip-00XX Dynamic market fees.md
OpenLedgerApp: We plan to collect garbage while maintenance interval once a day in order to remove old data from the index.

bsip-00XX Dynamic market fees.md
OpenLedgerApp: In the current implementation, the switch provides supplementary features for UIA owners.
We consider the switching to be the most convenient way for users as a table is restored in one click.

bsip-00XX Dynamic market fees.md
OpenLedgerApp: It's a good point! We will update BSIP.

bsip-00XX Dynamic market fees.md
OpenLedgerApp: The difference is in the methods of updating statistics.
In the first case, statistics are updated every day on the principle of a sliding window: the trading volume of the previous day is subtracted from statistics.
In the second case, the average number of trades volume for the entire period is subtracted.

bsip-00XX Dynamic market fees.md
OpenLedgerApp: Dynamic market fee is an additional marketing tool to encourage traders. So, this BSIP helps to boost trading volume on BitShares and attract new traders to the ecosystem.

ryanRfox:
> This proposal introduces a significant cost in terms of CPU and RAM for all node operators. IMO this needs to be balanced against the potential benefits.
> I agree that this might incentivize market makers, but it is unclear if this can justify the significant cost.

@sschiessl-bcp has suggested elsewhere that we add a **Risks** section to the BSIP Template. @pmconrad may I request your craft your comment above into a paragraph highlighting the potential risk(s) you foresee? My desire is the identified risks will be added to the PR by @OpenLedgerApp

pmconrad:
Suggestion:

# Risks

## CPU Overhead

The proposed fee calculations will have to be performed on *every* trade. This adds significant burden on top of the already quite heavyweight market engine. While this is probably not a big problem with today's transaction volume, it will further limit how far our chain scales in terms of tx/s.

## RAM Overhead

The proposed calculations require logging the trade volume of all traders over a period of time, separately for each traded asset. With both the number of users and the number of assets growing over time, the required amount of RAM will grow squarely over time. This does not scale well.

bsip-00XX Dynamic market fees.md
sschiessl-bcp: Is this already implemented?

ryanRfox:
My thoughts on calculation impacts/mitigation:
1. I now understand why performing the calculation on _every_ order would be problematic.  Can we calculate a static table hourly, daily so the market can just lookup the fee? 
2. I now understand that storing complete order history for a month to use in calculation will consume substantial RAM. Can we mitigate this by calculating it into a volume average per hour, day?
3. What other options can be explored to reduce the scaling impacts within this specification? 
4. I feel dynamic market fees are a valuable addition to the protocol and may provide an incentive for further adoption, so hope we continue to debate this topic.

OpenLedgerApp:
@pmconrad  Thanks for the input! we will add this section to the BSIP description.
@ryanRfox we are analyzing your thoughts and will provide our feedback.
In any case, we should think how we can have some metrics in advance (or at the beginning of implementation) to decrease RISKS of this BSIP.

pmconrad:
> 1. I now understand why performing the calculation on every order would be problematic. Can we calculate a static table hourly, daily so the market can just lookup the fee?

We must update the trade volume per user and asset for every trade.

> 2. I now understand that storing complete order history for a month to use in calculation will consume substantial RAM. Can we mitigate this by calculating it into a volume average per hour, day?

This is already described in the section "30-day volume method". It requires two values per user and asset.

sschiessl-bcp:
I would very much like this BSIP in two steps. First, introduce maker-taker, then maybe dynamic fees like indicated above.

Limit order would have fixed fee plus market fee, both for maker and taker. But, when filled
 - maker may receive a refund for fixed fee (maybe 50%)
 - asset owner can define how much % of the market fee is refunded to the maker

Thoughts?

xeroc:
> I would very much like this BSIP in two steps. First, introduce maker-taker, then maybe dynamic fees like indicated above.

This!
"Keep it simple stupid" and do one thing at a time. Also, a BSIP that proposes only maker/taker fees will be *much* easier to get approved than something that introduces it together with something that adds significant additional burden to the chain

OpenLedgerApp:
We have made a prototype in order to measure performance 

As you can see in the picture attached, our worker has no significant influence on BitShares performance: 
image


For more details please refer the prototype. Here is the link - https://github.com/openledger/bitshares-core/commits/issue-dmf

abitmore:
@OpenLedgerApp try test with 1M users? and around 1K trading pairs (although there can be more)?

OpenLedgerApp:
> @OpenLedgerApp try test with 1M users? and around 1K trading pairs (although there can be more)?
There is a result:
dmf-perf-result

The sources are located - https://github.com/openledger/bitshares-core/commits/issue-dmf

bsip-00XX Dynamic market fees.md
nathanhourt: Why not use an Exponentially Weighted Moving Average? This would seem the obvious win to me, as EWMAs are efficient to calculate and store.

ryanRfox:
Sorry, I'm not understanding the graph. What is the X-axis representing? It seems like there are both red and blue plots for about 2/3 of the data, but only blue when the value jump. What accounts for the jump? Are there red values as well?

pmconrad:
AFAICS your test creates 1 M users but only 1000 traders, and it trades only 1 asset pair.

OpenLedgerApp:
> AFAICS your test creates 1 M users but only 1000 traders, and it trades only 1 asset pair.

We have made test test with 100 asset pair.

As you can see in the picture attached, our worker has no significant influence on BitShares performance:

image

 sources are located - https://github.com/openledger/bitshares-core/commits/issue-dmf

pmconrad:
Now you have 100 assets and 1000 traders, but each trader is using only 1 or 2 assets. Like abit said, please repeat your test with 1M traders and 1k trading pairs where each trader trades in a significant portion of these pairs.

Also, please explain your graphics, as Ryan has requested. I don't understand them either.

OpenLedgerApp:
We will update our test in the near future.

OpenLedgerApp:
> Now you have 100 assets and 1000 traders, but each trader is using only 1 or 2 assets. Like abit said, please repeat your test with 1M traders and 1k trading pairs where each trader trades in a significant portion of these pairs.
> 

We have modified the tests. There are 100 traders and 100 assets where each trader trades 100 assets. We have got the following results: Origin bitshares takes 2095465 milliseconds, and with DMF logic 2099182 milliseconds. 
You are welcome to change tests' parameters and rerun them:
 // How many traders to create
   const static auto traders_count = 100;
   // Count of cycles for trading
   const static auto iterations = 1;
   // How many assets to sell/buy
   const static auto uia_to_sell = 20;
   // How many assets to create
   const static auto ui_assets_count = 100;

New sources are located https://github.com/openledger/bitshares-core/commit/e4f9f44da756ef60a69cb6529a56c976a4362cc4

> Also, please explain your graphics, as Ryan has requested. I don't understand them either.
Red line shows how much time origin bitshares take and an other line shows how much time the same test takes with DMF logic.

abitmore:

OpenLedgerApp:
We would like to put our BSIP on voting.
The prototype is ready for voting. We have finished all test and made Pull Request.
There are some worries about computation resources. 
That why, we have made several performance tests, which showed good results. 

What should we do to put this BSIP on voting.?

@ryanRfox please let us know. Thanks in advance!

abitmore:
@OpenLedgerApp please update the document and rename the file with the assigned BSIP number.

bsip-00XX Dynamic market fees.md
sschiessl-bcp: That still doesnt define what the degrees of freedom are

sschiessl-bcp:
Please address all comments, and update your document.

Also I am still wondering if only maker/taker differentiation should be an option as well when voting as the simple version.

PR #134, opened 2018-12-18
BSIP Draft: Market Fee Based Asset (MFBA)

OpenLedgerApp:
Dear BitShares Community,

We would like to introduce the Dynamic Market Fees BSIP.
The purpose is to support high-volume trading and market makers. Thus making it more profitable for them.  We believe it will bring more people to BitShares.

As per BitShares Core Team request we have spent some time drafting this BSIP.
And here's the resulting BSIP:
https://github.com/openledger/bsips/blob/bsip-mfba/bsip-00XX%20Market%20Fee%20Based%20Asset%20(MFBA).md
The pull request is here:

The forum discussion is here:

Please share your opinion.
If you think it might help BitShares, please voice your opinion and vote for it, when the worker is created.

Sincerely,
Denis Sokolov
OpenLedger

### DIFF
--- a/bsip-00XX Market Fee Based Asset (MFBA).md
+++ b/bsip-00XX Market Fee Based Asset (MFBA).md
@@ -1,0 +1,91 @@
+    BSIP: TBD
+    Title: Market Fee Backed Asset (MFBA)
+    Authors: OpenLedgerApp <https://github.com/OpenLedgerApp>
+    Status: Draft
+    Type: Protocol
+    Created: 2018-12-17
+    Updated: 2018-12-17
+    Discussion: 
+    Worker: TBD
+
+# Abstract
+Market Fee Backed Asset (MFBA) is a SmartCoin introduced by this BSIP. 
+The main idea of a MFBA is to provide opportunity for the Assets Owners to automatically share revenue received from trading operations (Market Fees) to the holders of the MFBA in proportion of the owned amount of the MFBA.
+
+The importance of a similar solution was raised in BSIPs-19, 20,  however never implemented. (link below).
+
+
+# Motivation
+By implementing this proposals, we let Asset Owners an opportunity to use their assets as profit-sharing assets.
+Distribution of fees will be conducted automatically (in a pre-defined period of time). Such functionality is currently not available in the blockchain.
+
+# Use case
+
+ 
+
+There will be a new type of asset created - Market Fee Backed Asset (MFBA). 
+
+Let's say a user (owner of XXX generic asset) wants to share the market fee from assets  XXX.BTC, XXX.ETH, XXX.YYY with other people. 
+Then this user creates an MFBA and calls it XXX.MFBA. Then he ties his other assets: XXX.BTC, XXX.ETH, XXX.YYY to this newly created XXX.MFBA.
+Then holders of the XXX.MFBA will get a share of profits from the market fees of all the assets tied to XXX.MFBA.
+
+# Specifications
+
+There will be a special structure for MFBA parameters storing **MFBA_options**. It will be a part of general asset_options
+
+## Market Fee Backed Asset Definition
+
+Asset for which **MFBA_options** structure is available will be called MFBA Asset.
+
+## Tied assets
+Issuer of the MFBA Asset will be able to tie his/her other assets (one or many) to the MFBA Asset. Tied assets cannot become MFBA. 
+
+The list of the tied assets will be stored in the **MFBA_options**
+
+## Dividends
+
+Market Fee, received from trading operations of the tied assets will be distributed between holders of the MFBA Asset according to the stake of the MFBA Asset that holders have. 
+Issuer of the MFBA Asset must be able to set share of the Market Fee, that will be distributed to the holders of the MFBA Asset as a parameter **(MFBA share, %)**.
+Whatever is left after distribution to the MFBA holders, will be the share of the Asset Owner **(100% - MFBA share%)**.
+Funds currently in the open orders will not be included in the MFBA share.
+
+**MFBA share, %** will be stored in **MFBA_options**.
+
+## Whitelist and Blacklist
+
+Whitelist and blacklist (configurable by the UIA Issuer) could provide control over who is eligible to earn dividends. 
+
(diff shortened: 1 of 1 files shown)


ryanRfox:

bsip-00XX Market Fee Based Asset (MFBA).md
```diff
@@ -0,0 +1,89 @@
...
+The distribution occurs in the nearest maintenance interval after the  ``latest MFBA distribution time + fee_accumulation_period.
+
+## Dividends Distribution Process Restrictions
+Distribution must be performed starting from the user, that has the biggest share of the MFBA Asset.
+Users will not receive their share of Market Fee if the amount to be transferred is less than should be accrued, .  
+The undistributed market fee will be returned to the asset issuer.
+
+#Discussion
+
+# Copyright
```
sschiessl-bcp: Please add a easy-to-understand Summary for the Shareholders

xeroc:
> [](https://bitsharescan.com/account/stealth-buyback)
> 
> Dividend-like payments to a large number of token holders are computationally expensive. That's the reason why STEALTH works like it does - it buys back tokens from the market (which is computationally cheap), which creates demand and drives the price up, to the benefit of all STEALTH holders.

Of course "buybacks" are different from "dividends" not only legaly but also *emotionally*. In the first case, the *price* goes up, while in the latter case, the *amount* goes up.
Depending on juristiction, buybacks are covered in different law and may be favorable ..

OpenLedgerApp:
We are planning to run a performance test and measure computation, in order to evaluate whether the maintenance interval can be elongated depending on token holders.

OpenLedgerApp:
> Dividend-like payments to a large number of token holders are computationally expensive. That's the reason why STEALTH works like it does - it buys back tokens from the market (which is computationally cheap), which creates demand and drives the price up, to the benefit of all STEALTH holders.

We are planning to run a performance test and measure computation, in order to evaluate whether the maintenance interval can be elongated depending on token holders.

bsip-00XX Market Fee Based Asset (MFBA).md
OpenLedgerApp: **Benefits** for BTS holders:
- Increasing the profitability of the BTS DEX
- Encouraging users to hold assets on the BTS DEX 
- Bringing new users

sschiessl-bcp:
I like this BSIP as it is the consequent evolution of the Market fee sharing BSIP that is already voted in. 

Market fee sharing allows the asset owner to share `market_fee_reward_percent` with the registrar/referrer, and now you can additionally choose to to share another % with asset holders.

pmconrad:
IMO the term SmartCoin is applicable here. In my understanding, it's a generalization of MPA, PM, FBA, and anything else that is based on some automatism.

sschiessl-bcp:
The wording SmartCoin / BitAsset is exclusively reserved for crypto-collaterized, user-loaned tokens like bitUSD in my understanding.

pmconrad:
Hm, ok. Various online glossaries show that you're right. Sorry.

OpenLedgerApp:
> We are planning to run a performance test and measure computation, in order to evaluate whether the maintenance interval can be elongated depending on token holders.

We have made a prototype in order to measure performance. 
A result you can see in the picture attached

image

pmconrad:
Please explain what exactly you're measuring there.

First impression is that processing time increases by 30%, which is not acceptable IMO.

OpenLedgerApp:
We would like to explain the previously attached image.
First of all, we want to update it with more relevant data.
We have our stock-asset with the currently develop branch https://github.com/openledger/bitshares-core/commit/acf57de289aa4b43d16d5629c5d9de8a6aad9286

First of all, this graphic measures the duration of the maintenance interval, and not the work of the node as a whole.
The horizontal axis shows the number of test runs, the vertical axis shows the time of the maintenance interval.
Before measuring the maintenance time, the following steps are taken:
1. 12000 regular accounts are created (without real actions with them, for the total load)
2. 4000 user-asset is created
3. 10,000 accounts are created that will produce the assets that were created in the previous step.
4. Set up a new interval for maintenance, which is guaranteed not to be fulfilled during the whole trading period
5. 5000 orders and 5000 counter orders are created (half with full collapse, half with incomplete)
6. Waiting for the next maintenance and measure its work time.

 The Red line is the result of these steps. 

A testnet for this test is going to develop bitshares branch, the last commit in which [0358f286cebf83cd498a11411c84ab0aa927b689] 


There are following steps for measuring the interval maintenance with implemented sharedrop:
1. 100000 regular accounts are created (without real actions with them, for the total load)
2. 20,000 accounts are created, which will be the owners of the stock-asset
3. 4 stock asset is created, each of which has 1000 user-assets (total of 4004 assets)
5. 10,000 accounts are created that will produce the assets that were created in the previous step.
6. Set up a new interval for maintenance, which is guaranteed not to be fulfilled during the whole trading period
7. 5000 orders and 5000 counter orders are created (half with full collapse, half with incomplete)
8. Waiting for the next maintenance and measure its work time.

 The Blue line shows the result of these steps. Since in this case, the logic of distribution of remuneration to stock-assets holders starts
namely, 20,000 accounts, the maintenance interval execution time is increased.

We hope these explanations will answer all your questions.

image

abitmore:
> 5000 orders and 5000 counter orders

This is way too small IMHO.

abitmore:

OpenLedgerApp:
We would like to put our BSIP on voting. 
Current prototype allows to trade and makes sharedrop operations under maintenance interval. Core-team also has some worries about computation resources. 

What should we do to put this BSIP on voting?

@ryanRfox please let us know. Thanks in advance!

abitmore:
IMHO this BSIP lacks of technical specifications. Also it didn't mention anything about legal risks.

bsip-00XX Market Fee Based Asset (MFBA).md
abitmore: Need to update the document but not only reply here.

sschiessl-bcp:
You have not adressed all comments. It makes no sense to define this as a separate asset type, and especially not only for SmartCoins.

PR #149, opened 2019-02-22
Bsip fix locked accounts

OpenLedgerApp:


bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+    Status: Draft
+    Type: Protocol
+    Created: 2019-22-02
+    Updated: 
+    Discussion: https://github.com/bitshares/bsips/issues/94
+    Worker:
+
+
+# Abstract
+BitShares allow to change permissions of account A to account A. See *Cycles* paragraph of https://bitshares.org/technology/dynamic-account-permissions for details. More information can be found here: https://steemit.com/blockchain/@hipster/sad-story-how-i-lost-bitshares-account
```
ryanRfox: The Abstract should contain a synopsis of the problem and proposed solution. An explanation of cycles is needed here to explain what happened in the past to allow accounts to become locked and how they will be prevented in the future with this change. 

References to supporting documentation should be placed in the "See Also" section at the end of this document.

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+
+# Motivation
```
ryanRfox: From BSIP1: The motivation is critical for BSIPs that want to change the BitShares protocol. It should clearly explain why the existing protocol specification is inadequate to address the problem that the BSIP solves. BSIP submissions without sufficient motivation may be rejected outright.

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+We offer to develop additional functionality "Prevent to create cycled accounts". This functionality will avoid to create or update account authority with circular dependencies in the future.
+
+# Rationale
```
ryanRfox: From BSIP1 - The rationale fleshes out the specification by describing what motivated the design and why particular design decisions were made. It should describe alternate designs that were considered and related work, e.g. how the feature is supported in other languages. The rationale should provide evidence of consensus within the community and discuss important objections or concerns raised during discussion.

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+According to our investigation, BitShares has about 384 locked accounts with circular dependencies. Most of these accounts are locked by mistake. 
+
+# Specifications
+
+## Prevent circular dependencies
+We must extend create and update account evaluators adding code that detect cycled authorities and throws exception in do_evaluate() method if cycle detected.
```
ryanRfox: This is the best part so far. Within the Specifications section we typically see some pseudocode blocks. Please take guidance from others on what level of code/logic detail should be included.

bsip-0061.md
```diff
@@ -0,0 +1,169 @@
...
+    2. To find all locked accounts with circular dependencies and undo the last account_update operation.
+
+There are some cases that led to locked account:
+
+    - Alice updates, via account_update_operation, her active and owner authority to herself;
+    - Bob updates, via account_update_operation, his active and owner authority to a locked account;
+    - Jill updates, via account_update_operation, her active and owner authority threshold greater than a sum of authorities weights.
+
+This proposal is to implement a hard-fork that would prevent it in the future and to make posible unlock locked accounts.
+New operation unlock_account will be added, account_update_evaluator and account_create_evaluator will be modified.
```
pmconrad: IMO it does not make sense to add an operation that will at most be used by the few accounts that are locked now. Since the goal is to prevent such locks in the future, the operation would be useless once the accounts have been unlocked.
It is much simpler to automatically unlock them at hardfork time.

bsip-0061.md
```diff
@@ -0,0 +1,169 @@
...
+```
+```
+class sign_state
+{
+    sign_state( old_parameters, bool verify_only = false )
+
+    //
+    // New method added to parametrize class behaviour
+    //
+    bool is_key_available(const public_key_type &key)
```
pmconrad: Where will this be used?

bsip-0061.md
```diff
@@ -0,0 +1,169 @@
...
+## Unlocking account using previous authority. Fully re-done operation
+There is following use case:
+
+    1) Account "Alice" was created
+    2) Account "Alice" was updated with new authorities and options (opts 1) - here: prior state
+    3) Account "Alice" was updated with new options (opts 2)
+    4) Account "Alice" was updated with new options (opts 3)
+    5) Account "Alice" was updated with new authorities - here: locked state
+
+To unlock "Alice" we need to reapply update operation at step 2. But latest account options (opts 3) at step 4 will be lost.
```
pmconrad: The goal would be to recreate the account state after step 4.
Therefore, the `account_update` evaluator in step 5 should save the previous `owner_authority` if it detects a cycle after the update. Restoring the account then simply means moving the saved `owner_authority` back to its original place. (This should be recorded in the form of a virtual `account_update` operation.)

bsip-0061.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Unable to unlock account
+Use case:
+
+    1) Account "Alice" delegates its authority to "Bob" - here: prior state
+    2) Account "Charlie" delegates its authority to "Alice"
+    3) Account "Alice" delegates its authority to "Charlie" (transaction was signed with Bob's private key) - here: locked state
+    4) Account "Bob" delegates its authority to "Charlie" - here: locked state
+
+To unlock "Alice" we need to reapply update operation at step 1. But "Bob" also locked, we have double lock here.
```
pmconrad: The algorithm I outlined above would detect the locks on Alice and Bob. It would also resurrect both, although it would be sufficient to resurrect Alice. I think that's acceptable. Opinions?

bsip-0061.md
```diff
@@ -0,0 +1,169 @@
...
+
+# Summary for Shareholders
+We offer to develop a solution for BitShares users.
+This solution allows unlocking accounts, which are locked by mistake.
+The most important goal is to return access to the wallets and unlock their money.
+
```
pmconrad: Please add a cost/benefit discussion.
How many such locked accounts exist today, how much are the locked balances worth, and what would it cost to implement this proposal?

README.md
```diff
@@ -66,3 +66,4 @@ Number             | Title                                                    |
 [58](bsip-0058.md) | Global Settlement Protection Through Price Feeding | Jerry Liu | Protocol | Draft
 [59](https://github.com/bitshares/bsips/issues/140) | Adjustment of MSSR and MCR Through Voting | Jerry Liu | Informational | Draft
 [60](https://github.com/bitshares/bsips/issues/131) | BitShares URI scheme | John Titor, Stefan Schießl, Abit More | Informational | Draft
+[61](https://github.com/bitshares/bsips/pull/) | Fix locked accounts with circular dependencies             | OpenLedgerApp | Protocol      | Draft
```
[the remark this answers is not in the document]
OpenLedgerApp: Done.

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+## Fix locked accounts
+There are two approaches here:
+
+    1) Automatic (unconditional)
+        - Detect all locked accounts in first maintenance after hard-fork.
+        - Find latest operation ("lock"-operation) among operations of cycled accounts. Do it for each cycle.
+        - Undo all "lock"-operations. Find previous account update/create operation that changes authorities and redo it.
+
+    2) Manual (on-demand)
```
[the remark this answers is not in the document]
OpenLedgerApp: Done.

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+According to our investigation, BitShares has about 384 locked accounts with circular dependencies. Most of these accounts are locked by mistake. 
+
+# Specifications
+
+## Prevent circular dependencies
+We must extend create and update account evaluators adding code that detect cycled authorities and throws exception in do_evaluate() method if cycle detected.
+To avoid any misunderstanding of existing authorization mechanism we propose to reuse actual code of sign_state::check_authority(). Make template class traverse_authorities_state and parametrize it with approve() method. By default approve = signed_by(), an existing method of sign_state class. We add keys_available() method to resolve signing path. So, check_authority() returns true, if signing path fully ends with existing keys.
+
+## Fix locked accounts
+There are two approaches here:
```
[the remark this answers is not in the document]
OpenLedgerApp: See details in updated version. You are welcome to discussion.

bsip-0061.md
OpenLedgerApp: Done.

bsip-0061.md
OpenLedgerApp: Done.

bsip-0061.md
OpenLedgerApp: Done.

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
+    BSIP: 0061
```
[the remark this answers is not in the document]
OpenLedgerApp: Done.

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+        - Undo "lock"-operation.
+
+# Discussion
+Should we identify authority as locked in case of cycle only? Bitshares has *max_authority_depth* parameter. So, authorities with delegetion depth > *max_authority_depth* are potentially locked. Should we prevent this?
+
+## Examples
+Account A has key. Account B delegates it's authority to A - ok. C -> B - ok. D -> C - fail.
+Transaction can never be signed by D. D exceed depth limit. Current depth limit in Bitshares = 2.
+
+# Summary for Shareholders
```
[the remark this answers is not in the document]
OpenLedgerApp: Done.

bsip-0061.md
OpenLedgerApp: Done. Updated Rational and Specifications

bsip-0061.md
OpenLedgerApp: According to this post (https://github.com/bitshares/bsips/issues/94#issuecomment-412261693), we have proposed options. We can only leave an automatic approach if the community agrees.

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+        - Find latest operation ("lock"-operation) among operations of cycled accounts. Do it for each cycle.
+        - Undo all "lock"-operations. Find previous account update/create operation that changes authorities and redo it.
+
+    2) Manual (on-demand)
+        - Check if account is locked.
+        - Find latest operation ("lock"-operation) in cycle.
+        - Undo "lock"-operation.
+
+# Discussion
+Should we identify authority as locked in case of cycle only? Bitshares has *max_authority_depth* parameter. So, authorities with delegetion depth > *max_authority_depth* are potentially locked. Should we prevent this?
```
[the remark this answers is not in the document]
OpenLedgerApp: We have now removed everything about GRAPHENE_MAX_AUTHORITY_DEPTH from the document. But according to the rule 'verify_cycled_authority ()', accounts in which the authority delegation level exceeds max_authority_depth are considered blocked. Correspondingly, the create_account_operation and update_account_operation operations will not pass evaluation. Thoughts?

bsip-0061.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Examples
+Account A has key. Account B delegates it's authority to A - ok. C -> B - ok. D -> C - fail.
+Transaction can never be signed by D. D exceed depth limit. Current depth limit in Bitshares = 2.
```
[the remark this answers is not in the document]
OpenLedgerApp: See comment above.

bsip-0061.md
pmconrad: Oh, nice catch. To (allegedly) quote our former German chancellor Adenauer - "Why should I care about the silly things I said yesterday?" :-)

I think the question is relevant to the cost/benefit discussion. Perhaps the community should decide, like you said.

bsip-0061.md
OpenLedgerApp: Updated.

pmconrad: cost/benefit discussion?

bsip-0061.md
OpenLedgerApp: Updated.

bsip-0061.md
OpenLedgerApp: Updated. See: Fix locked accounts - on-demand

bsip-0061.md
OpenLedgerApp: Here I meant that account recovery in one step is not always possible. See diagram below.

Unlock accounts

So, automatic recovery requires repeating unlock procedure to a normal state. In manual mode, another operation may also be required before full unlocking.

bsip-0061.md
OpenLedgerApp: Could you please clarify the **_cost/benefit_**?

pmconrad: > How many such locked accounts exist today, how much are the locked balances worth, and what would it cost to implement this proposal?

OpenLedgerApp: Today exist over 300 locked accounts. 
The most important goal is to prevent irreversible blocking of accounts and help to regain access to the already blocked accounts and their balances.

pmconrad: Thanks. Didn't expect that many...

bsip-0061.md
pmconrad: Right, thanks.

ryanRfox: In the first the first transition (3), Alice changes her authority from Bob to Charlie, which creates a cycle. I thought we had guards on that to a certain depth, no? Same for Bob: do we guard against him delegating into an cycle. 

Why does the undo happen in the forward order 3, then 4? I thought it would undo the most recent and repeat until the cycle is eliminated.

bsip-0061.md
nathanhourt: Is there a list of these accounts somewhere?

bsip-0061.md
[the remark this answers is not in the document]
nathanhourt: In answer to your question, @ryanRfox, that limit is found in the committee-controlled parameters [here](https://github.com/bitshares/bitshares-core/blob/8c1a78b7f3f0083abee695c1b8a9b944f2c4a1be/libraries/chain/include/graphene/chain/protocol/chain_parameters.hpp#L70) and is used when verifying transaction authorization [here](https://github.com/bitshares/bitshares-core/blob/8c1a78b7f3f0083abee695c1b8a9b944f2c4a1be/libraries/chain/db_block.cpp#L642). It is defaulted to 2 in the static configuration [here](https://github.com/bitshares/bitshares-core/blob/8c1a78b7f3f0083abee695c1b8a9b944f2c4a1be/libraries/chain/include/graphene/chain/config.hpp#L36), although that default is irrelevant by now: the committee controls the value.

bsip-0061.md
```diff
@@ -0,0 +1,182 @@
...
+    - Initiator (potentially *account_to_unlock*) pays fee for this operation.
+    - *account_to_unlock* should have sufficient balances to perform this operation.
+    - Check if *account_to_unlock* is locked in unlock_account_evaluator.
+    - Check if *account_to_unlock* is authorized to perform the operation (transaction signed with previous authority).
+    - Find previous account update/create operation that changes authorities - restore point.
+    - Undo "lock"-operation: extract owner and active authorities from restore point, update account with extracted data.
+    - Push corresponding virtual account_update_operation.
+
+# Specifications
+## Prevent circular dependencies
```
pmconrad: The specification is incomplete. It does not specify the unlocking process.

Also, please keep in mind that the target audience of a BSIP consists largely of non-programmers. Code as specification is insufficient. Pseudocode is welcome for data structures etc., but logic should be describe in a human-readable way. (Some of the "Rationale" can be included in the specification.)

bsip-0061.md
abitmore: @ryanRfox :
1. we need to fix historical locked accounts, the logic is described above by @OpenLedgerApp ;
2. after that, in the long run, we avoid new account locking with the logic described by you.

bsip-0061.md
OpenLedgerApp: List of locked accounts
1.2.352708
1.2.363525
1.2.312657
1.2.133051
1.2.100886
1.2.117772
1.2.182288
1.2.1150329
1.2.380405
1.2.777808
1.2.863848
1.2.996565
1.2.989841
1.2.31223
1.2.166289
1.2.169641
1.2.186918
1.2.206921
1.2.166880
1.2.12418
1.2.98043
1.2.501729
1.2.740449
1.2.665269
1.2.960390
1.2.1259335
1.2.179504
1.2.155110
1.2.262650
1.2.380718
1.2.378612
1.2.456139
1.2.403919
1.2.107093
1.2.165413
1.2.165714
1.2.184024
1.2.907383
1.2.843615
1.2.834573
1.2.1057592
1.2.1218022
1.2.405903
1.2.265235
1.2.355200
1.2.433810
1.2.460835
1.2.122787
1.2.93429
1.2.665231
1.2.443082
1.2.1107208
1.2.1147487
1.2.1150333
1.2.290543
1.2.151149
1.2.163782
1.2.515217
1.2.96015
1.2.20080
1.2.182258
1.2.417925
1.2.663843
1.2.608383
1.2.1150281

bsip-0061.md
OpenLedgerApp: Thanks, abitmore!
This is exactly what we meant.

bsip-0061.md
abitmore: These are 65 accounts, but not about 300?

OpenLedgerApp: We decided to change the search algorithm according to the comments - https://github.com/bitshares/bsips/issues/94

As a result, the number of locked accounts has decreased.

List of the locked balances - https://docs.google.com/document/d/1Dmcr9QzSWnCbSKcpaN08iENJyn6AWogbKjZ_iHH4OM0/edit

OpenLedgerApp: We have summarised the most liquid currencies of locked accounts.

Here the list:
BTS - 1,641,712.7494
USD - 14,546.7176
BTC - 6.86072047
CNY - 16,396.3406
ETH - 0.278826
LTC - 8.60401198
DASH - 0.23722333
ICOO - 348.1812
OBITS - 3,102.1454

Total locked budget is 144,619.841 bitUSD

As you can see the total budget is huge. We offer to unlock these accounts.

pmconrad: Thanks. That is indeed significant.

How about taking 10% from each account as payment for the implementation? (Seriously. The accounts are locked which means a 100% loss for the owner. We're offering to return 90%.)

OpenLedgerApp: > How about taking 10% from each account as payment for the implementation? (Seriously. The accounts are locked which means a 100% loss for the owner. We're offering to return 90%.)

It makes sense.
@pmconrad  could you please share your idea of how this can be implemented in a technical way?

pmconrad: At unlock time, walk through the account balances and auto-transfer 10% from each non-zero balance to a specific account, e. g. committee-account. That's pretty straightforward.

The non-technical side may be more tricky. :-/

OpenLedgerApp:
We need to determine the way how to fix locked accounts. 
There are two ways:
 1) Automatic (unconditional)
    - Detect all locked accounts in first maintenance after hardfork.
    - Find latest operation ("lock"-operation) among operations of cycled accounts. Do it for each cycle.
    - Undo all "lock"-operations. Find previous account update/create an operation that changes authorities and redo it.

 2) Manual (on-demand)
    - Check if account is locked.
    - Find latest operation ("lock"-operation) in cycle.
    - Undo "lock"-operation.

Let's decide it will be automatic or manual way.

OpenLedgerApp:
Due to the fact that the algorithm for finding and unlocking accounts is quite complex and will require significant costs, we suggest considering another opportunity to unlock in the near HF:
1) add functionality - prevent creation cycled authority
2) compile a list of blocked accounts, find blocking operations and previous account states
3) perform undo blocking operations and return the account to the previous state (hardcoded operations list)

After HF, we check if there are still blocked accounts. They may appear after adding the unlock code, but before the HF. If there are any, then add them hardcoded to the next HF.

pmconrad:
IMO the algorithm for finding locked accounts is exactly as simple as detecting the lock in the first place.
Since the detection logic will have to be implemented anyway, you can simply apply it to all account_update operations:
* Before the hardfork, if you detect a lock, put the account and its previous authorities into a list.
* At hardfork time you can simply take the list and restore account authorities to the saved state, which unlocks them.
*After the hardfork, if you detect a lock, fail the transaction.

OpenLedgerApp:
The logic of finding a blocking operation is not as simple as determining whether an account has been blocked. In addition to the fact that the account is blocked, we must find the last operation in the chain of blocked accounts that led to this state. If you perform such calculations in bitshares-core in maintenance, then you should also take into account that the history plugin may be inactive and the nodes must come to consensus even without it. Without history plugin, this is problematic. You need to build a cache of recent authority change operations during replay, which will require a large amount of additional memory.

pmconrad:
> The logic of finding a blocking operation is not as simple as determining whether an account has been blocked.

Yes it is. An account that is blocked cannot be updated. So a blocking operation is an `account_update` after which the account has been blocked.

Note that it is not necessary to find the complete original `account_update` that caused the lock. It is sufficient to remember the owner authority as it was before the lock, and restore that.

abitmore:

sschiessl-bcp:
Massive amount of hours went into this. What is the status @OpenLedgerApp ?

PR #153, opened 2019-03-06
Proposed details for BSIP-22 specifications (vote decay)

MichelSantos:
These are proposed details about vote decay

Code specifications can be added if these details are acceptable to @grctest


MichelSantos:
@grctest The original specification for "Accounts blacklisted from voting (exchanges/services/scam accounts)" was temporarily removed with the idea that it might be better addressed with its own BSIP unless you think that it is critical to include as a part of this BSIP.

The original specification also suggests that work proposals are not affected by vote decay at all.  Does that match your vision, or do you have something more nuanced in mind?

grctest:
I agree with the removal of vote blacklists from the BSIP. Regarding worker proposals, since they have a short fixed duration of payment I don't think they need vote decay.

bsip-0022.md
```diff
@@ -37,34 +37,202 @@ Currently within Bitshares DPOS system, an user's votes for committee/witness/pr
...
+
+
 # Specifications
 
-## Committee controlled variables
+## <div id="referendum-categories" /> Referendum Categories
+
+BitShares has four categories of refereundums that core token holders may vote on: witnesses, committee members, worker proposals, and proxy delegation.  Prior to this proposal, any vote within any category applies until changed or until the proxy assignment had been assigned.  The removal of proxy assignment re-instates a core token holders' prior voting selection for witnesses, commitee members.
```
pmconrad: Typo: refere*u*ndums

bsip-0022.md
```diff
@@ -37,34 +37,202 @@ Currently within Bitshares DPOS system, an user's votes for committee/witness/pr
...
+
+This proposal shall retain the ability of every core token holder to vote on all referendum categories and the optional selection of a single proxy.
+
+Vote for three of the referendum categories shall be "decayable": witnesses, committee members, and proxy delegation.  Votes for worker proposals [shall continue to be tabulated as they were prior to this proposal](#out-of-scope) because they are generally fixed in length.
```
pmconrad: ~length~ duration

bsip-0022.md
```diff
@@ -37,34 +37,202 @@ Currently within Bitshares DPOS system, an user's votes for committee/witness/pr
...
+After the last decay period has expired, the voting power shall reduce to zero until either the account re-casts its votes, or the [vote decay parameters are adjusted](#vote-decay-parameters) such that the age of the account's last vote within that category possibly moves from "expired" to the the "decay age" or "full-power age".
+
+In summary, the age of a category vote by an account will typically decay from "full-power" through "decaying" to "expired" unless and until the account re-casts its voting slate in any referendum category.
+
+---
+<center>**Example of Vote Decay**</center>
+
+The vote decay parameters for witness voting have been set to 365 days for the full-power duration, 180 days for the total vote-decay duration, and 3 vote decay steps.  Each vote-decay sub-period has a duration of 60 days.
+
+A voter changes their vote for witnesses on December 31, 2019, and then does not vote again until December 31, 2025.
```
pmconrad: I suppose he does vote on Jan 1, 2026?

bsip-0022.md
```diff
@@ -37,34 +37,202 @@ Currently within Bitshares DPOS system, an user's votes for committee/witness/pr
...
+- Core token holders who select [a proxy account that fails to re-cast its own votes periodically will have their effective voting power decay over time](#proxy-power) despite re-casting their own votes for that proxy.
 
-if (days_since_decay_began > max_days_decay) {
-    decaying_vote_weight = 0; //Or nullify the vote
-} else {
-    decaying_vote_weight = original_vote_weight - ((original_vote_weight/max_days_decay)*days_since_decay_began)
-}
-```
 
 # Discussion
```
pmconrad: The calculations to be performed will significantly increase the duration of the maintenance interval, and with it replay time. I would like to see some numbers on the effect of the proposed changes assuming they had been activated at the beginning of our history (1. with the assumption that proxies are never updated, and 2. with the assumption that proxies are regularly updated).

dexy-dex: To avoid increasing of maintenance period, it would be better just assume that accounts with the last outgoing transaction(or other initiated by user operation) more than 1 year ago have zero vote power. imho.

If the last account activity was more than 1 year ago (for example), it means that the holder has died, lost his keys or simply doesn't participate in the community now, therefore we can't consider his voting power to be actual.

bsip-0022.md
MichelSantos: > Typo: refere_u_ndums ... commit_t_ee

A correction will be made


> I don't think proxy delegation is a Category of its own. It's not like we're voting _for_ proxies, we're voting _via_ proxies.

@pmconrad Is your objection specifically about calling it a "referendum category"?  Or is it about decaying the delegation for a proxy?


> Removal of proxy reinstates prior votes also for workers.

A correction will be made

bsip-0022.md
MichelSantos: The intent is to retain the current scheme of vote tallying for worker proposals: 

- If a token holder has not selected proxy then their stake is assigned to whatever worker proposal is selected by the token holder _at the time of vote tallying_.

- If a token holder has selected a proxy then their stake is assigned to whatever worker proposal is selected by the proxy _at the time of vote tallying_.

bsip-0022.md
MichelSantos: > I suppose he does vote on Jan 1, 2026?

The vote tallying mechanism has not yet been explicitly described in the BSIP.  If it occurs only during the current maintenance interval which is currently once every 24 hours, if I remember correctly, then a vote on December 31, 2025 might not be detected until the following day.  Should a sentence be added here that states the assumption that vote tallying will occur during the next maintenance interval which is assumed to occur on the following day?

Should the BSIP explicitly specify that vote decay will be done as part of the current vote tally activity?  Should this be left to the code technical specifications?

The frequency of vote tally is also related to your performance question.

bsip-0022.md
MichelSantos: > To avoid increasing of maintenance period, it would be better just assume that accounts with the last outgoing transaction(or other initiated by user operation) more than 1 year ago have zero vote power. imho.

I think that this suggestion can be an optimization that can avoid _some_ of the arithmetic overhead.

MichelSantos: > I would like to see some numbers on the effect of the proposed changes assuming they had been activated at the beginning of our history (1. with the assumption that proxies are never updated, and 2. with the assumption that proxies are regularly updated).

I will make sure that someone follows up on this.

bsip-0022.md
pmconrad: What I mean is - you say that proxy delegation is decayable but worker votes are not. If I have set a proxy and the proxy votes for workers, do my votes for the worker count in full or as per decay rules?

bsip-0022.md
pmconrad: A few lines further down you say that he has 100% voting power starting Jan 1 2026, so I assume he must have voted at that point?

Yes, it might be a good idea to specify that vote decay takes effect during vote tallying, which happens in the maintenance interval.

bsip-0022.md
pmconrad: > To avoid increasing of maintenance period, it would be better just assume that accounts with the last outgoing transaction(or other initiated by user operation) more than 1 year ago have zero vote power.

I think this suggestion side-steps the original intent of this proposal. In my understanding, the idea is that users must actively participate in our governance model, by regularly updating their votes. Activity unrelated to voting should not prevent decay.

bsip-0022.md
```diff
@@ -87,15 +87,15 @@ A voter changes their vote for witnesses on December 31, 2019 20:00 UTC.  This v
 
 ![Qualitative decay of voting power](bsip-0022/general-vote-decay.png)
 
-The voting power of an [account's voting stake](#account-voting-stake) shall be set to full voting power whenever an account's vote is re-cast for _any_ [referendum category](#referendum-categories).  The voting power shall remain at 100% for each of the referendum categories for the entire ["full-power period"](#vote-decay-parameters).
+The voting power of an [account's voting stake](#account-voting-stake) shall be set to full voting power whenever an account's vote is re-cast for _any_ [referendum category](#referendum-categories) or is re-delegated to a voting proxy.  The voting power shall remain at 100% for each of the referendum categories for the entire ["full-power period"](#vote-decay-parameters).  _Retracting the delegation from a voting proxy without also casting a vote does not reset the voting power to 100%_.
```
pmconrad: > Retracting the delegation from a voting proxy without also casting a vote does not reset the voting power

Right now it is technically not possible to change/unset the voting proxy without also specifying a (possibly empty) list of votes. This could change in the future, so maybe leave this in.

bsip-0022.md
```diff
@@ -52,11 +52,11 @@ A [proposal for Peerplays](https://github.com/peerplays-network/pips/blob/master
...
 
-This proposal shall retain the ability of every core token holder to vote on all referendum categories and the optional selection of a single proxy.
+This proposal shall retain the ability of every core token holder to vote on all referendum categories and the optional delegation of a voting proxy.
 
-Vote for three of the referendum categories shall be "decayable": witnesses, committee members, and proxy delegation.  Votes for worker proposals [shall continue to be tabulated as they were prior to this proposal](#out-of-scope) because they are generally fixed in length.
+Votes for two of the referendum categories shall be "decayable": witnesses, committee members.  Votes for worker proposals [shall not be decayable and shall continue to be tabulated as they were prior to this proposal](#out-of-scope) because they are generally fixed in length.  [Vote delegation to a proxy shall be decayable](#proxy-power) [except for votes for worker proposals](#proxied-votes-for-worker-proposals).
```
pmconrad: > except for votes for worker proposals

I think this contradicts your clarification in the other commit. Vote delegation *does* decay if the delegator doesn't update. Only if the proxy neglects his duties, the votes will count for workers with 100% and for witnesses/committee with the proxy's decay factor.

bsip-0022.md
```diff
@@ -68,6 +68,21 @@ _The full power of an account's vote will automatically be updated to reflect an
 This proposal shall retain the calculation for the maximum voting stake of an account.
 
 
+## Schedule of Voting Power Tallying
+
+Vote power tallying retain the existing schedule for vote tallying which is occurs during every maintenance interval.
```
pmconrad: ...retain*s*... which ~is~ occurs ...

clockworkgr:
I mentioned the following in TG the other day:

> Alex M - clockwork, [03.04.19 13:08]
> wouldn't per votable object proxying make more sense?
> If I want to proxy my voting stake to someone
> it kinda makes sense that some proxies will be more appropriate than others
> e.g. I'd probably trust Abit to pick witnesses
> but Bitcrab for governance
> and someone else for Budgeting (workers)
>

It could be a different BSIP, but possibly also part of this as part of a general voting "overhaul" hence why I thought I'd mention it.

@MichelSantos thoughts?

bsip-0022.md
MichelSantos: > I don't think proxy delegation is a Category of its own. It's not like we're voting _for_ proxies, we're voting _via_ proxies.

After some thought, I have come to agree.  The draft has been updated to make this distinction.

MichelSantos:
The voting stake of the BitShares blockchain was analyzed at the start of every month since November 2015.

## Voting Stake

Historical BTS Voting Stake

This plot shows the voting stake split into two categories.  The first category called "Direct" (shown in blue) is the voting stake that is directly voting for any referendum category.  The second category called "Delegated" (shown in gold) is the voting stake that is delegated to any proxy.

### Commentary

This plot demonstrates an interesting decrease in the delegated voting over most of 2018, and a sharp increase in the delegated voting stake after 2019-03-01.  _This latter increase is mostly drawn from stake that had not been voting at all._


## <div id="voting-participation-rate"/>Voting Participation Rate

Historical Participation Rate

This plot visualizes how much of the voting stake is participating in governance of the blockchain.  The voting stake is expressed relative to the circulating supply at the start of every month.  The "Direct" line (shown in blue) is the voting stake that is directly voting for any referendum category.  The "Delegated" line (shown in gold) is the voting stake that is delegated to any proxy.  The "Total" line (shown in black) is the summation of the direct and delegated voting stakes.

### Commentary

This plot demonstrates an interesting decrease in the delegated voting over most of 2018, and a sharp increase in the delegated voting stake after 2019-03-01.  The 2019-06-01 participation rate of 46% is the highest rate in the history of the blockchain.



## <div id="undecayed-relative-weight"/> Relative Weight of Direct versus Delegated

Historical BTS Voting Stake Distribution

This plot visualizes the direct and delegated voting stakes relative to each other.  The "Direct" fraction (shown in blue) is the voting stake that is directly voting for any referendum category.  The "Delegated" fraction (shown in gold) is the voting stake that is delegated to any proxy.

### Commentary

This plot demonstrates the predominant tendency of approximately 80% of _voting_ account holders to delegate their voting stake to another account.



## <div id="decayed-voting-stake"/>Decayed Voting Stake

Historical BTS Voting Stake with Post-hoc Decay


This plot visualizes the same voting stakes decayed by different durations and conditions.  The "Direct" line (shown in solid blue) is the voting stake that is directly voting for any referendum category.  The "Direct 90" line (shown in dotted blue) is the voting stake that is directly voting for any referendum category _within the prior 90 days_.  The "Direct 360" line (shown in dash-dot blue) is the voting stake that is directly voting for any referendum category _within the prior 360 days_.  The "Delegated" line (shown in solid gold) is the voting stake that is delegated to any proxy.  The "Delegated 90" line (shown in dotted gold) is the voting stake that is delegated to any proxy **and** where _the proxy has voted within the prior 90 days_.  The "Delegated 90 by Active Delegation" line (shown in solid green) is the voting stake that is **"actively"** delegated to any proxy **and** where _the proxy has voted within the prior 90 days_.  ("Active" delegation is a delegation by an account **after account creation**.)    The "Delegated 90 by Active Delegation 360" line (shown in solid green) is the voting stake that is **"actively"** delegated to any proxy _within the prior 360 days_ **and** where _the proxy has voted within the prior 90 days_.

### Commentary

This plot demonstrates how decaying votes reduces the amount of voting stake over time.

The Direct 90 and Directo 360 voting stakes are less than the full-power direct but they do not exhibit any major or sustained changes with the possible exception around 2019-09 when Direct 360 suddenly approaches the Direct voting stake.

The Delegated 90 voting stake is mostly identical to the Delegated voting stake until 2018 when it begins to slightly diminish.

The Delegated 90 by Active Delegation 360 voting stake exhibits the most dramatic changes in voting stake.  One year after the start of the blockchain, this voting stake begins to deviate significantly from Delegated 90.  It exhibits a significant decrease after 2018-02-01 and then a sharp increase after 2019-04-01.


## <div id="decayed-relative-weight"/> Relative Weight of Decayed Direct versus Decayed Delegated

Historical BTS Voting Stake Distribution with Post-hoc Decay

This plot visualizes an _arbitrary selection_ of decayed voting stakes relative to each other.  The "Direct 360" fraction (shown in blue) is the voting stake that is directly voting for any referendum category _within the prior 360 days_.  The "Delegated 90 by Active Delegation 360" fraction (shown in gold) is the voting stake that is **"actively"** delegated to any proxy _within the prior 360 days_ **and** where _the proxy has voted within the prior 90 days_.


### Commentary

This comparison of an arbitrarily decayed direct and delegated stake is intended to be compared against the [un-decayed relative weight](#undecayed-relative-weight) to give some idea about how decaying voting stake can give certain stake more or less voting weight relative to the other.  This comparison is for voting data in BitShares _where there is no vote decay_ and it might not reflect how account holders will behave under a regime of vote decay.  But it does demonstrate the potential effects of activating vote decay in BitShares.

pmconrad:
Excellent report, thanks!

Some observations:
* People who vote on their own seem to vote more or less regularly (compare blue lines)
* Most proxies seem to be voting regularly as well (compare yellow lines)
* People rarely switch their proxy if they have set one (compare yellow / dotted green)

It follows that vote decay would not have a very large (in the sense of damaging or dangerous) effect on voting, whereas proxy decay would. IMO it would be important to notify users of their vote decay in the UI to mitigate the possible effects.

oxarbitrage:
Found some issues in the Rationale section of this BSIP that i think can be updated in the context of this pull request:

There are 4 items in "Campaigning efforts since the upgrade from BTSX (0.9x) to BTS 2.0 have drastically stagnated. ".

- There were plans for a 'DPOSHub' to stimulate campaigning efforts, this has yet to materialize.
**DPOSHub link is not working anymore.**
- Elected witnesses rarely make an active effort to attend BeyondBitcoin hangouts nor hold their own scheduled hangouts, they are largely a silent group of individuals.
**Not valid anymore, beyond bitcoin hangouts do not exist since a while and witnesses are not a silent group anymore.**
- The 'Stakeholder proposal' subforum on Bitsharestalk is not as active as it should be; many elected witnesses haven't provided a discussion URL and many such threads on bitsharestalk haven't received a new post/update in months/years.
**Majority of active witnesses do have an url.**
- The list of active witnesses/committee members rarely changes.
**Not so true in this days IMO.**

MichelSantos:
> I mentioned the following in TG the other day:
> 
> > Alex M - clockwork, [03.04.19 13:08]
> > wouldn't per votable object proxying make more sense?
> > If I want to proxy my voting stake to someone
> > it kinda makes sense that some proxies will be more appropriate than others
> > e.g. I'd probably trust Abit to pick witnesses
> > but Bitcrab for governance
> > and someone else for Budgeting (workers)

@clockworkgr Please see #177

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
+
+Votes for two of the referendum categories shall be "decayable": witnesses and committee members.  [Votes for worker proposals shall not be decayable](#out-of-scope) and shall continue to be tabulated as they were prior to this proposal because they are generally fixed in duration.  [Vote delegation to a proxy shall also be decayable](#proxy-power).
+
+
+## <div id="account-voting-stake" /> Voting Stake of an Account
+
+The full power of an account's vote equals the [the sum of](https://github.com/bitshares/bitshares-core/blob/bf7ff54d9a17aa43f4663521e371b8c0ddfc2284/libraries/chain/db_maint.cpp#L1184-1186) an account's liquid core tokens (BTS), core tokens in vesting balances, and core tokens in open orders including [collateral that is used for shorting smart assets into existence](https://github.com/bitshares/bitshares-core/blob/bf7ff54d9a17aa43f4663521e371b8c0ddfc2284/libraries/chain/market_evaluator.cpp#L234-245).
+
+_The full power of an account's vote will automatically be updated to reflect any changes to the liquid balance, vesting balance, or core tokens in open orders_.
```
pmconrad: Perhaps add a note that this list may be expanded on by future BSIPs, e. g. Margin Trading.
The important sentence here is the next one: current calculation method shall be retained.

bsip-0022.md
pmconrad: Note that with the proposed explicit voting operation, not changing anything *does* reset the voting power. This is a contradiction.

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
+
+A reduction of either full-power duration or the decay duration could have the effect of reducing the voting power of an account by effectively aging an account's vote to "expired" status from either "full-power" or "decaying" status.  A qualitative example of this can be seen in the figure if the decay parameters are changed from Settings A to Settings B.
+
+Similarly, an increase of either the full-power duration or the decay duration can increase the voting power of an account by effectively rejuvenating an account's vote from "expired" status to either "decaying" or "full-power" status, or from "decaying" to "full-power" status.  A qualitative example of this can be seen in the figure if the decay parameters are changed from Settings B to Settings A.
+
+
+### Changing the Decay Steps of a Referendum Category or Proxy Delegation
+
+Changing only the quantity of decay steps can potentially change the voting power of accounts _whose voting age fall within the decay period_.  The change in voting power for these accounts is a function of both the change in the quantity of decay steps, and the account's voting age.
+
```
pmconrad: Note: the voting power of an account or proxy is recalculated during vote tallying using the chain parameter at that point in time. The chain is ignorant of the "change" in the sense that the outcome of vote tallying depends only on the *current* settings of the chain parameters, not on the settings that were valid at the previous vote tallying.

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
+**Note 2:** "Implementation hints" are not to be considered part of the formal specification, but merely as a *possible* implementation.
+
+### Vote Decay parameters
+
+The following vote decay parameters shall be selectable by the BitShares Committee.
+
+```
+struct category {
+   bool active;
+   unsigned_int full_power_duration; // days
```
pmconrad: Note: "days" means periods of 86400 seconds. Local calendars are ignored, as are DST and other irregularities.

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
 
-## Should the committee have the power to blacklist users from voting?
+It is here that the vote decay could be applied by calculating a voting stake multiplier (&alpha;) which for vote decay would be equal to or less than one (&alpha; &le; 1).  This new function (`calculate_voting_power(opinion_account)`) will calculate the multipler as a function of the current decay parameters and the account's age of last vote.
 
-Whilst exchanges/services voting with user funds (without permission) is bad, such functionality opens the doors for normal users to be included in the blacklist. One would hope that the network would vote such a committee out, but what's to stop the committee then blacklisting users that don't vote for them (or who vote for competitor committee users) so as to maintain power indefinetley? Currently highly unlikely, but still theoretically plausible.
+Any account may have four (4) different multipliers for each of the three (3) referendum categories and the one (1) proxy delegation.  Therefore the `calculate_voting_power(opinion_account)` function can return a multiple-valued structure:
+
+```
+struct voting_power {
+   double for_witness;
```
pmconrad: No `double` in consensus code. Each multiplier is a rational value of `1 - (periods passed)/(total periods)`

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
+       d._vote_tally_buffer[offset] += voting_power.witness * voting_stake;
+...
+```
+
+The subsequent logic of `db_maint` will remain unchanged.
+
+
+### <div id="software-specs-api"/> API Calls
+
+- An API call shall be added, or an existing one modified, to get the blockchain's current vote decay parameters
```
pmconrad: This happens automatically because the new parameters are returned as part of the existing chain_parameters structure.

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
+
+- An API call shall be added, or an existing one modified, to get an account holder's date of last vote
```
pmconrad: This should already exist from https://github.com/bitshares/bitshares-core/pull/1449/

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
+
+- At least one (1) API call shall be added added to get either the _current_ voting power of an account _or_ the voting power of an account at the _preceding_ maintenance interval
```
pmconrad: Do we really need this? Current voting power can be calculated by the client from known values (balances + chain parameters). Recording previous voting power will require a lot of additional database writes during maintenance.

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
+* An argument that is repeatedly raised is that voters are apathetic towards voting on the BitShares platform, and that they may not repeatedly vote which could lead to a reduced cost to attack the network (by voting in malicious users).
+* Whilst the potential consequences of catastrophic voter apathy is a legitimate concern, perhaps by promoting a healthier democratic process we can offset the risk that voter apathy poses. We aught to be promoting the regular campaigning for these positions of power so as to encourage healthy competition which will lead to innovation. Sitting back and not addressing/improving the voter apathy issue isn't a solution nor a reason to block the implementation of this proposal.
```
pmconrad: Typo: aught

bsip-0022.md
```diff
@@ -9,113 +9,439 @@
...
 
 ## Concerns of a lack of central communications channel
 
-* Some users have voiced concerns that because there isn't one mutually utilized central communications channel within the Bitshares community, we shouldn't implement vote weight decay.
-* I can sympathise with this concern, our community is distributed around the world & some countries block communication channels that others have access to (Russia rumoured to potentially block Telegram, China's "great firewall", venezuela blocking whatsapp, etc), likewise there are several language barriers between sub-communities.
-* To account for this concern, we should not implement this change without sufficient discussion nor consensus & transcript translation (and distribution of such information) for our users worldwide.
+* Some users have voiced concerns that because there isn't one mutually utilized central communications channel within the BitShares community, we shouldn't implement vote weight decay.
+* I can sympathise with this concern, our community is distributed around the world and there are several language barriers between sub-communities.
```
pmconrad: Please re-word - IMO individual feelings should not be expressed here. The concerns (distribution + language barrier) should be mentioned of course.

bsip-0022.md
MichelSantos: This is not necessary but it does provide a convenient way for a regular account holder to check their voting power from an API node.  Otherwise the account holder must _hope_ that their voting power is actually what their client claims it to be.

The node is calculating this at every maintenance interval anyway.  Can this be lazily calculated so that it need not be persisted when shutting down?

bsip-0022.md
MichelSantos: An alternate variable type and pseudo-code was introduced for your review

pmconrad: I like your suggestion to use a scaled version of vote tallying. Perhaps best leave the `struct voting_power` completely out of this spec.

bsip-0022.md
pmconrad: Since this is not a consensus related change I'd suggest to move this discussion into a normal "core" issue, e. g. https://github.com/bitshares/bitshares-core/issues/1884

MichelSantos: This has been removed from the API section

bsip-0022.md
MichelSantos: The `voting_power` struct and the related scaling implementation has been removed

PR #180, opened 2019-07-15
Update market URI in BSIP60

abitmore:
Due to historical reasons the reference wallet and some blockchain explorers implemented `market/asset_asset` scheme, so I think it's better to stick with it. E.G.
* https://wallet.bitshares.org/#/market/BTS_CNY
* https://bts.ai/market/BTS_CNY
* https://cryptofresh.com/a/BTS_CNY

### DIFF
--- a/bsip-0060.md
+++ b/bsip-0060.md
@@ -131,11 +131,19 @@ lowalpha             = "a" | "b" | "c" | "d" | "e" | "f" | "g" | "h" |
 
 Market paths start with the word "market", followed by two `asset_id_or_symbol`s.
 
-Example: `market/BTS_USD`
+Example:
+* `market/BTS/USD`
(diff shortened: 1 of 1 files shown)


pmconrad:
Hm.
Now that the BSIP is officially included in this repo I think we can't just change things. After all, it's supposed to be a kind of standard.
Also, I think using `_` as a separator in an URL is a poor choice, so defining this to be the standard is not a good idea.

Perhaps better add an explanation that for historical reasons `_` can also be used as the separator, but it is considered deprecated and applications should prefer `/`.

abitmore:
When drafting new standards we should take existing applications into consideration. When updating a standard we should also consider influence of the already published old standard. So I'd like to support both format in the standard. This may create technical debts though.

Issue #175, opened 2019-06-08
BSIP69: Additional Assert Predicates


christophersanborn:
BSIP: 0069
    Title: Additional Assert Predicates
    Authors: Christopher J. Sanborn, ...
    Status: Draft
    Type: Protocol
    Created: 2019-06-08
    Discussion: https://github.com/bitshares/bsips/issues/175

# Abstract

New predicates for the `assert_operation` are proposed to enable conditional validity of transactions for various use cases, including safety of transactions that depend on recent prior transactions, safety of transactions signed by hardware wallets, and transactions signed in advance but intended to be broadcast at a later time.  Removal of a stale committee parameter `max_predicate_opcode` is also proposed, as it appears to be unstewarded, is currently set to a value that prevents utilization of one of the three existing predicates, and is not in line with current conventions for activating new features.

# Motivation

This BSIP is motivated by a multitude of potential use cases in which conditional validity of a transaction may be valuable.  BitShares currently supports conditional validity of a transaction via the `assert` operation which will cause a transaction to fail validation if a supplied “predicate” is not met.  However, at the present time, the assert operation supports only three distinct predicate types.  We propose here additional predicates which will enable new use cases. 

One new use case which could be enabled by the new predicates is the implementation of payment channels described in [Short-lived Unidirectional Payment Channels](bsip-0063.md), which depends on refund transactions being secured by a time lock.  A time lock could be trivially implemented as an assert on a particular head block time.

Another case where new predicates for conditional validity would be useful is for transactions which must be signed without the ability to verify blockchain state at the time of signing.  A particular instance of this is transactions signed by a hardware wallet.  Many hardware wallets display details of a transaction to the user for confirmation prior to signing.  Since BitShares transactions refer to accounts and assets by object ID rather than name or symbol, a hardware wallet cannot present these details to the user in human-readble form unless there is an assurance of a truthful mapping between the displayed account name or asset symbol and the corresponding object IDs in the transaction to be signed.  Including an assert operation will ensure that if the mapping is NOT truthful, then the transaction will NOT be valid to include in a block, thus hardware wallet UX can be greatly improved.

In what follows we propose to extend the list of currently available assert predicates with new predicates to serve the above use cases and more.

# Rationale

## Review of existing predicates

[...]

PR #183, opened 2019-07-17
PR for BSIP 69 "Assert Predicates"

christophersanborn:
Discussion: #175

### DIFF
--- a/bsip-0069.md
+++ b/bsip-0069.md
@@ -158,13 +158,13 @@ struct head_block_time_ge_time_predicate
 
 ### Activation of new predicates and removal of max_predicate_opcode parameter:
 
-A hard-fork date shall be chosen for activation of these new predicates and shall be sufficiently future-dated to allow time for nodes to upgrade.  The outdated committee parameter `max_predicate_opcode` shall be removed, as it is not needed for safe upgrade (see next section for discussion).
+A hard-fork date shall be chosen for activation of these new predicates and shall be sufficiently future-dated to allow time for nodes to upgrade.  The outdated committee parameter `max_predicate_opcode` shall be disabled, and, if possible without impacting history, removed, as it is not needed for safe upgrade (see next section for discussion).
 
 # Discussion
 
 ### Removal of max_predicate_opcode parameter:
 
-The set of committee parameters currently includes a parameter called `max_predicate_opcode`, which in the current protocol must be increased in order to activate new predicates.  The parameter is not a necessary protection in the protocol upgrade procedure, as the hard-fork mechanism is already both sufficient and necessary, making this parameter redundant.  Nor is there any clear justification for involving the committee in the mechanics of protocol upgrade.  In fact, the parameter appears to have already been forgotten, as it was not incremented after the addition of the most recent assert predicate, with the result that `block_id_predicate`, while defined and supported in the protocol, is not in fact usable on the network.  It seems quite unlikely that this was an intentional decision.  Thus we think the removal of this parameter and the associated validation checks are justified.
+The set of committee parameters currently includes a parameter called `max_predicate_opcode`, which in the current protocol must be increased in order to activate new predicates.  The parameter is not a necessary protection in the protocol upgrade procedure, as the hard-fork mechanism is already both sufficient and necessary, making this parameter redundant.  Nor is there any clear justification for involving the committee in the mechanics of protocol upgrade.  In fact, the parameter appears to have already been forgotten, as it was not incremented after the addition of the most recent assert predicate, with the result that `block_id_predicate`, while defined and supported in the protocol, is not in fact usable on the network.  It seems quite unlikely that this was an intentional decision.  Thus we think the removal or disabling of this parameter and the associated validation checks are justified.
 
 # Summary for Shareholders
 
--- a/bsip-0069.md
+++ b/bsip-0069.md
@@ -73,8 +73,12 @@ These predicates are useful to hardware wallets because HW wallets must sign tra
 
 #### account_authorities_match_predicate:
 
-Pseudocode:
 ```c++
+/**
+ * This predicate takes one or more authority or special_authority objects and
+ * asserts that each supplied authority matches the corresponding account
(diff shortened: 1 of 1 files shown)


bsip-0069.md
```diff
@@ -0,0 +1,165 @@
...
+   bool validate() const;
+};
+
+```
+
+# Discussion
+
+### Impacted committee parameters
+
+An affected committee parameter is `max_predicate_opcode`, which limits the number of defined predicates supported by the assert operation.  If new predicate types are added to the core protocol, this parameter will need to be increased in order to “activate” the new predicates on the network.  Thus, the committee should be ready to increase this parameter if this BSIP finds support and is implemented.  Using a committee parameter to activate protocol additions after block-producer adoption is an alternative to setting a hard-fork date for activation.
```
pmconrad: Nice find.

IMO we should do away with that parameter. We still will have to add hardfork protection because old nodes can't deserialize the new predicates, and the existing check does not prevent new assertions from being included in proposals.

The parameter currently has value "1" but we have three predicates defined (see above). The implementation of the third (block_id) is incomplete, which is probably the reason why that parameter exists. Question is - do we want to remove the broken block_id predicate, or fix it? @abitmore ?

bsip-0069.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Head block time predicate:
+
+We propose to add the following predicate, which supports applications such as payment channels:
+
+| Predicate | Description |
+|:----------|:------------|
+| `head_block_time_ge_time_predicate` | Head block time meets or exceeds `time`. Ensures a transaction is only valid at or after a specified time. |
+
+This predicate allows for the signing of future-dated transactions (up to the `maximum_time_until_expiration` limit set by committee).  Payment channel constructs, such as proposed in [BSIP-0063](bsip-0063.md), can use this to sign refund transctions to be used if a counterparty fails to meet an obligation to close out a channel within an agreed upon time window.
```
pmconrad: The first sentence is wrong. Perhaps instead
> This predicate allows for the pre-signing of future-dated transactions (within the limits of the TAPOS range, i. e. 65k blocks).

christophersanborn: Interesting; I had assumed `maximum_time_until_expiration` would be w.r.t. the referenced TaPoS block.  But now finding the code [here](https://github.com/bitshares/bitshares-core/blob/e8567d0425ec27d29036a9a9178df5afd8ca6f25/libraries/chain/db_block.cpp#L668:L672) it does seem it's w.r.t. the head_block_time at the time the transaction is seen by the network and applied to a block.  This actually extends the usable life of a payment channel.  Great!

Thanks @pmconrad.  Will revise.

As an interesting aside:  This actually provides a way to have an implicit head_block_time assertion without actually having the head_block_time assert predicate:

One could simply create a transaction with an expiration time of `maximum_time_until_expiration` + `intended_validity_time`, and then that transaction could NOT be included in a block until `intended_validity_time` arrives.  (Still subject to the TaPoS constraint, of course.)

**Q: Does this negate the value of having the predicate explicitly available?**   I think it's still valuable to have, for the clarity it provides.  (Should make it easier for a wallet managing a payment channel to validate the construction of a refund transaction.)  Also it provides protection against the chance of the `maximum_time_until_expiration` parameter changing within the lifetime of the channel and therefore messing with the trigger times, which I think is important.  But aside from that risk... it does now seem that the predicate duplicates an existing (implicit) capability.  Thoughts?

pmconrad: `maximum_time_until_expiration` is configurable by the committee. The implicit assertion you describe is therefore not reliable, so the proposed new assertion is an actual improvement.

bsip-0069.md
christophersanborn: > Please add the removal of the check to the "Specification" section. What do you think @abitmore ?

Would like @ryanRfox and perhaps others' takes on this.  We had some discussion on this during a previous core call.  I feel a little apprehensive "stealing" a committee parameter away from them.  But... if it's reason for being is no longer relevant,... then I guess so.  Is it appropriate to "bundle" removal of this parameter into this BSIP?  Or should it be it's own?

ryanRfox: The predicates need to be approved by BSIP approval vote (Community) and protocol upgrade code (Block Producers), there fore I do not see the need for the Committee to also "activate" them. Removing this ability is intentional, but I feel reflects the process we currently use for all other protocol upgrades. Perhaps we should seek some comments from current Committee members such as @bitcrab @xeroc @roelandp @OpenLedgerApp @clockworkgr

xeroc: I agree with @ryanRfox 
The committee has nothing to do with blockchain features perse.

christophersanborn: Text has been updated to propose removal of `max_predicate_opcode`.

bsip-0069.md
```diff
@@ -156,15 +156,19 @@ struct head_block_time_ge_time_predicate
 
 ```
 
+### Activation of new predicates and removal of max_predicate_opcode parameter:
+
+A hard-fork date shall be chosen for activation of these new predicates and shall be sufficiently future-dated to allow time for nodes to upgrade.  The outdated committee parameter `max_predicate_opcode` shall be removed, as it is not needed for safe upgrade (see next section for discussion).
```
abitmore: Technically we're unable to remove the `max_predicate_opcode` field since it affects serialization of certain operation. We can have a hard fork to no longer use the parameter. And we can check whether the old guarding code can be removed, primarily due to proposals.

pmconrad: Perhaps "disabled" instead of "removed".

christophersanborn: Not sure I know the specifics enough as to why it can't be removed (I guess any historical transactions that updated the parameter would become invalid if it was removed?), but I updated to say "disable, and, if possible without impacting history, remove".  Hopefully this covers sufficient latitude for voting purposes.

ryanRfox:
I think that's it for edits now. I just need someone to give me "done" confirmation and I'll merge it.

Issue #182, opened 2019-07-17
New BSIP: Settlement protection fund


froooze:
The settlement fund is used to buy the BTS back for the reserve pool, when CR reaches settlement protection ratio.

##Inflow for settlement fund: 
- 0.02% market fee
- margin call fee (0.2%)
- settlement fund holder
- default settlement fund ratio (10%)

##Outflow parameters for settlement fund: 
### Reserve Pool & Settlement fund holder
- settlement protection ratio (1.05)
- target settlement protection ratio (1.06)
- unstake time (31 days)
- settlement fee (1%)
- daily BTS buyback percentage (0.2%)

###Settlement fund holder
Every owner of a debt asset can lock up a certain amount of security to be part of the protection settlement fund. When settlement is happening he gets BTS back and the settlement fee is payed buy the bad debt holder to the reserve pool. 

####Unstaking
Unstaking from the security fund is only possible, when there is the same proportional change in fund/debt. To change the ratio to a less security fund ratio, the unstake time has to pass, before the asset can be removed from the security fund.

####Default fund ratio
The asset owner can suggest a default value for the borrower to invest in the settlement fund. This should help to build some security for the borrower and help also to increase the acceptance for the asset holder.

###Daily BTS buyback percentage
When the market is very liquid and there was no use case for the settlement protection, the settlement fund, can be used to stabilize the BTS with a daily percentage buyback program with the non staking settlement fund.

abitmore:
Another related discussion is #164 .

froooze:
https://medium.com/aave/decentralized-lending-pool-dlp-protocol-launched-on-testnet-swap-between-fixed-and-variable-e72ae43992c
> **Decentralized insurance fund**, which will collect a small part of the pool income to cover losses in case of liquidation issues.

Why should we not do the same?

Inmortak:
[...]

PR #189, opened 2019-08-10
BSIP-70: Peer-to-Peer Leveraged Trading

pmconrad:
Initial PR created - we're at a point where we're abusing the issue discussion for versioning (see variant A/B), so it's about time to move this into git.
This PR is based on the current OP of #170 with some fixes and clarifications, plus a means to let issuers decide which markets should be opened for lending.


froooze:
I want to add some fundamental design improvements to this protocol:

- The CR of a margin position can only be increased when debt is reduced or collateral is increased. With a option to lock up the debt asset side by side of the margin position to increase CR, the debt must not be payed back to increase CR.
- Locking up the debt in a separated way for the margin position, no extra collateral is needed for the borrowed debt asset
- Collateral is only needed for the interest rate, which is payed in front
- Borrower can set a time limit how long he wants to borrow the debt asset
- Before the time limits ends, the borrowed debt asset is reduced linear to avoid jumps in CR
- Supply/Demand for the lending pool does affect the interest rate
https://medium.com/hydro-protocol/how-lending-pool-interest-rates-actually-work-375794e71716
- The amount of debt which can borrowed from the lending pool depends on the amount of BTS, which are locked in the specific margin position

MichelSantos:
> I want to add some fundamental design improvements to this protocol:

@froooze I appreciate the recommendations.  I think that you'll find many of the recommendations are already in the specifications.  I'll reply to each of your points.


>    The CR of a margin position can only be increased, when debt is reduced or collateral is increased.

Naturally the CR is always fluctuating with the valuation of the collateral.  Therefore I presume from your recommendation that you would like to place an additional restriction on what the borrower may do with the margin trading account.  I'll first describe how the current specifications constrain what the borrower may and may not do.

_Current Specifications_
The borrower may pay back the **entire** debt when [closing the position](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#process-loan-closure-initiation) (which instantaneously raises the CR to infinity).  The borrower may not pay back only portion of the debt under normal conditions.  Paying back part of the debt while still keeping the loan open could be a variation in the future.

The borrower _may not_ withdraw the loan asset _ever_ while the loan is open.  [The borrower may withdraw the tradable asset](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#withdrawals) **if and only if** the _remaining_ portolio, which _might_ consist of some loan asset _and_ some tradable asset, keeps the the post-withdrawal CR &ge; MCR.  Any withdrawal will reduce the CR.  If CR<sup>before</sup> &ge; MCR and if CR<sup>after</sup> &ge; MCR, it is still a reduction of the CR.

_Additional Specification_
I presume that your recommended constraint also comes with an implied ability to partially repay the debt while keeping the loan/margin position open.  Therefore a successful leveraged trader may have a portfolio increasing in valuation with an increasing CR but will never be able to withdraw his profits _without first reducing some of the debt_.  This is a reasonable constraint.  But given that there is already a restriction on withdrawals to respect the MCR, how important do you think this additional constraint is?


>    With a option to lock up the debt asset side by side of the margin position to increase CR, the debt must not be payed back to increase CR.

You might be pleased to learn that the current specification (a) does not permit partially paying back the debt, and (b) [does not permit withdrawing the loan asset](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#withdrawals) from the portfolio.  So I think that this recommendation is already in the current specification.


>   Locking up the debt in a separated way for the margin position, no extra collateral as security is needed

This concern can be handled with two features of the current specification:

(a) the [trading limit](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#trading-limit) prevents the borrower from trading away all of the loan asset to ensure that the portfolio always has enough the loan asset such that

B<sub>liquid</sub><sup>after</sup> &ge; K = (MCR - 1) &times; B<sub>0</sub>

(b) therefore a concerned lender can be sure that the portfolio always has enough loan asset to repay the lender by [creating a loan offer](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#process-loan-offers) with an MCR of 2+x (or 200+%).

 
>    Collateral is only needed for the interest rate, which is payed in front

Alternate collateral assets is certainly an interesting possibility for a future version of this proposal.


>    Borrower can set a time limit how long he wants to borrow the debt asset

This is already present in the specifications with

- [minimum duration of the loan](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#borrow-offer-minimum-duration)
- [maximum duration of the loan](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#borrow-offer-maximum-duration)
- [early repayment of the loan](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#loan-duration) as long as the early repayment is not earlier than the [lender's minimum duration of the loan](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#loan-offer-minimum-duration)


>    Before the time limits ends, the borrowed debt asset is reduced linear to avoid jumps in CR

I don't understand this idea.  Even without this recommendation, why would there be jumps in the CR?

With this idea, what benefit is provided?  Is this recommendation a form of automatic partial repayment of the debt?


>    Supply/Demand for the lending pool does affect the interest rate for debt asset
>    https://medium.com/hydro-protocol/how-lending-pool-interest-rates-actually-work-375794e71716

The current specification allows every lender and borrower to come to an agreement to any fixed interest rate independent of the interest rate that other parties are supplying/demanding.  There is no lending _pool_.  All of the various participants will discover the interest rate for any pair of assets and that interest rate will vary over time.

Furthermore that interest rate will almost definitely be different for different loan durations.

Is your recommendation to have a separate leveraged trading where a central entity or a single contract is changing the loan offers depending on the demand?  If  that is case then I submit that people can build their own bots or hire fund managers to dynamically adapt their offers on this new order book system.


>    The max amount of debt which can borrowed from the lending pool depends on the amount of BTS, which are locked in the specific margin position

I must reiterate that _there is no lending pool_.  Lenders each make their own lending offers which are placed into an offer book.  Each matching of a loan is a match with another counter-offer.

Furthermore the loan asset can be any asset: BTS, bitCNY, GATEWAY.X, etc.

Is your recommendation to treat BTS as a special collateral asset that _must_ be used as collateral?  Or that _can optionally_ be used as collateral?


>    Option to borrow 0-100% from the lending pool when CR < MCR

The current specification is voluntary lending between two accounts.  There is no pool.  The lender picks a minimum MCR.  The borrower must provide a CR &ge; any compatible MCR.

By definition this is impossible.


>    Prioritize the margin calls for the interest rate

The current specification has two rules for [automatic matching of offers](https://github.com/bitshares/bsips/blob/bsip-70/bsip-0070.md#automatic-matching-offer-selection).  Both rules prioritize interest rate _of compatible offers_ above all other parameters.

froooze:
@MichelSantos: 
1. The actual version of the lending protocol is decoupled from the margin positions, which limits the efficiency. 
2. The borrower of the debt asset has no intention to trade, but to increase his CR to reach MCR.
3. The lending pool has the advantage no agreement between two parties are needed. 
4. To make things more clear I start a complete write up in a separate topic, where we can discuss the details, which are now not as clear for you as I thought.

Inmortak:
What's up guys!

I haven't read the whole thing for it's pretty long and complex. One question though, would this help to solve the current sell wall on bts/bitUSD pair?

froooze:
> I haven't read the whole thing for it's pretty long and complex. One question though, would this help to solve the current sell wall on bts/bitUSD pair?

The current version of this protocol does not solve the fundamental problems.
My approach does reduce the margin calls by about 95%.  
The trick is to borrow bitUSD to increase CR, without paying back the debt.

dls-cipher:
What about rules on market conditions for the assets ? 
Is there any protection against UIA issuer rigging market price in order to protect users with loan ? 
Price for gateway assets are determined how ? By internal market price (DEX) or through external price feeds ?

Chee®s

dls-cipher:
> What's up guys!
> 
> I haven't read the whole thing for it's pretty long and complex. One question though, would this help to solve the current sell wall on bts/bitUSD pair?

Not really.

froooze:
> In my humble opinion not really. It would just moved more bitUSD to some other assets instead of providing more liquidity for the mentioned pair.

Yes, here is a liquidity conflict. I can only borrow the debt asset once. For non-debt assets this makes sense, but for debt-assets this approach is not constructive.

froooze:
I see a liquidity conflict with the margin position liquidity pool:
https://github.com/bitshares/bsips/issues/182

pmconrad:
@froooze @dls-cipher @inmortak I think you may be confusing this leveraged trading proposal with our existing smartcoins. There is no direct connection between the two. Neither trade asset nor debt asset need be a smartcoin. Even if one is, the other need not be its backing asset. E. g. it would be possible to lend bitUSD for trading against bitCNY, or lending BTS for trading against OPEN.BTC.

froooze:
> @froooze Efficiency of what objective or process? I believe that your objective, although related, is different than this BSIP's objectives.

Yes, the #189 approach is not design to save the margin positions, but to enable lending.  

> I think that your points here and the objective of your proposal in #182 are valuable and interesting but they are specifically focused on the CR of smartcoins

The only important thing I have now, is the prioritization of the `margin position liquidity pool` for debt assets, when there is demand from the margin positions. When interest rate for `margin position liquidity pool` rises to a certain level, user can only lend the debt asset to the `margin position liquidity pool`.

froooze:
How can BSIP-70 save the margin position ? Can you provide an example?

MichelSantos:
>How can BSIP-70 save the margin position?
>Can you provide an example?

(@froooze I will presume that "saving the margin position" means to increase the CR of "bad debt" or to reduce some of the bad debt.  Please tell if you intended some other meaning.)

It will not.  That is not its primary objective.  BSIP-70 cannot be used directly by a debt holder who has no other assets, goods, or services to help their difficult position where the collateral has lost value.

All legitimate ideas for "saving the margin position" involve a rescue by people
(a) with assets, goods, or services _separate from the collateral already in the margin positions_ ("other assets")
(b) who **anticipate** an increase of demand for the collateral (relative to the debt asset) _in the future_

Without these two conditions, the bad debt will only get worse and there will be no saving of margin positions.

What a "rescuer" _could do_ depends on her risk tolerance, and the confidence of her predictions.

_BSIP-70 **might** help save margin positions if a rescuer has a particular forecast_.  For example, if a rescuer anticipates with **high confidence** that the collateral asset type (e.g. BTS) will increase in value in 2 years but has **low confidence** about its price **during** the 2 years, she _could_:
1. Borrow bitUSD for BSIP-70 trading against BTS with a loan duration of 2.5 years
2. After the BSIP-70 loan is created, she immediately sells her bitUSD into the margin wall to obtain BTS; _this trade helps to reduce overall bad debt_.
3. **If** her predictions are correct she will be profitable before she needs to repay/close her loan within 2.5 years (she can always pay back earlier if she desires).

MichelSantos:
@jmjatlanta Please review three clarifications about early closure

pmconrad:
@MichelSantos it has always been my understanding that the minimum duration is agreed upon by both parties, and that early payback is not possible. (Payback can be initiated by the borrower while the minimum duration has passed but the maximum duration has not been reached yet.)

shulthz:
What‘s the preventive method for the situation which the network halted on a block last night？

pmconrad:
> What‘s the preventive method for the situation which the network halted on a block last night？

Chain halts are rare and typically short in relation to loans. They need no special consideration in this context.

jmjatlanta:
@MichelSantos @pmconrad Was the comment at https://github.com/bitshares/bsips/pull/189#issuecomment-528226576 resolved? I believe Peter is saying that minimum duration should not be set when the offer is created, but after negotiation between borrower and lender. The latest changes seem to indicate it is decided by the lender only.

pmconrad:
Hm, somehow Michel's latest push didn't invalidate your approval, then Ryan merged it. In a PM from Michel I understood that he'd prefer to honour the minimum duration as well.
We may need another PR to undo the last commit.

abitmore:
> somehow Michel's latest push didn't invalidate your approval

`Dismiss stale pull request approvals when new commits are pushed` is now enabled. It was not.

MichelSantos:
@jmjatlanta Peter is correct in that my latest update made the BSIP inconsistent with regards to early repayment.  I'm preparing an update to rectify.  I'll make a new PR and reference this one.

froooze:
Did you consider a option to use BSIP-70 through a lending pool, instead of a order book, to improve performance and user experience?

MichelSantos:
> Did you consider a option to use BSIP-70 through a lending pool, instead of a order book, to improve performance and user experience?

@froooze I find that unpooled lending allows every person to decide what risk, interest, and loan duration that they are comfortable with for borrowing or lending.  This is much more flexible and avoids contentious and continuous debates about the appropriate risk, interest, loan duration, and fairness when under-collateralized situations occur.

Anyone who is interested in pooled lending or pooled borrowing could use the mechanism from this BSIP and operate a "managed fund" through a single account.  A pooled lending pool, for example, could work as follows:

1. The managers create a single managed account
2. Investors deposit funds into managed account in accordance with the terms and conditions of that fund
3. The managers create loan offers from some of the investments in the managed account
4a. The mechanism described in the BSIP operates as described
4b. The managers must retain sufficient funds in the managed account, or add their own funds into the managed account, to cover withdrawals and interest that are described by the terms and conditions from Step 2.

Issue #181, opened 2019-07-17
BSIP73: Match Force-Settlement Orders with Margin Calls and Limit Orders


abitmore:
BSIP: 0073
    Title: Match force-settlement orders with margin calls and limit orders
    Author: Abit More <https://github.com/abitmore>
    Status: Draft
    Type: Protocol
    Created: 2019-06-09
    Discussion: https://github.com/bitshares/bsips/issues/181
    Worker: TBD

# Abstract

This BSIP proposes a protocol change to improve user experience (UX) of
force-settlements by trying to fill force-settlement orders
(*settle orders* in short) at a better price
and optionally fill them before expiration when certain conditions are met.

# Motivation

Force-settlements were designed for debt asset holders to convert an amount
of the debt asset into corresponding collateral asset at a fair price *when
there is nobody willing to buy back the debt at a fair price*.

To mitigate malious behavior and market manipulation, a *delay* and a *price
offset* were designed. But the mechanism has flaws.
* If a positive *offset* is configured in the debt asset's options,
  force-settlement requesters will always "buy expensive" even when there are
  orders "selling low". This also leads to occasional spikes in the market
  history charts.
* Force-settlement requesters have to wait for the *delay* even when there
  appear traders buying the debt asset at a fair price in the market.

These flaws have led to certain confusion and anger among market participants.

# Rationale

Actually, when a margin call appears, it means there is somebody
willing to buy back the debt. Thus, it makes sense to fill the margin calls
with the settle orders.

Similarly, if there are limit orders selling the debt asset
below the feed price,
it makes sense to match the limit orders with the settle orders as well.

When one of these opportunities appears, it makes sense to fill certain
settle orders immediately if it's desired by the owners of the settle orders.

These changes would improve user experience (UX).

# Specifications

## Protocol upgrade

A time will need to be scheduled for applying the change. In this document,
terms "before the protocol upgrade", "at the protocol upgrade" and "after
the protocol upgrade" may or may not be used to indicate things happen before
the scheduled time, at the scheduled time and after the scheduled time.

## `settle_order_object`

The `settle_order_object` stores current status of a force-settlement order.

A new field is needed within it:

* `bool fill_asap;`

By default this field is set to `false`, which means the order will be
processed after the *delay* defined in the asset's options, which is current
behavior.

If this field is set to `true`, the order will be filled or partially filled
when a debt position enters margin call territory. It will also be filled or
partially filled when someone places a limit order selling the collateral
asset below the feed price.

[...]

PR #200, opened 2019-08-25
BSIP73: Match force-settlement orders with margin calls and limit orders

abitmore:
PR for #181.

### DIFF
--- a/bsip-0073.md
+++ b/bsip-0073.md
@@ -1,4 +1,4 @@
-    BSIP: TBD
+    BSIP: 0073
     Title: Match force-settlement orders with margin calls and limit orders
     Author: Abit More <https://github.com/abitmore>
     Status: Draft
--- a/bsip-0073.md
+++ b/bsip-0073.md
@@ -253,9 +253,13 @@ https://github.com/bitshares/bitshares-core/issues/1958)).
 It is probably harmful for performance in the long run, which is particularly
 undesirable for functionality that can mostly be emulated client-side.
 
-# Summary for Shareholders
+# Non-Technical Summary
 
-TBD
+When force-settling a SmartCoin, the user currently has to wait for the
+*settlement delay* before her tokens are exchanged for the collateral asset.
+This BSIP introduces a new flag that allows settlement requests to be matched
+with market orders during the waiting period, potentially resulting in faster
+settlement and a better price.
 
 # Copyright
 
--- a/bsip-0073.md
+++ b/bsip-0073.md
@@ -108,7 +108,7 @@ The `proposal_create_evaluator` is used to evaluate and apply the
 `asset_settle_operation` objects. New logic is needed:
 * only allow `fill_asap` to be set after the hard fork.
 
-## Matching and filling settle orders before the delay
+## Matching and filling settle orders whose `fill_asap` field is `true`
 
 ### When a new limit order is created
 
@@ -181,6 +181,18 @@ After the hard fork, when processing a settle order after the delay,
 * if the settle order still exists, process it with the logic before the
   hard fork.
   Note: this step does affect `force_settled_volume` as before.
+
+## Logic related to BSIP 71 (Prevent Global Settlement)
+
+The behaviors proposed in this BSIP would be impacted by BSIP 71.
+
+In general, this BSIP proposes that
+* the settle orders whose `fill_asap` field is `true` will be treated as
+  limit orders buying at feed price when the market engine processing
+  limit orders or margin calls, and
+* when processing a settle order after the delay, it will "buy the way up"
+  before being matched with a debt position that need to apply
+  `force_settlement_offset`.
 
 ## API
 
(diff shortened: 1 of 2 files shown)


bsip-0073.md
```diff
@@ -0,0 +1,210 @@
...
+field.
+
+APIs which return combined order book can combine settle orders whose
+`fill_asap` is true with limit orders in the same direction.
+
+## CLI
+
+There was a `settle_asset` command in CLI.
+
+Need to add a new command E.G. `settle_asset_ext` in CLI so that users would
```
pmconrad: How about adding an optional boolean flag to the parameter list of `settle_asset`?

bsip-0073.md
```diff
@@ -0,0 +1,210 @@
...
+as special buy orders which are buying at feed price in the order book.
+
+# Discussion
+
+With this BSIP, we provided a tool that can be used by debt asset holders to
+convert their debt asset to collateral asset more conveniently and flexibly.
+However, it's not guaranteed that a settle order will be filled at better
+price if the owner chose to fill it "as soon as possible". Market paticipants
+should always make their own decisions on whether to use the new tool.
+
```
pmconrad: Please add these discussion points:

* From the perspective of the chain it is more desirable to execute settlements against margin calls than against market orders, because it improves the collateral ratio of these. In that sense it contradicts the goals of https://github.com/bitshares/bsips/pull/193 .
* The proposed functionality is mostly convenience for users. Manual selling on the market is (almost) always available when it would be used automatically by this proposal (exception: new incoming limit orders).
* The proposed functionality complicates the market engine and possibly the get_order_book call (see https://github.com/bitshares/bitshares-core/issues/1958). It is probably harmful for performance in the long run, which is particularly undesirable for functionality that can mostly be emulated client-side.

bsip-0073.md
abitmore: Need to test whether this approach would break existing client applications which relies on wallet API. I know node API supports optional parameters already, but wallet API is a bit different than node API.

bsip-0073.md
abitmore: > From the perspective of the chain it is more desirable to execute settlements against margin calls than against market orders

I think this is not really from the perspective of the chain, but rather from individual asset owners for the assets' performance, although some MPAs in BitShares are owned by committee-account.

IMHO the UX improvement is more important.

#193 is mostly about positions whose CR < MCR. For this BSIP,
* if no debt position's CR is below MCR, and there are limit orders selling below `feed_price * (1+offset)`, IMHO it doesn't make sense to force the debt position to increase CR;
* if a debt position's CR is below MCR,
  * if no limit order is selling below MSSP, the debt position would be matched and filled at MSSP (but not at `feed_price*(1+offset)` as pre-HF), which would likely increase CR of the position;
  * if there are limit orders selling below MSSP which caused margin calls unable to be filled, I think a bidding mechanism is better for increasing overall CR.

bsip-0073.md
```diff
@@ -74,91 +84,98 @@ an `extensions` field:
...
+it doesn't affect `force_settled_volume` (which indicates how much of the debt
+asset has been force-settled in the current maintenance interval).
 
 ## Processing settle orders after the delay
 
 Currently, when filling a settle order after the delay (note: it implies
-some conditions E.G. price feed is valid and total settled volume in current
-maitenance interval doesn't exceed maximum allowed volume),
+some conditions are met, E.G. the feed price is valid and total settled volume
+in the current maitenance interval doesn't exceed the maximum allowed volume),
```
pmconrad: Typo: mai*n*tenance

bsip-0073.md
```diff
@@ -168,34 +185,59 @@ After the hard fork, when processing a settle order after the delay,
...
+option.
 
-Need to add a new command E.G. `settle_asset_ext` in CLI so that users would
-be able to create force-settlement orders with the new `fill_asap` option.
+One option is to add an optional boolean parameter to the parameter list of
+the existing `settle_asset` command, if it doesn't break existing client
+applications which rely on that wallet API. Otherwise, a new command is
+needed, E.G. `settle_asset_ext`.
 
 ## GUI/UX
```
pmconrad: Perhaps add a note that UI changes are only suggestions and formally not part of the specification.

pmconrad:
Thanks.
FTR, my approval only means that I think this BSIP meets formal requirements. I still think it does not add sufficient value to justify the cost (performance impact, more complicated market engine, implementation cost).

pmconrad:
Suggested summary:

----
When force-settling a SmartCoin, the user currently has to wait for the *settlement delay* before his tokens are exchanged for the collateral asset. This BSIP introduces a new flag that allows settlement requests to be matched with market orders during the waiting period, potentially resulting in faster settlement and a better price.

abitmore:
@pmconrad thanks. Please review again.

sschiessl-bcp:
> FTR, my approval only means that I think this BSIP meets formal requirements. I still think it does not add sufficient value to justify the cost (performance impact, more complicated market engine, implementation cost).

Should this be incorporated in the summary, or do we rather not burden the voters with this technical aspect?

abitmore:
>>FTR, my approval only means that I think this BSIP meets formal requirements. I still think it does not add sufficient value to justify the cost (performance impact, more complicated market engine, implementation cost).
>
> Should this be incorporated in the summary, or do we rather not burden the voters with this technical aspect?

Create a PR please? Thanks.

PR #201, opened 2019-08-26
Update README: Add BSIP73 and Update Reference for BSIP72

ryanRfox:


abitmore:
@ryanRfox I dropped "Immediately" from the title of BSIP73 because it's only optional.

ryanRfox:
Changed this PR to include additional updates to the README.

PR #208, opened 2019-09-10
Create bsip-0074.md

bitcrab:


bsip-0074.md
```diff
@@ -0,0 +1,36 @@
...
+As the margin call fee ratio may be obvious higher than market fee, the margin call orders should be placed at the real price without margin call fee to avoid misleading the buyer.
+
+# Specification
+Add one new parameter Margin Call Fee Ratio(MCFR) for each smartcoin, which is controlled by the smartcoin owner.
+
+`Margin call order price = settlement price/(MSSR-MCFR)`.
+
+Here settlement price is a new introduced parameter which is defined in BSIP71, settlement price = feed price when there is no bad debt.
+
+When a margin call trading happens, the buyer sells smartcoin with quantity X and get collaterals in quantity `X*(MSSR-MCFR)/settlement price`, the margin call order owner sells collaterals in quantity `X*MSSR/settlement price` and get smartcoin in quantity X, the delta in paid and received collaterals in `quantity X*MCFR/settlement price` will be paid to system as margin call fee.
```
pmconrad: Please explain what "paid to system" means, i. e. where the fee goes.
Note that the collateral asset is not necessarily BTS, i. e. it can not generally go into the reserve pool nor into the fee pool of the asset.

bitcrab: oh, yes, it should be "paid to the owner of the smartcoin".

pmconrad: Fees are usually paid into a vesting balance from which the owner can later claim them. Would that be ok here too?
(The reason is that it simplifies account history. Without the vesting balance you'd have an entry for each individual fee payment, which can be a lot.)

bitcrab: that's ok.

shulthz:
I think maybe need to add two options：

1.The fee can share a part of it (e.g. like 20%) to the margin call buyer which can be decided by the asset owner.
These shared fees must vest for a total of 60 days.

2.The assets owner have the choice to decide which kind fee they want to charge, the Smartcoin or Backing asset.

PR #209, opened 2019-09-10
Remove SHA1

jmjatlanta:
This PR will request removal of the SHA1 algorithm from HTLCs as par of BSIP 64.

### DIFF
--- a/bsip-0064.md
+++ b/bsip-0064.md
@@ -10,13 +10,13 @@ Discussion: https://github.com/bitshares/bsips/issues/163
 
 Abstract
 ===================
-This BSIP proposes three changes to Hashed Time Lock Contracts (HTLC) defined in [BSIP 44](https://github.com/bitshares/bsips/blob/master/bsip-0044.md). The first makes the preimage `length` parameter optional. The second adds the HASH160 algorithm to the set of allowed hash functions. And the third creates an optional `memo` field.
+This BSIP proposes three changes to Hashed Time Lock Contracts (HTLC) defined in [BSIP 44](https://github.com/bitshares/bsips/blob/master/bsip-0044.md). The first makes the preimage `length` parameter optional. The second adds the HASH160 algorithm to the set of allowed hash functions. The third removes the SHA1 algorithm. And the fourth creates an optional `memo` field.
 
 Motivation
 =================
 The original implementation of Hashed Timelock Contracts included a protection to prevent an attack based on the size of the preimage. While it is still believed this is a valuable feature, this makes the implementation incompatible with a strict implementation of the popular [BIP199](https://github.com/bitcoin/bips/blob/master/bip-0199.mediawiki) implemented by Bitcoin.
 
-In addition, BIP199 includes the ability to use HASH160, which is not currently part of the BitShares implementation.
+In addition, BIP199 includes the ability to use HASH160, which is not currently part of the BitShares implementation. SHA1 is not part of BIP199, but was implemented although not recommended. It is best to remove SHA1.
 
 And finally, when participating in atomic cross-chain swaps (a major benefit of HTLCs), some extra information is necessary. An optional `memo` field should be added.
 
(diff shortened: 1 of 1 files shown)


bsip-0064.md
```diff
@@ -10,13 +10,13 @@ Discussion: https://github.com/bitshares/bsips/issues/163
 
 Abstract
 ===================
-This BSIP proposes three changes to Hashed Time Lock Contracts (HTLC) defined in [BSIP 44](https://github.com/bitshares/bsips/blob/master/bsip-0044.md). The first makes the preimage `length` parameter optional. The second adds the HASH160 algorithm to the set of allowed hash functions. And the third creates an optional `memo` field.
+This BSIP proposes three changes to Hashed Time Lock Contracts (HTLC) defined in [BSIP 44](https://github.com/bitshares/bsips/blob/master/bsip-0044.md). The first makes the preimage `length` parameter optional. The second adds the HASH160 algorithm to the set of allowed hash functions. The third removes the SHA1 algorithm. And the fourth creates an optional `memo` field.
```
ryanRfox: Fixup: "...proposes **four** changes..."

pmconrad:
Are there any other chains out there that we'll no longer be able to HTLC with? IOW they have SHA1 but no other hash algorithm that is compatible with us?

Technical note: we cannot simply remove SHA1 from the `htlc_hash` static_variant. We'll have to forbid its usage in the hardfork, and if it isn't used until then we can later replace it with an invalid dummy. Code-wise this will be much uglier than simply keeping SHA1.

pmconrad:
Also note that SHA1 is broken in that it is practical to produce two documents that have the same hash value.
However, this would not allow an attack against HTLC. SHA1 in HTLC would be insecure if it was possible to produce a plaintext for any chosen SHA1 hash value. I mean, we still should discourage its use, but I don't see an urgent need to actually remove it.

ryanRfox:
Perhaps it is best to leave it in. Suggest to note the risks in documentation. (Out of scope for this PR: Also suggest @nathanhourt not use SHA1 in the TNT design.)

nathanhourt:
> Are there any other chains out there that we'll no longer be able to HTLC with? IOW they have SHA1 but no other hash algorithm that is compatible with us?
> 
> Technical note: we cannot simply remove SHA1 from the `htlc_hash` static_variant. We'll have to forbid its usage in the hardfork, and if it isn't used until then we can later replace it with an invalid dummy. Code-wise this will be much uglier than simply keeping SHA1.

Well now... that depends... if, by the time of the hardfork, no one ever *used* the SHA1 HTLC, then we *could* just quietly rewrite that type tag to reference HASH160... it would be an epic revisionist hack, but it would make the code prettier :D

pmconrad:
>  if, by the time of the hardfork, no one ever used the SHA1 HTLC, then we could just quietly rewrite that type tag to reference HASH160

Hehe, yes, that would work. We could create a softfork in advance to prevent SHA1 from being used until the hardfork. With the softfork this would be a risky undertaking.

abitmore:
>We could create a softfork in advance to prevent SHA1 from being used

This sounds like a good idea. Which version should we sneak it in?

pmconrad:
The softfork would be a witness-only update. I believe they are running various 3.x versions so we'd have to create another set of patch versions.

bsip-0064.md
jmjatlanta: Done

jmjatlanta:
> Who is willing to confirm SHA1 has **not** been used thus far? I support a soft fork in this case.

Working on this.

pmconrad:
Thanks. So SHA1 stays in, technically. No softfork. Next question is, do we forbid using SHA1 in the future? I don't see a reason to do so.

ryanRfox:
Let's leave it in. Suggest we note the potential risks to users if using this hash function in our documentation. Ready for close.

Issue #164, opened 2019-05-02
BSIP74: Margin Call Fee Ratio


shulthz:
When the margin call was closed by the market，the system can charge the shorter penalty fees from the transaction of margin call.

The margin call penalty fees will place in a insurance fund.

The penalty fees can be  0.1% ~ ?%（Adjusted by the community）. 

This will encourage the shorter to close the margin in time.
This wll suppress the aggressive shorter partly to reduce the global systematic risk.
The insurance fund will play a role in some place.
MSSR still will be MSSR.

When the shoter was be margin called, he will sell the bts to get the BITASSET to close the position, he sold 1 bts and got 0.4 bitcny, so the system charge 10% from the 0.4 bitcny as the penalty fees.

pmconrad:
Thanks.

IMO it makes little sense to open many issues here with individual ideas about how to optimize MPAs. Better discuss elsewhere and come up with **one** BSIP containing the results of the discussion.

abitmore:
IMHO charging a fee would cause less margin calls being bought thus lead to a bigger premium in a downtrend. Due to the impossible trinity, we have to accept trade-offs. If we want to improve one side, the other side would get weaken.

For optimizing MPAs, perhaps we can wait for @biophil's project: https://bitsharestalk.org/index.php?topic=28340.0 and/or work along with his team.

Adding more parameters so the community can tune is generally good IMHO, more tools means greater flexibility. However, more tools also means greater complexity, we should have guidelines about how to tune the parameters, when should tune what and etc.

shulthz:
Charge the fees was independent，MSSR still is 1.02, it wouldn't cause less margin calls being bought.

The research didn't have a exact name，maybe one month, maybe one year, and the result need to test and will change too much.

Some project of MPAs have charged this fee for a long time, didn't find any problems, i think we can learn somethings from it and do what we can do now.

abitmore:
If to charge extra fee from collateral, that means the debt position would loss more collateral (than current rule) when it's partially filled, that means it's harder for the debt position to return to a healthy CR, that means stronger sell pressure, that means it's closer to black swan.

shulthz:
> 
> 
> If to charge extra fee from collateral, that means the debt position would lose more collateral (than current rule) when it's partially filled, that means it's harder for the debt position to return to a healthy CR, that means stronger sell pressure, that means it's closer to black swan.

We didn't charge extra fee from collateral, we charge the fee from the transaction of margin call.

[...]

Issue #156, opened 2019-03-28
BSIP62: Close short position (i.e. manual margin call)


sschiessl-bcp:
BSIP: 0062
    Title: Close short position (i.e. manual margin call)
    Authors:
      Stefan Schießl <stefan.schiessl@blockchainprojectsbv.com>
    Contributors and Reviewers:
      Jerry Liu bitcrab@qq.com
    Status: Draft
    Type: Protocol
    Created: 2019-03-28
    Discussion: https://bitsharestalk.org/index.php?topic=28211.0
    Worker: Unpaid

# Abstract
Closing a short position comes with a risk of getting margin called if no spare funds are available to buy the long and reduce the debt to zero. This BSIP introduces a way to put the short position directly on the market.

# Rationale
Closing a short position with a healthy CR should come with no risk of getting margin called.

# Specification
Introduce a way that triggers something similar to a margin call. The user defines:
 - amount of short to sell
 - price of short/long

The price must be restricted such that the CR is not reduced when orders are matched against it. The order must appear as a normal order on the order book and not as a margin call. If the position is in a state of being margin called, this manual margin call will remain alongside the margin call.

This can be done by adding optional parameters to the update of a short position, or by addition of a new operation that activates the manual margin call (and deactivates on setting amount to 0)

# Discussion and Summary for Shareholders
Reducing the risk of a margin call is favorable for the network and the position holder, while there is no risk for the market or long holders in allowing to put the position on the market.

# References
- https://bitsharestalk.org/index.php?topic=28211.msg328509
- User custom price for his margin call order https://github.com/bitshares/bsips/issues/129

sschiessl-bcp:
Question:

Include in update of short position, or extra operation?

abitmore:
Also see #129.

sschiessl-bcp:
> Also see #129.

Adjusted the text to include the idea there.

bitcrab:
it's easy to ensure that the CR is not reduced with close short orders by limiting the selling price >margin call price/MCR.

how to coordinate this close short logic with the margin call logic?
in my view the margin call logic should take high priority, means if a position is margin called with "close short" orders, all the close short orders should be deleted and then the margin called orders will be placed.

if necessary(although not very necessary in my view) maybe we can introduce independent MSSR for each position, however independent MSSR should be equal or greater than global MSSR.

how to coordinate this close short logic with adjusting position?
[...]

PR #211, opened 2019-09-14
BSIP62: Close Margin Position

bitcrab:


ryanRfox:
Please add the Issue to the Discussion heading: https://github.com/bitshares/bsips/issues/156

sschiessl-bcp:
@abitmore 

from the mechanics, would it make sense to additionally allow that the user defines "sell X of the collateral, and set the price dynamically (like a real margin position) Y% below feed price"?

abitmore:
@sschiessl-bcp a dynamic price or a user-defined MSSR is "good to have" to me, but I think it would complicate the implementation (see also #51).

sschiessl-bcp:
> @sschiessl-bcp a dynamic price or a user-defined MSSR is "good to have" to me, but I think it would complicate the implementation (see also #51).

I would like to include it in this BSIP, with a note to postpone it to the second next protocol upgrade. Feasible?

bsip-0062.md
```diff
@@ -0,0 +1,43 @@
...
+  ```
+
+# Abstract
+Closing a short position comes with a risk of getting margin called if no spare funds are available to buy the long and reduce the debt to zero. This BSIP introduces a way to put the short position directly on the market.
+
+# Rationale
+Closing a short position with a healthy CR should come with no risk of getting margin called.
+
+# Specification
+Introduce a way that user can directly place orders to sell collateral in a margin position, and the received SmartCoins will be used to reduce the debt automatically.
```
pmconrad: A possible implementation would be an extension to `limit_order_create`that points to the debt position. This would be similar to the `limit_order` extension specified in BSIP-70 (trading from a loan portfolio).

bsip-0062.md
```diff
@@ -0,0 +1,43 @@
...
+
+User needs to define the amount to sell and amount to receive while placing the sell orders, and the 
+price should be greater or equal to margin call price/MCR, to ensure the deposit position CR will not be reduced after the filling of the sell orders.
```
pmconrad: I think MCR is not relevant. The price must be better than debt/collateral.

bsip-0062.md
```diff
@@ -0,0 +1,43 @@
...
+
+When a debt position is margin called with close short orders, 2 ways are possible to handle this:
```
pmconrad: Please specify only one way. The other can be part of a "Discussion" section, together with a reason why it wasn't chosen.

bsip-0062.md
```diff
@@ -0,0 +1,44 @@
...
+  Type: Protocol
+  Created: 2019-09-14
+  Worker: TBD
+  ```
+
+# Abstract
+Closing a short position comes with a risk of getting margin called if no spare funds are available to buy the long and reduce the debt to zero. This BSIP introduces a way to put the short position directly on the market.
+
+# Rationale
+Closing a short position with a healthy CR should come with no risk of getting margin called.
```
MichelSantos: I think that it would be helpful to any reader to expand this rationale to describe the particular process of closing a short position from the [the forum post](https://bitsharestalk.org/index.php?topic=28211.msg328509) this is considered risky.  The general process of closing a short position is not risky in itself.

bsip-0062.md
```diff
@@ -0,0 +1,44 @@
...
+
+# Specification
+Introduce a way that user can directly place orders to sell collateral in a margin position, and the received SmartCoins will be used to reduce the debt automatically.
+
+User needs to define the amount to sell and amount to receive while placing the sell orders, and the
```
MichelSantos: I recommend that this proposal should explicitly have a maximum amount (a cap) of collateral that can be offered.  The reason being that while a "close short position" remains on the books, enough collateral must be retained by the blockchain to honor any possible redemption requests by force settlements.

For example if a debt position's CR = 1.7x and the asset's MCR = 1.6x, the "close short order" should not be able to **initially** offer more than 0.1x worth of the debt (denominated in collateral asset) in an open order at a time.  Ideally the the size of the "close short order" could be dynamic to fluctuate with the valuation of the collateral; but this could be technically difficult and might be relatively computationally expensive.

Some might argue that initial amount could be 0.7x of the debt (denominated in collateral) but then this amount must certainly be dynamic because there would be no buffer for the blockchain to honor force settlements.

pmconrad: That is a very important point. Thanks!

sschiessl-bcp: > For example if a debt position's CR = 1.7x and the asset's MCR = 1.6x, the "close short order" should not be able to initially offer more than 0.1x worth of the debt (denominated in collateral asset) in an open order at a time.

To ensure all other mechanics work as intended it is sufficient to ensure this via the price. First come first serve also holds for Force Settlements (if this position would be indeed the one with least collateral, and the "close short position" clears it before the force settlement hits, the force settlement will simply take the next least collaterized one). Keep in mind that the force settlement offset actually benefits the margin position holder!

The ultimate restriction on the price is: `sell_price < CR * feed_price`. If it would be equal, it would not reduce your collateral ratio. The ultimate restriction on the amount is that the `receive_amount is <= debt_amount`. It would also be possible (and should be allowed!) to sell for less than feed price (making a profit while closing the position).

bsip-0062.md
sschiessl-bcp: Done

bsip-0062.md
sschiessl-bcp: Would this approach also allow a dynamic price through specifying the offset from feed_price and the target to receive amount?

bsip-0062.md
bitcrab: > In that sense a close position order would be automatically cancelled if force-settlement hits before (close short position may not protect against force settlement, but should also not be hindered by it, same as with an open margin call).

agree, means no need to limit the close short order quantity, however when force settlements hits, it will not prohibit the force settlement from executing.

abitmore:
No creation fee refund when the order is cancelled.

The order should be cancelled automatically if the asset is globally settled.

The order should be cancelled automatically if the position is manually closed.

The order should be cancelled automatically when remaining collateral in the position is less than the amount selling in the order.
* This can happen when an order is placed but then collateral price fell below the order's price and the position is partially margin called.
* This can also happen when the position is manually adjusted.
* This can also happen when the position is force-settled.

What if the order is asking for more than its debt, due to margin called or debt position updated?
* If "asking for more" is not allowed, how to deal with the order? Cancel it or shrink it?
* If "asking for more" is allowed, should we fully fill it, or only fill the amount it borrowed then cancel the remainder? Not cancelling the remainder means to turn the order into a normal limit order. Perhaps make this an option, because perhaps some traders want it, but others don't want it.

Will the order has an `expiration` too? And what about `fill_or_kill`?

bsip-0062.md
```diff
@@ -0,0 +1,44 @@
...
+# Rationale
+Closing a short position with a healthy CR should come with no risk of getting margin called. Currently the user has to take out collateral from the short position in order to put it on the market. Reducing the collateral increases the risk of margin call. This is especially cumbersome and risky if you want to close a position that has CR close to MCR (in that case you can only take out a small amount of collateral, sell it, reduce debt, repeat until position closed).
+
+# Specification
+Introduce a way that user can directly place orders to sell collateral in a margin position, and the received SmartCoins will be used to reduce the debt automatically.
+
+User needs to define the amount to sell and amount to receive while placing the sell orders, and the 
+price should be greater or equal to debt/collateral, to ensure the deposit position CR will not be reduced after the filling of the sell orders.
+
+When a debt position is margin called with close short orders,system delete all the close short orders and place margin call orders.
```
abitmore: IMHO the system should allow the user to sell collateral below MSSP. The system should allow the user to sell collateral at a very high price even if it won't be filled ever. That said, it's unnecessary to cancel the order when the debt position enters margin call territory.

bsip-0062.md
```diff
@@ -0,0 +1,44 @@
+ ```
+  BSIP: 0062
+  Title: Close short position
```
abitmore: This title is not good.

The BSIP is talking about a new option or operation to sell a debt position's collateral directly to pay the debt. The title should reflect the content.

bsip-0062.md
MichelSantos: > To ensure all other mechanics work as intended it is sufficient to ensure this via the price. First come first serve also holds for Force Settlements (if this position would be indeed the one with least collateral, and the "close short position" clears it before the force settlement hits, the force settlement will simply take the next least collaterized one). In that sense a close position order would be automatically cancelled if force-settlement hits before (close short position may not protect against force settlement, but should also not be hindered by it, same as with an open margin call).

My concern is over the proposed **dual-use** of a certain amount of an asset (in this case 1.0x worth of the debt in the form of the collateral asset): to honor force settlements, and to offer in a "close short position" order.

The proposed idea of automatic cancelling of a "close short position" if and when that debt position is force settled can financially work.  But it also make the order books _slightly misleading_ in a financial sense by making an order book appear deep with "close short position" offers _which are simultaneously backing_ potential redemption requests.

> The ultimate restriction on the price is: sell_price < CR * feed_price ...

I agree with the ideas about limiting the sell price to maintain or increase the CR

bsip-0062.md
```diff
@@ -0,0 +1,44 @@
...
+  Type: Protocol
+  Created: 2019-09-14
+  Worker: TBD
+  ```
+
+# Abstract
+Closing a short position comes with a risk of getting margin called if no spare funds are available to buy the long and reduce the debt to zero. This BSIP introduces a way to put the short position directly on the market.
+
+# Rationale
+Closing a short position with a healthy CR should come with no risk of getting margin called. Currently the user has to take out collateral from the short position in order to put it on the market. Reducing the collateral increases the risk of margin call. This is especially cumbersome and risky if you want to close a position that has CR close to MCR (in that case you can only take out a small amount of collateral, sell it, reduce debt, repeat until position closed).
```
MichelSantos: > Currently the user has to take out collateral from the short position in order to put it on the market.

This is not generally true.  It would be better to qualify this statement with something like the phrase that is used in the abstract.

"Currently a user, **who has no spare funds**, has to take out collateral from the short position in order to put it on the market."

bsip-0062.md
pmconrad: Probably not.

bsip-0062.md
sschiessl-bcp: > But it also make the order books slightly misleading in a financial sense by making an order book appear deep with "close short position" offers which are simultaneously backing potential redemption requests.

This would also be true for margin calls, right?

bsip-0062.md
sschiessl-bcp: Do you have an alternative suggestion? I find it still suitable as it is looking to exactly do this.

bsip-0062.md
sschiessl-bcp: Done

sschiessl-bcp:
All comments up till now have been addressed, resolved where possible. Please have a look!

bsip-0062.md
MichelSantos: > This would also be true for margin calls, right?

Touché :)

In that case the `sell_price` constraint is essential to ensure that a partial match of the "close short position" still allows the blockchain to honor possible redemption afterwards

sschiessl-bcp:
Considering the mentioned edge cases I am still wondering if it would not be easier to allow putting the order out with a % off the feed price (could also be additional). That would essentially merely re-use existing margin call logic.

bsip-0062.md
sschiessl-bcp: I suppose it's fine to have both ideas .. I added it to the BSIP

abitmore:
> merely re-use existing margin call logic.

It looks like that we can reuse the logic, but not that simple when coding. I agree to add it to the BSIP though, if it's too hard to implement, we can leave it out in the first release.

sschiessl-bcp:
@abitmore please approve if you agree this is ready for voting.

bsip-0062.md
```diff
@@ -0,0 +1,63 @@
...
+7. `expiration` must be defined
+
+## Manual margin call
+
+User needs to specify the manual margin call price ratio (value must be positive, and a value of 1 represents the feedprice) that will be used to calculate the sell price depending the feed price, and a manual margin call will be created that follows the logic of a margin call, except that the price is calculated from the user-defined percent instead of MSSR.
+
+# References
+https://bitsharestalk.org/index.php?topic=28211.msg328509
+
+User custom price for his margin call order #129
```
abitmore: This link does not work, please update.

abitmore: Done.

shulthz:
If want BSIP62 to become really useful?It must charge the MCFR fees; #164
If someone's CR<=MCR, and want to use BSIP62, the system will charge the MCFR fees from the user;

shulthz:
If this BSIP didn't charge fees when some body CR<ICR like BSIP74, it will become very foolish like before.
Marke will prove it in the future.

sschiessl-bcp:
I see no reason not too include it as well, can be added to BSIP74 specification.

shulthz:
> I see no reason not too include it as well, can be added to BSIP74 specification.

Maybe.
But this BSIP is very different with BSIP74, in some case, they are against, i think BSIP62 not only need MCFR fees, but also need some Liquidity penalty fee, as BSIP62 consume the depth and liquidity of BSIP74 advance.
If no margin call in the market, nobody want to use it, if somebody want use it, that's mean there will have a quick market volatility, he want to close his position in the market quickly, that's mean he will consume the depth and liquidity of the market advance, other margin call will not have enough liquidity and depth, if he can't close his position all, he will make a lower price than margin call price make sure the market can consume it all, that's very dangerous.

BSIP74 should have a function of share fees, as this, the Liquidity penalty fee of BSIP62 must adapt with the proportion of share fees.

Thank you.

Issue #150, opened 2019-02-22
BSIP61: Operation to Update Limit Orders


nathanhourt:
BSIP: 0061
    Title: Operation to Update Limit Orders
    Authors:
      Nathan Hourt <nat.hourt@gmail.com>
    Contributors and Reviewers:
      TBD
    Status: Draft
    Type: Protocol
    Created: 2019-02-22
    Discussion: https://github.com/bitshares/bsips/issues/150
    Worker: Unpaid

# Abstract
This BSIP proposes a new operation, `limit_order_update_operation`, which will allow direct modification of a limit order. This operation will be smaller to serialize than the cancel/create operation combination presently required to achieve the same result, as it must only store new field values rather than all field values. Moreover, the update operation will require lower processing and RAM consumption. In return for this reduction in consumption, the overall fee charged for an update operation can be set lower than the equivalent cancel/create combo, thus incentivizing market makers to keep freshly updated orders at the top of the books.

# Motivation
At present, the BitShares protocol does not allow direct modification of limit orders in the markets. Instead, to modify a limit order, the order must be canceled and and a new one created in its place. This results in greater-than-necessary overhead when updating a limit order, an action which is performed by market maker robots with great frequency (hundreds of times per day, per market). Additionally, the cancel/create combination requires payment of two fees, which has led to some pressure from market makers for BitShares to provide a cheaper mechanism for maintaining fresh order walls in the markets.

# Rationale
The specification below was chosen on the following criteria:
 - Minimal storage space for historical data (i.e. serialized operation size)
   - Cancel/create combination requires two distinct operations, and explicit restatement of all order parameters
   - Update operation requires only one operation, which elides all fields which are not being updated
 - Minimal processing
   - No deletion/construction is necessary, object is modified in place
   - Cancel/create requires two separate modifications of balances; update requires zero or one
   - Order matching is only triggered if order is moved past previous top of book
 - Minimal RAM consumption for undo storage
   - Object modification requires slightly less space than object deletion and creation
   - Balances and account statistics are not updated unless necessary, avoiding unnecessary undo storage
 - Finer granularity of fees
   - Low fee if only the order object is modified
   - Higher fee if balances must be updated as well

# Specification
A complete implementation of this BSIP is available for review [here](https://github.com/bitshares/bitshares-core/compare/develop...nathanhourt:issue-1604).

The `limit_order_update_operation` contains the following fields:
```cpp
       asset fee;
       account_id_type seller;
       limit_order_id_type order;
[...]

PR #215, opened 2019-09-17
BSIP 61: Operation to update limit orders

nathanhourt:


bsip-0061.md
```diff
@@ -0,0 +1,65 @@
+    BSIP: 0061
+    Title: Operation to Update Limit Orders
+    Authors:
+      Nathan Hourt <nat.hourt@gmail.com>
+    Contributors and Reviewers:
```
ryanRfox: Please remove the Contributors and Worker headings.

bsip-0061.md
```diff
@@ -0,0 +1,65 @@
...
+       optional<time_point_sec> new_expiration;
+
+       extensions_type extensions;
+```
+
+A single new evaluator is required, which shall check that the new field values are valid for the order being updated, and update the order and balances as necessary, then trigger order matching only if the price was modified such that the order has moved past the previous top-of-book order.
+
+No new indexes are required. No modifications to existing indexes are required. No new database objects are defined. No modifications to existing database objects are required.
+
+# Discussion and Summary for Shareholders
```
ryanRfox: Consider changing this to just "Summary" and Replacing "See Also" with "Discussion"

bsip-0061.md
```diff
@@ -0,0 +1,63 @@
...
+       account_id_type seller;
+       limit_order_id_type order;
+       optional<price> new_price;
+       optional<share_type> delta_amount_to_sell;
+       optional<time_point_sec> new_expiration;
+
+       extensions_type extensions;
+```
+
+A single new evaluator is required, which shall check that the new field values are valid for the order being updated, and update the order and balances as necessary, then trigger order matching only if the price was modified such that the order has moved past the previous top-of-book order.
```
pmconrad: Please specify validity checks.

nathanhourt: OK how's that?

pmconrad: Good, thanks.
Typo: ev~~w~~*a*luation
Typo: ~~m~~*n*ot

nathanhourt: D'oh! Thanks, fixed :)

abitmore:
We need business specialists to join the discussion.

pmconrad:
Since `limit_order_create_operation` fees do not depend on lifetime  either I think this should be discussed / changed separately.

> However, in this document, the logic about order creation fee refunding is not mentioned, thus readers would be misled (see BSIP 26).

In my understanding the fees of the update operation are collected immediately and the deferred fee of the limit order is not touched. Perhaps this should be mentioned in a "Discussion" section.

abitmore:
>> However, in this document, the logic about order creation fee refunding is not mentioned, thus readers would be misled (see BSIP 26).
>
> In my understanding the fees of the update operation are collected immediately and the deferred fee of the limit order is not touched. Perhaps this should be mentioned in a "Discussion" section.

In order to encourage users to use the order updating feature rather than `cancelling + recreating`, the order updating fee need to be even smaller than the order cancellation fee even when balance is changed. However, it means a user would be able to "roll" a partially filled order by paying a smaller order updating fee each time, rather than paying a bigger order creation fee, which would significantly impact network income. I think this should be addressed in this BSIP, but not only mention it in a *Discussion* section.

pmconrad:
> it means a user would be able to "roll" a partially filled order by paying a smaller order updating fee each time, rather than paying a bigger order creation fee, which would significantly impact network income

This is debatable (and has been, see #150 ), hence belongs in a Discussion section.
Sometimes lower prices mean more income due to higher volume. That's what we hope to achieve here.

> this should be addressed in this BSIP, but not only mention it in a Discussion section

One idea could be to have two fees - one for "fresh" limit orders and one (higher) for limit orders that have been partially filled. If the order has already been partially filled, the fee (or 90% of it) will be moved to the order's `deferred_fee`.
Downside is it's much more complicated for clients, and comes with the danger that transactions are invalidated when a "fresh" order is filled between tx creation and tx inclusion.

Other opinions what's better? @MichelSantos perhaps?

bsip-0061.md
```diff
@@ -75,3 +75,8 @@ This document is created for the betterment of humanity and is hereby placed int
...
+
+The limit order update operation may be used to "top up" an existing order and keep it on the books longer, possibly matching against more orders than would have been predicted when the order was created. This implies that for long-standing orders, the limit order update operation may constitute a discount over normal order creation, cancelation, and processing fees. The fee premium for modifying the order balance offsets this discount; however, a discount still exists for long-standing orders. Long-standing orders, however, are more valuable to the DEX than short-lived orders, as they promote market depth and high volume trading: market attributes which attract traders and more short-lived orders. The value added to the DEX as a whole by making it more economical for market makers to maintain long-standing orders significantly outweighs the discount the update operation affords to market makers; therefore, this proposal does not regard this discount on processing as detrimental to the blockchain as a whole.
```
abitmore: This is not necessarily true. The fee mechanism proposed in this BSIP encourages traders to move their partially-filled orders away from the top of the order book when the market is moving against their desired direction, and move them back when the market is moving in their desired direction, rather than cancelling it first and place a new order later. That said, it costs nodes more RAM to maintain the long-standing orders which won't fill nor help liquidity.

abitmore:
> lower prices mean more income due to higher volume.

The price of order cancellation is 10x lower than order creation according to current fee schedule. To gain the same fee we need 10x volume aka 10 times the operations added to the chain, which would bloat the chain much harder than what we have now, so I think it's against this BSIP's purpose.

abitmore:
>  For each fill at least one open order is consumed, which generates 10 times as many fees as an unfilled cancel+place alone.

OK, I'm almost convinced. Better get some data from the chain. How many orders are cancelled manually after partially filled (with no refund), how many orders are cancelled with a refund.

> If the market moves towards a sitting order, it will eventually be consumed, which is what the trader wanted in the first place.

Although it's wanted in the first place, it doesn't mean it's wanted forever. It's common practice for MM bots to cancel orders which would lead to a loss or lead to difficulties in order management.

By the way, I still think having 2 fees differentiating whether balance is updated is unnecessarily complex. KISS, one flat fee is the best.

MichelSantos:
# Short Version

I appreciate the cost factors raised by @abitmore, and the two-tiered fee structure raised by @pmconrad.  I think the proposed fee structure is good as an initial rough cut.

My longer reply organizes your ideas a little differently in an attempt to provide _the beginning of a more thorough reply_.  It comes down to asking "What expenses are we attempting to pay for?" and "Who do we want to pay for it, and when?".  The answers to those questions will then involve a proper business plan **with a fee schedule that is more realistic than the existing fee schedule** and which is probably outside the scope of this BSIP.


# Long Version

Are we truly attempting to have operation fees pay for a substantive portion of the blockchain expenses?

**If the answer is "no" then the fee schedule doesn't really matter.**  The current fees pay only a small fraction of the blockchain expenses.   And the limit order create and cancel operations are only a small fraction of that.  Therefore if the limit order update pays a little more or a little less than the current fees from cancel-create combinations, _we're just pretending to balance the books of the blockchain._

If the answer is "yes" then we need to discuss which expenses are we trying to cover?  Is it to cover the expense of witness pay?  Is it to cover up to the maximum daily amount of worker proposals?  Both?

Who do we want to pay for those expenses and when?

## Assumptions

I'll assume for the rest of this discussion that we are attempting to have limit order create, cancel, and update operations to pay _at least_ the expenses that are incurred by witnesses running a single block producer node.  I realize that witnesses run multiple block producer nodes but let's get some idea of the expenses related to limit orders on a single node.

I also realize that there are possible alternatives for getting other activities and other users to pay for the expense of limit orders but this is as good a place as any to start an analysis.


## Costs Related to Limit Orders

### Before Adding Limit Order Update Operation

|Type of Node|Lifecycle Phase of Limit Order|RAM|CPU|Disk Space|
|-|-|-|-|-|
|Any node|On the books|Order uses RAM|Adds some computation when orders are added and removed to indices, and during order matching|Limit order create and cancel operations use disk space. Limit order object uses storage for persisted object database.|
|Any node|After removal from books due to matching, cancellation, or expiration|None|None|Limit order create and cancel operations use disk space.|
|ES node|On the books and afterwards|?|?|Limit order operation and limit order cancel operation use disk space|

### After Adding Limit Order Update Operation

Assuming the limit order update operations are used

|Type of Node|Lifecycle Phase of Limit Order|RAM|CPU|Disk Space|
|-|-|-|-|-|
|Any node|On the books|Order uses RAM|Adds some computation when orders are added and removed to indices, and during order matching|Limit order create, cancel, and update operations use disk space. Limit order object uses storage for persisted object database.|
|Any node|After removal from books due to matching, cancellation, or expiration|None|None|Limit order create, cancel, and update operations use disk space.|
|ES node|On the books and afterwards|?|?|Limit order operation, limit order cancel operation, and limit order update operation use disk space.|

### Comparison

The introduction of a limit order update **could reduce** CPU usage _while the order is on the books_.

The introduction of a limit order update **could reduce** the overall disk space used _on any node_ if it is used instead of cancel and create operations.

The introduction of a limit order update **could reduce** the overall disk space used _ES nodes_ _at any time_.

### Estimates of Financial Expense

To estimate the cost of limit orders on block producers we'll need to estimate:

- Amount of RAM consumed by a limit order object in a regular node
- Amount of disk space consumed by a limit order object when persisted on a regular node
- Amount of disk space consumed by a limit order create, cancel, and update operation when persisted on a regular node
- Amount of RAM consumed by a limit order create, cancel, and update operation in an ES node
- Amount of disk space consumed by a limit order create, cancel, and update operation in an ES node

There may be other nuances or expenses that we need to consider.  And we might ignore the cost of ES nodes **if** we expect that the costs of ES nodes will be paid for by some other revenue stream.

We can then estimate financial expenses per limit order under various assumptions.


## Pricing / Fee Schedule to Cover Those Expenses

We'll need to estimate

- the number of witnesses
- the duration that an "average witness" will run a node
- the duration that an "average API" will run an ES node

We'll need to estimate whether this fee schedule will dissuade users from using the existing cancel-create combination versus the update operation.

And after we make these estimates we'll probably be wrong.  But at least we'll be able to make some justification for the fee schedule.

# Summary

It does not matter what the fee schedule is if we are not attempting to pay for specific expenses.

pmconrad:
> It's common practice for MM bots to cancel orders which would lead to a loss

Yes. MM's move their orders with the market. Simplifying things for market makers is the idea here.

> KISS, one flat fee is the best.

Agree.

> And after we make these estimates we'll probably be wrong. 

Lol, agree. Thanks for your insights @MichelSantos . This discussion is out of scope of this ticket, but interesting and necessary in a different context nonetheless.

bsip-0061.md
```diff
@@ -0,0 +1,79 @@
...
+ - At least one of the optional update fields is not null
+ - If `new_price` is not null, run its `validate` method
+ - If `delta_amount_to_sell` is not null, check it is not zero
+
+Additionally, the following stateful checks will be applied during evaluation:
+ - `seller` matches referenced order object's owner
+ - If `new_price` is not null, check its assets match those in the referenced order
+ - If `delta_amount_to_sell` is not null and positive, check it does not exceed seller's balance
+ - If `delta_amount_to_sell` is not null and negative, check it does not exceed order's balance
+ - Check updated order value is greater than zero
```
abitmore: Technically, we should do a `cull_small` check. Question: is it intended that this operation can be used as a cheaper alternative to cancel existing orders? What does "order value" mean?

nathanhourt: I utilized the same logic as the existing `maybe_cull_small_order` call and reject the transaction if it turns the order to dust.

nathanhourt: Revised the language to be clearer on this point

Issue #187, opened 2019-08-05
New BSIP: Add option to update an asset's precision


abitmore:
Currently precision of asset can't be updated, it's annoying when a user accidentally set an incorrect precision when creating the asset.

This BSIP propose a protocol change so that precision can be updated when there is zero supply.

To implement, can add a field in `extensions` of `asset_update_operation`.

xiao93:
support, thank abit.

abitmore:
This will be addressed in BSIP48 (https://github.com/bitshares/bsips/pull/216). Closing.

Issue #96, opened 2018-08-17
BSIP75: Asset Owner Defines MCR and MSSR Values


bangzi1001:
MCR (Maintenance Collateral Ratio) & MSSR (Maximum Short Squeeze Ratio) both normally are static and price feed publisher change it when requested by the asset owner. 

Looking that BitCNY MSSR case, why not let the asset owner or committee account have direct control of MCR & MSSR, it will be more efficient for asset owner to test what is the best number MCR & MSSR that are right for their assets. 

(Edited by @abitmore)
A draft is here: https://github.com/bitshares/bsips/pull/218.

bitcrab:
I agree, although I am not sure whether to do this kind change is possible/easy.

in my view, MCR & MSSR of committee created smartcoins are key parameters of the leverage system, change to these parameters need strong consensus of the community, it should not be as dynamic as feed price, in practice, witnesses tend to just keep publishing a fixed value.

xeroc:
The problem here is that market pegged assets can also be **privately owned** and the committee should not have any say about it. That said, the only option left is that the **issuer** is granted permission to change those parameters.

It's great to see a discussion about this.

Anyone has any idea why it is the responsibility of the feed producers to publish the MSSR?

One reason that I can think of is that it allows to have *many* opinions and the blockchain just takes the median.

Another reason might be that the witnesses can act *faster* than the committee can. However, there never was a discussion about how to deal with it and how much they should be allowed to tune the MSSR.

Good to see this finally addressed in discussion ..

bitcrab:
what happened has tell something:

abit has asked witnesses to positively participate in MCR&MSSR determination but get little response.

some witnesses from China just pubulished a different MSSR and then are rapidly voted out. 

all these tell that these parameters are sensitive and need strong consensus, it's not a good way to just let witnesses to publish and adopt the median, they are not very dynamic like feed price, so smartcoin issuer is the right account to manage these parameters.

grctest:
> Another reason might be that the witnesses can act faster than the committee can. However, there never was a discussion about how to deal with it and how much they should be allowed to tune the MSSR.

For private MPA an asset owner could react immediately, but since these values rarely change I don't believe delayed action via committee consensus would be an issue.

I think it's a good idea for these values to be set by the asset owner, also the docs should elaborate with scenarios where different values would be needed & the impact of such change on MPA operation.

Price feed script repos would need updated to not handle the MCR/MSSR values - some extra maintenance to notify private MPA owners of.

[...]

PR #216, opened 2019-09-17
Update PR 115 modify_max_supply

jmjatlanta:
This PR will adjust the language and add a few details to PR #115 to hopefully clarify the desires of the BSIP.

### DIFF
--- a/bsip-0048.md
+++ b/bsip-0048.md
@@ -11,44 +11,40 @@
 # Abstract
 
 The possibility of creating assets on the BitShares Blockchain is a core feature
-of our platform. In the case of an ico, the ico-holders can create an asset with
+of our platform. In the case of an ICO, the ICO holders can create an asset with
 a fixed maximum supply. However this maximum supply can be changed by the
-asset creator (issuer) anytime. This exposes participators ico to the risk,
-of loosing value of their bought asset.
+asset creator (issuer) anytime. This exposes participants in the ICO to the risk
+of losing value of their purchased asset.
 
 With this BSIP a flag will be introduced which forbids the modification of the maximum
 supply of an issuer created asset.
 
 # Motivation
 
-The implementation of a flag, to disable the modification of the maximum supply for 
-issuer created asset, has effects for both sides. On the one side the security for the
-participators or non-issuer users increases, because the value of their assets (tokens)
-can't be manipulated by exposing the market to more created tokens. And on the other side
-it is good marketing for the ico-hoders, because they can advertise with their fixed
-maximum supply.
+The implementation of a flag to disable the modification of the maximum supply for 
+issuer created asset, has benefits bor both the participants and the ICO holder. Participants int the ICO are assured the value of their assets (tokens) will not be diluted by issuers creating more tokens. ICO holders are able to market the fact that the maximum supply is fixed.
 
 # Rational
 
-The `disable_modify_max_supply` is a new created flag in the `asset_issuer_permissions`.
+The `allow_modify_max_supply` is a newly created flag in the `asset_issuer_permissions`.
 The flag can be activated to forbid the modification of the `max_supply` (`asset_object.options`).
-Once the flag is activated, it can't be deactivated.
+Once the flag is activated, it can only be deactivated if `current_supply` is zero.
 
 # Specifications
 
-This BSIP comes with a change to the protocol, which requires a propotocol upgrade.
+This BSIP comes with a change to the protocol, hence requires a propotocol upgrade.
 However the changes to the code are minor.
 
-1. Add the `disable_modify_max_supply` flag to the issuer permissions
-2. Add a `can_modify_max_supply()` the `asset_object`
-3. Change the `asset_update_evaluator::evaluate` method to assert false for
-    1. the deactivation of the `disable_modify_max_supply` flag
-    2. the modification of the `max_supply` when the `disable_modify_max_supply` is set
+1. Add the `allow_modify_max_supply` flag to the issuer permissions
+2. Set `allow_modify_max_supply` to true for existing assets at hardfork time
+3. Add a `can_modify_max_supply()` the `asset_object`
+4. Change the `asset_update_evaluator::evaluate` method to assert false for
+    1. the activation of the `allow_modify_max_supply` flag if `current_supply` is not zero or
+    2. the modification of the `max_supply` when the `enable_modify_max_supply` is false
   
 # Summary for Shareholders
 
-This BSIP proposes a way to improve the user security in the context of issuer created
-assets.
+This BSIP proposes a way to improve the financial security of users against the dilution of issuer created assets.
 
 # Copyright
 


bsip-0048.md
```diff
@@ -11,44 +11,40 @@
...
 # Motivation
 
-The implementation of a flag, to disable the modification of the maximum supply for 
-issuer created asset, has effects for both sides. On the one side the security for the
-participators or non-issuer users increases, because the value of their assets (tokens)
-can't be manipulated by exposing the market to more created tokens. And on the other side
-it is good marketing for the ico-hoders, because they can advertise with their fixed
-maximum supply.
+The implementation of a flag to disable the modification of the maximum supply for 
+issuer created asset, has benefits bor both the participants and the ICO holder. Participants int the ICO are assured the value of their assets (tokens) will not be diluted by issuers creating more tokens. ICO holders are able to market the fact that the maximum supply is fixed.
```
abitmore: Typo `bor` `int`

bsip-0048.md
```diff
@@ -11,44 +11,40 @@
...
 
 # Rational
 
-The `disable_modify_max_supply` is a new created flag in the `asset_issuer_permissions`.
+The `allow_modify_max_supply` is a newly created flag in the `asset_issuer_permissions`.
```
abitmore: We need to use the `disable_*` methodology here to be compatible with existing data and client applications. See https://github.com/bitshares/bitshares-core/pull/1375#pullrequestreview-164580457.

abitmore:
Also there is a discussion in https://github.com/bitshares/bitshares-core/pull/1375 about adding a `disable_issue` flag, perhaps also add it in this PR?

And maybe https://github.com/bitshares/bsips/issues/187?

bsip-0048.md
jmjatlanta: Thank you for the reference. It is unfortunate, but necessary. I will change it back.

bsip-0048.md
jmjatlanta: fixed

bsip-0048.md
```diff
@@ -10,45 +10,61 @@
 
 # Abstract
 
-The possibility of creating assets on the BitShares Blockchain is a core feature
-of our platform. In the case of an ico, the ico-holders can create an asset with
-a fixed maximum supply. However this maximum supply can be changed by the
-asset creator (issuer) anytime. This exposes participators ico to the risk,
-of loosing value of their bought asset.
+The possibility of creating assets on the BitShares Blockchain is a core feature of our platform. However some things are too maleable when they should not be. Other things are inflexible when they should be.
```
pmconrad: Typo: mal*l*eable

pmconrad: > Other things are inflexible when they should be.

This sounds wrong. Perhaps better "are not flexible when they should be".

bsip-0048.md
```diff
@@ -10,45 +10,61 @@
...
 
 # Motivation
 
-The implementation of a flag, to disable the modification of the maximum supply for 
-issuer created asset, has effects for both sides. On the one side the security for the
-participators or non-issuer users increases, because the value of their assets (tokens)
-can't be manipulated by exposing the market to more created tokens. And on the other side
-it is good marketing for the ico-hoders, because they can advertise with their fixed
-maximum supply.
+A flag to disable the modification of the maximum supply for issuer created asset will be implemented. This has benefits for both the participants and the ICO holder. Participants in the ICO are assured the value of their assets (tokens) will not be diluted by issuers creating more tokens. ICO holders are able to market the fact that the maximum supply is fixed.
```
pmconrad: Better "token holder" not "ICO holder". "ICO" is just one example where this might be important.

bsip-0048.md
```diff
@@ -10,45 +10,61 @@
...
+A flag to disable issuing allows token holders to be assured that the current supply will remain fixed, even if there are unissued tokens.
+
+The ability to change the precision of a created asset adds flexibility in the asset creation process.
 
 # Rational
 
-The `disable_modify_max_supply` is a new created flag in the `asset_issuer_permissions`.
-The flag can be activated to forbid the modification of the `max_supply` (`asset_object.options`).
-Once the flag is activated, it can't be deactivated.
+The `disable_modify_max_supply` is a newly created flag in the `asset_issuer_permissions`. The flag can be activated to forbid the modification of the `max_supply` (`asset_object.options`). Once the flag is set to `true`, it can only be set to `false` if `current_supply` is zero.
```
pmconrad: Please be more specific. "flags" are different from "permissions". "permission" implies the permission to change the flag value, whereas the "flag" implies a specific restriction.
Same for "disable_issue".

bsip-0048.md
```diff
@@ -10,45 +10,61 @@
...
+
+1. Add the `disable_issue` flag to the issuer permissions
+2. Set the `disable_issue` flag to `false` for existing assets at hardfork time
+3. Add a `can_issue()` to the `asset_object`
+4. Change the `asset_update_evaluator::do_evaluate` method to assert false for the setting of the `disable_issue` flag to `false` if it is currently `true`
+5. Change the `asset_issue_evaluator::do_evaluate` method to assert false if `can_modify_max_supply()` returns false
+
+To change the precision:
+
+1. Add the `fc::optional<uint8_t> new_precision` to the `asset_update_operation` struct.
```
pmconrad: No. It must be added to the extensions.

bsip-0048.md
```diff
@@ -10,45 +10,61 @@
...
+2. Change the `asset_update_evaluator::do_evaluate` method to assert false if `new_precision` exists and is not equal to `asset_object.precision` and `current_supply` is not zero.
```
pmconrad: Please add a "Discussion" section that explains why the new flags have negative meaning.

Please add a "Risks" section that highlights the dangers of changing precision.

bsip-0048.md
```diff
@@ -10,45 +10,61 @@
...
   
 # Summary for Shareholders
 
-This BSIP proposes a way to improve the user security in the context of issuer created
-assets.
+This BSIP proposes a way to improve the financial security of users against the dilution of issuer created assets. It also provides a mechanism for asset creators to modify the precision of their asset.
```
pmconrad: "user issued assets" instead of "issuer created assets"? Not sure what you mean there.

bsip-0048.md
```diff
@@ -10,45 +10,61 @@
...
+    1. the setting of the `disable_modify_max_supply` flag to `false` if `current_supply` is not zero or
+    2. the modification of the `max_supply` when the `disable_modify_max_supply` is `true`
+   
+For `disable_issue`:
+
+1. Add the `disable_issue` flag to the issuer permissions
+2. Set the `disable_issue` flag to `false` for existing assets at hardfork time
+3. Add a `can_issue()` to the `asset_object`
+4. Change the `asset_update_evaluator::do_evaluate` method to assert false for the setting of the `disable_issue` flag to `false` if it is currently `true`
+5. Change the `asset_issue_evaluator::do_evaluate` method to assert false if `can_modify_max_supply()` returns false
```
pmconrad: Does the flag have a meaning for bitAssets too?

bsip-0048.md
pmconrad: Oh, you mean "ICO holder" as the entity who make the offer. Then perhaps "asset issuer". "ICO" is still just an example.

abitmore:
Perhaps address #96 in this BSIP as well?

bsip-0048.md
jmjatlanta: Fixed

bsip-0048.md
jmjatlanta: I went through the document and attempted to standardize the language. Please let me know if it is now clearer.

bsip-0048.md
jmjatlanta: Clarified.

bsip-0048.md
jmjatlanta: Done.

abitmore:
> Do you feel the specifications of #96 were discussed enough

IMHO #96 is clear. Thanks.

bsip-0048.md
jmjatlanta: I would assume so. Does that carry additional implications?

ryanRfox:
I agree #96 is clear by now. However, I prefer to have it drafted as a distinct BSIP, rather than bundled with this one. @jmjatlanta are you willing to assist with drafting that BSIP, else I can find another resource to do so.

bsip-0048.md
```diff
@@ -10,45 +10,75 @@
...
 
-The implementation of a flag, to disable the modification of the maximum supply for 
-issuer created asset, has effects for both sides. On the one side the security for the
-participators or non-issuer users increases, because the value of their assets (tokens)
-can't be manipulated by exposing the market to more created tokens. And on the other side
-it is good marketing for the ico-hoders, because they can advertise with their fixed
-maximum supply.
+A flag to disable the modification of the maximum supply for issuer created asset will be implemented. This has benefits for both the participants and the asset creator. Asset holders are assured the value of their assets (tokens) will not be diluted by issuers creating more tokens. Asset creators are able to market the fact that the maximum supply is fixed.
+
+A flag to disable issuing allows asset holders to be assured that the current supply will remain fixed, even if there are unissued tokens.
```
pmconrad: Supply can still decrease, so it won't remain fixed.

bsip-0048.md
```diff
@@ -10,45 +10,75 @@
...
 
 # Motivation
 
-The implementation of a flag, to disable the modification of the maximum supply for 
-issuer created asset, has effects for both sides. On the one side the security for the
-participators or non-issuer users increases, because the value of their assets (tokens)
-can't be manipulated by exposing the market to more created tokens. And on the other side
-it is good marketing for the ico-hoders, because they can advertise with their fixed
-maximum supply.
+A flag to disable the modification of the maximum supply for issuer created asset will be implemented. This has benefits for both the participants and the asset creator. Asset holders are assured the value of their assets (tokens) will not be diluted by issuers creating more tokens. Asset creators are able to market the fact that the maximum supply is fixed.
```
pmconrad: ...for ~~issuer created~~*an* asset...

bsip-0048.md
pmconrad: bitAssets are not issued with `asset_issue_operation` but with `call_order_update`.

bsip-0048.md
```diff
@@ -10,45 +10,75 @@
...
+1. Add the `fc::optional<uint8_t> new_precision` to a new `asset_update_operation.extensions` struct.
+2. Change the `asset_update_evaluator::do_evaluate` method to assert false if 
+   1. `new_precision` exists and is not equal to `asset_object.precision` and `current_supply` is not zero
+   2. `new_precision` causes `max_supply` to be an invalid number
+
+# Risks
+
+It is believed that the addition of `disable_modify_max_supply` and `disable_issue` will add security for asset holders.
+
+The modification of the precision value carries no techincal risk, as asset creators will be unable to modify the precision once tokens have been issued.
```
pmconrad: I would leave that out. It also contradicts the next paragraph which I find more important.

bsip-0048.md
pmconrad: Good. Perhaps add a sentence about the ICO example use-case?

bsip-0048.md
jmjatlanta: Fixed

bsip-0048.md
jmjatlanta: Fixed.

bsip-0048.md
jmjatlanta: Added the check, and now calling the correct method (`can_issue` instead of `can_modify_max_supply`)

bsip-0048.md
jmjatlanta: Done.

bsip-0048.md
jmjatlanta: Let me see if I understand you correctly. From the programmer standpoint, disable_modify_max_supply is a flag, as it is a bit. But that muddies the waters, as others see it as a "permission". So I should re-word the paragraph to make `disable_modify_max_supply`  to be a "permission", as it is part of the "asset_issuer_permissions". 

If the above is correct, I will modify the paragraph below as well.

bsip-0048.md
jmjatlanta: @pmconrad  Please take a look at new line 27. Let me know if this is what you're looking for. Thanks for your input!

bsip-0048.md
pmconrad: What I mean is this: https://github.com/bitshares/bitshares-core/blob/master/libraries/protocol/include/graphene/protocol/asset_ops.hpp#L56-L59
Assets have *both* flags and permissions. The `flag` tells what is currently allowed. The `permission` tells if the flag setting can be changed.
An inactive `permission` can only be enabled when `supply == 0`.

jmjatlanta: The lightbulb just went on. Thanks for flipping the switch.

bsip-0048.md
```diff
@@ -10,45 +10,76 @@
...
+A flag to disable issuing allows asset holders to be assured that the current supply will not increase, even if there are unissued tokens.
+
+The ability to change the precision of a created asset adds flexibility in the asset creation process.
 
 # Rational
 
-The `disable_modify_max_supply` is a new created flag in the `asset_issuer_permissions`.
-The flag can be activated to forbid the modification of the `max_supply` (`asset_object.options`).
-Once the flag is activated, it can't be deactivated.
+`disable_modify_max_supply` will be added to `asset_issuer_permissions`. The permission can be activated to forbid the modification of the `max_supply` (`asset_object.options`). The related flag can only be set to `false` if `current_supply` is zero.
```
pmconrad: I think you have mixed up the meaning of permissions and flags.

This negative logic puts knots in my brain just from reading the BSIP. It will also be a PITA to implement. I think we should move these flags to positive logic. How many clients for creating assets are out there that might break?

abitmore: > How many clients for creating assets are out there that might break?

All clients.

pmconrad: I think requiring clients to upgrade is better than being forever being stuck with this twisted logic.

jmjatlanta: I think I clarified my permissions/flags statement. Please check. I have also switched to using positive logic. See https://github.com/bitshares/bsips/pull/216/commits/858b7ee5dc39a6b2acdb6d0ccfd8c5fd6083a0d3

bsip-0048.md
sschiessl-bcp: Is this considered resolved?

bsip-0048.md
sschiessl-bcp: > I think it's better to revert to negative logic.

I agree here. If that twists user's minds we can present it positively in UI

bsip-0048.md
abitmore: It's partially resolved. In the last commit one sentence of two got updated.

Issue #161, opened 2019-04-23
BSIP77: Require Higher CR When Creating/Adjusting Debt Positions


abitmore:
BSIP: 0077
    Title: Require Higher CR When Creating/Adjusting Debt Positions
    Authors: Abit More
             @shulthz
    Status: Draft
    Type: Protocol
    Created: 2019-09-30

# Abstract

This BSIP proposes a protocol change which aims to improve overall collaretal ratio (CR) of smartcoins.

# Motivation

Risk-loving traders tend to borrow as much debt as possible with as little collateral as possible,
which impacts overall CR of that smartcoin. Smartcoin owners would like to have a tool to counter
this behavior.

# Rationale

Requiring a higher CR for new debt positions would introduce a buffer thus would increase the
stability of the smartcoin.

# Specification

Add one parameter "Initial Collateral Ratio" (ICR) to each smartcoin asset, which can be updated by the asset owner.

When adjusting a position, apply [BSIP30](https://github.com/bitshares/bsips/blob/master/bsip-0030.md)
if current CR is lower than ICR but not MCR.

Before the protocol upgrade:
```
when creating a new debt position
  require CR > MCR

when adjusting a debt position
  if new CR <= MCR
    require CR increased and debt not increased

when current CR <= MCR
  trigger margin call
```

After the protocol upgrade:
```
Require ICR >= MCR

when creating a new debt position
  require CR > ICR

when adjusting a debt position
  if new CR <= ICR
    require CR increased and debt not increased

when current CR <= MCR
  trigger margin call
```

Note: logic around *target collateral ratio* (TCR) is not affected.

# Discussion

This BSIP introduces a new tool for smartcoin asset owners to fine tune their assets.
Having an ICR greater than MCR would probably result in less risk-loving traders borrowing that
smartcoin, thus may lead to fewer market activities and worse liquidity. It's up to the asset owners
to decide whether to use this tool.

# References
* https://github.com/bitshares/bsips/issues/129#issuecomment-483937118
* https://github.com/bitshares/bsips/issues/96#issuecomment-484126930
* https://bitsharestalk.org/index.php?topic=29638.0

# Copyright
This document is placed in the public domain.

clockworkgr:
Would this make TCR be X at minimum too?

abitmore:
I don't think we need to limit TCR.

pmconrad:
[...]

PR #217, opened 2019-09-19
Update README for BSIPs 47, 59, 63, 75, 76, 77

ryanRfox:
Updates the README for many new/updated BSIPs

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Number             | Title                                                    | 
 [60](bsip-0060.md) | BitShares URI scheme | John Titor, Stefan Schießl, Abit More | Informational (Client Protocol) | Accepted
 [61](https://github.com/bitshares/bsips/issues/150) | Operation to Update Limit Orders | Nathan Hourt | Protocol | Draft
 [62](https://github.com/bitshares/bsips/issues/156) | Close Short Position | Stefan Schießl | Protocol | Draft
-[63](https://github.com/bitshares/bsips/issues/172) | Short-lived Unidirectional Payment Channels | Christopher J. Sanborn | Informational | Draft
+[63](bsip-0063.md) | Short-lived Unidirectional Payment Channels | Christopher J. Sanborn | Informational | Draft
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,3 +81,4 @@ Number             | Title                                                    |
 [73](bsip-0073.md) | Match Force-Settlement Orders with Margin Calls and Limit Orders | Abit More | Protocol | Draft
 [74](bsip-0074.md) | Margin Call Fee Ratio | Jerry Liu | Protocol | Draft
 [75](https://github.com/bitshares/bsips/issues/96) | Asset Owner Defines MCR and MSSR Values | John Jones | Protocol | Draft
+[76](https://github.com/bitshares/bsips/issues/221) | ommittee-Defined SmartAsset Collateral Threshold | Abit More | Protocol | Draft
```
abitmore: Typo

abitmore: I'd like to create two BSIPs for the topic. The first one is informational for witnesses to execute via price feeds before the second one is implemented which is a protocol change. So IMHO the title and type listed here doesn't actually reflect the approach.

ryanRfox: Sorry, I neglected to update 76 to Informational when I cut/paste. I'll do that now for 76, then reserve 77 for the Protocol. Reply back here with the titles for both 76 and 77 and I'll update my README doc accordingly. I'll need an Issue # for 77 to point to (assuming you will keep them as distinct discussions).

abitmore: Actually the protocol one only makes sense if the informational one is supported by the majority. I see the potential, but no guarantee. So IMHO don't expect it in the near future, nor need to reserve a number for it so far.

ryanRfox: Good point about the process. I corrected the typo. Still not clear what you prefer the title to be for 76. Separately, assigned 77 for #161

abitmore: Thanks.

abitmore:
Please resolve conflicts.

ryanRfox:
I think that may be correct.

Issue #221, opened 2019-09-25
BSIP76: Threshold for price feeds through voting


abitmore:
BSIP: 0076
    Title: Threshold for price feeds through voting
    Authors: Abit More
    Status: Draft
    Type: Informational
    Created: 2019-09-25

# Abstract

This BSIP defines a process to influence the pegging status of committee-owned smartcoins through poll voting.

# Motivation

In order to maintain the peg of the smartcoins, smartcoins holders now have the right to be always
possible to buy collateral at a "fair" price. For committee-owned smartcoins, the collateral is BTS.
However, when the BTS token is being dumped in Centralized Exchanges with fake volume and even fake supply,
the value of the token is distorted, thus, forcing BTS token holders who have their BTS token trapped in
debt positions as collateral to sell their collateral at the distorted price via margin calls or
force-settlements is unfair.
The same situation applies to non-committee-owned smart coins as well.

Smartcoins are products of the BitShares platform, their rules are defined by the platform
for the benefit of the platform and the BTS token holders.
If a product brings more harm than benefit to the platform, its rules need to be optimized,
under extreme conditions, the product needs to be halted.
For smartcoins, when the cost to maintain the peg is too high to afford, it's time to give up the peg.

# Rationale

Ideally a smartcoin owner should be able to define a set of conditions about when to give up the peg
and how to de-peg, but it needs time to design and implement. Before the final solution got worked out,
it's up to the price feed producers to execute the will for the asset owner on whether to maintain
the peg at a given point and how to de-peg. For committee-owned smartcoins, the owner is actually
all BTS holders who could express their will by voting, the price feed producers are the active witnesses.

# Specification

The BTS token holders define a threshold, if the market trading price of BTS is below the threshold,
witnesses should feed the threshold but not the market trading price, thus would effectively break
the peg as intended. When the market trading price of BTS is above the threshold, witnesses should
feed the market trading price, so the peg will be restored automatically. The mechanism is similar to
the Global-Settlement Protection mechanism as described in [BSIP 58](bsip-0058.md).

Changes to the threshold will be decided on individual committee-owned smartcoins by voting, each time
one change is issued 2 poll worker proposals will be created standing FOR and AGAINST the change.

For example, to decide whether to change the price feed threshold of bitCNY to 0.2 CNY each BTS,
2 poll worker proposals will be created for voting:

Poll-BSIP**-Change bitCNY feed threshold [from the previous value when applicable] to 0.2 CNY per BTS

[...]

Issue #83, opened 2018-07-08
New BSIP: stake lock-up mechanism, count only real "locked" stake as voting stake


abitmore:
IMHO, for a healthy ecosystem, decision making on important things should be serious and relatively stable, so governance voting should only be allowed for long-term self-owned stake only, and interests should be aligned.
* stakes in balances can be liquidized / transferred / sold at any time, allow voting with balance means instability, may negatively impact decision making, because everyone can obtain (buy) some stake and participate in and influence decision making (voting) at any time immediately, then sold them at any time immediately when no longer want to participate, with no consequence;
* utilized stakes should be considered as "used" but not "being owned", thus stakes in collateral or market orders should not be counted in;
  * for stakes in collateral, it means debts exist and the positions can be liquidized in various ways, thus the stakes are not 100% owned;
  * for stakes in market orders, it means they're for sale, which showed unwillingness to hold;

To reduce influence caused by random stake movements, it's a good strategy to adopt some kind of stake lock-up mechanism.
* long-term stake holders might be willing to lock up their stakes for a period;
* people who have influenced decision making (voting) should take responsibility, especially if negative consequences have been made, they should not be able to sell their stake quickly enough when the price started to drop;
* one controversial topic is, should people be allowed to suddenly start influencing decision making with a big stake, especially if said decisions will invalidate earlier efforts made by other stake holders. IMHO it's better to minimize influence of this behavior, e.g. requires a holding period for voting.

The vesting mechanism used in Steem, namely VESTS, has some behaviors described above. However, VESTS in Steem is being heavily utilized, thus IMHO is not a good base for governance voting.

My proposal:
* stake holders can lock up a part of their stakes, to gain corresponding voting power
  * only locked-up stakes can be used to vote, don't count stakes in balances/orders/collateral
* voting power accumulates by time, generally, stakes locked longer have more voting power
  * cap voting power in some way to avoid endless accumulating
* stake holders can start unlocking their locked-up stakes partially or fully at any time
  * unlocked stake will go to balance after a relatively long period, say, 3 months;
  * immediately clear accumulated voting power when started unlocking

More things to think about:
* If really want to cash-out, people can sell their accounts with locked-up stakes as a whole.
* Legal issues?

Thoughts?

abitmore:
Related to #81.

TheTaconator:
[...]

PR #224, opened 2019-09-30
Incorporate core changes necessary for BSIP-0076 to allow a threshhold price

sschiessl-bcp:


abitmore:
Updated wording. Now looks good to me.

ryanRfox:
Yes. Looks good to me. Thank you @sschiessl-bcp for updating this BSIP. I've requested @pmconrad or @MichelSantos to provide the final review as they contributed on the previously merged BSIP71.

bitcrab:
not bad

pmconrad:
I will not support such a protocol change, not even with a review.

sschiessl-bcp:
The realization of this BSIP allows debt holders to massively increase voting power, artifically in my opinion. This might hinder progress in the future, or give incentive to drive progress into one single direction (to accomodate margin position holders).

Thinking more about this I think it is a necessary, simple and justifiable addition to this BSIP is that the voting power of a margin position is reduced, and only the voting power of the excess BTS count towards the margin position holders. Then we have one question to discuss: Should the created BitAsset give voting power equal to the equivalent amount of BTS, or should the voting power that is subtracted from margin positions simply be lost?

Example: Margin position with CR 1.6. The voting power of this margin position is reduced by a BTS amount equivalent of CR 1. 

Thoughts?

abitmore:
IMHO the logic around voting (or weight) is in the scope of [BSIP24](https://github.com/bitshares/bsips/blob/master/bsip-0024.md) and #83, but not this BSIP.

thul3:
This BSIP allows EVERYONE to increase their voting power not just debt holders

abitmore:
It's essentially a negotiation between smartcoin holders and smartcoin creators (aka borrowers or debt position holders).

Let's score the options on the desk. Assuming the option that gives the smartcoin holders the most benefits has a score of `1`, the option that gives the smartcoin creators the most benefits has a score of `5`.

Here are my scores for the options:
* the old GS rule, score `3`.
* the already-merged draft of BSIP71 favors smartcoin holders over creators, but still shows some merit for the creators e.g. the "fill the least collateralized call order first" rule, score `2`;
* in the discussion when drafting BSIP71 (#193), some people came up with ideas that favors smartcoin holder more (IMHO), e.g. "always settle at feed price" and "filling as much margin-called debt as possible", score `1`;
* the "price threshold" rule discussed in this issue and BSIP76 favors smartcoin creators over smartcoin holders under certain circumstances, score `4`.

People are taking side.

sschiessl-bcp:
> This BSIP allows EVERYONE to increase their voting power not just existing debt holders

That is correct, still that assumes that all BTS holders are risk lovers.

sschiessl-bcp:
> IMHO the logic around voting (or weight) is in the scope of [BSIP24](https://github.com/bitshares/bsips/blob/master/bsip-0024.md) and #83, but not this BSIP.

Also correct. This BSIP allows debt holders to margin to the max  risk-free, while significantly and overwhelmingly increasing their voting power. That is why I see it relevant in here.

ryanRfox:
What do we want to do with BSIP71 at this time? 
- Vote on the merged BSIP71 draft
- Continue discussions on this Change Request

liondani:
> The realization of this BSIP allows debt holders to massively increase voting power, artifically in my opinion. This might hinder progress in the future, or give incentive to drive progress into one single direction (to accomodate margin position holders).
> 
> Thinking more about this I think it is a necessary, simple and justifiable addition to this BSIP is that the voting power of a margin position is reduced, and only the voting power of the excess BTS count towards the margin position holders. Then we have one question to discuss: Should we do anything with the "lost voting power", or simply have it lost?
> 
> Example: Margin position with CR 1.6. The voting power of this margin position is reduced by a BTS amount equivalent of CR 1.
> 
> Thoughts?

What if we give the option to bitASSETs holders to give the "lost" voting power (or even generate more voting power!) back to margin position holders by activating a yearly interest rate of 1% (or more) to bitASSETs owners that get's paid straight from the margin position holders pockets + partially from a worker if absolutely necessary?

bsip-0071.md
```diff
@@ -57,22 +60,24 @@ The problem of Partial GS is that it does not resolve itself automatically when
...
 
+
+
 # Specifications
 
-Add one flag "Prevent Global Settlement" to each smartcoin asset.
+Add one flag "Prevent Global Settlement and External Attacks" to each smartcoin asset.
 
-Introduce new parameters: *settlement price* and *call execution price*.  They shall be calculated as:
+Introduce new parameters: *threshold price*, *settlement price* and *call execution price*. The remaining two shall be calculated as:
```
MichelSantos: The specifications make no mention of what is stated in the abstract: "this BSIP allows the asset owner to define an optional threshold price".  A specification should describe _who_ will set and unset the threshold price, and _when_.

MichelSantos:
@sschiessl-bcp Please also change "%div;" to "&div;"

bsip-0071.md
sschiessl-bcp: That sentence somehow got lost

sschiessl-bcp:
i adressed all comments.

 - now two flags, one for threshold one for prevent GS
 - added voting reduction

bsip-0071.md
```diff
@@ -51,59 +75,126 @@ In my view, a good way to handle bad debt need to follow below principles:
...
+yet all the smartcoin trading features, including borrowing, margin call and force settlement, 
+will all switch to refer to GS price to ensure the continuity and fairness of all the features.
 
 # Specifications
 
-Add one flag "Prevent Global Settlement" to each smartcoin asset.
+#### Add flags "Prevent Global Settlement" and "Set Threshhold price" to each smartcoin asset.
+
+Those flags are set by the asset owner, and come with their respective permission like all others. 
+Default values
```
MichelSantos: The "Default values" thought is incomplete

bsip-0071.md
```diff
@@ -122,6 +126,10 @@ FP<sub>M</sub> is still used for determining if a debt position is margin called
 
 **Note:** when comparing prices, `max` is to be understood in terms of debt divided by collateral.
 
+#### Adjustment to voting power
+
+The voting power for BTS backed SmartCoins added to the owner of a margin position is reduced. Only the exccess collateral counts as voting power. Example: If a position has CR 1.8, the voting power from this position only counts the BTS that represent the excess of 0.8. Reasoning for this is given in the section Discussion / Voting power.
+
```
abitmore: Changes about voting power should not be in the scope of this BSIP.

sschiessl-bcp: This BSIP cements the ability for risk-free margin to the max, multiplying voting power, and setting a status quo and "confirmed by core". In the worst case scenario, this is an irreversible action.

Planning for the worst-case scenario, I see it very much in relevance to this BSIP, and would rather not want to remove it. This is certainly an effect that we already see now, still having it inside the core is of different magnitude IMO.

bsip-0071.md
sschiessl-bcp: Done

abitmore:
I think it's better to create a new BSIP for this feature rather than adding it to BSIP71. Some voters may want the GSP feature but not this one, some others may want this but not the GSP feature.

sschiessl-bcp:
> I think it's better to create a new BSIP for this feature rather than adding it to BSIP71. Some voters may want the GSP feature but not this one, some others may want this but not the GSP feature.

In a perfect world, yes. And that is the point. Surely you can say, this is what the voters want. I say the voting power distribution is already heavily skewed (was already in the past, simply not so much apparent to me), and I am trying to present reasonable compromises. If given the choice, I don't believe that voting power correction for margin positions will ever make it.

abitmore:
>  If given the choice, I don't believe that voting power correction for margin positions will ever make it.

I believe it won't pass the voting, so why we waste time/efforts/funds including it in the BSIP? Having it in the BSIP means other features in the BSIP won't pass as well.

sschiessl-bcp:
I'd like to hear some more opinions.

Option a)
Put it up in one BSIP

Option b)
Put threshold as one, and margin position correction as another. Both would need to be available for voting at the same time.

clockworkgr:
> > If given the choice, I don't believe that voting power correction for margin positions will ever make it.
> 
> I believe it won't pass the voting, so why we waste time/efforts/funds including it in the BSIP? Having it in the BSIP means other features in the BSIP won't pass as well.

That's the whole point. Including it in the BSIP forces the affected parties to compromise.

abitmore:
> I'd like to hear some more opinions.
>
>Option a)
>Put it up in one BSIP
>
>Option b)
>Put threshold as one, and margin position correction as another. Both would need to be available for voting at the same time.

You're misleading. The main point is not whether to put them into one BSIP, but whether to have the voting power change in the BSIP.

> Including it in the BSIP forces the affected parties to compromise.

Why to force?

clockworkgr:
> > I'd like to hear some more opinions.
> > Option a)
> > Put it up in one BSIP
> > Option b)
> > Put threshold as one, and margin position correction as another. Both would need to be available for voting at the same time.
> 
> You're misleading. The main point is not whether to put them into one BSIP, but whether to have the voting power change in the BSIP.
> 
> > Including it in the BSIP forces the affected parties to compromise.
> 
> Why to force?

because as you said...it would never get voted on its own.

When its in one BSIP it's give and take

bsip-0071.md
bitcrab: I totally disagree with the adjustment to voting power, voting power is inside BTS, one should own the voting power of the BTS in his account, either his account is debt free or not, debt just means he would like to bear the relevant risk, no others.

bitProfessor:
> 应该增加三个风险
> 
> 首先，摘要指出：“该BSIP允许资产所有者定义可选的阈值价格，以防止外部市场操纵。” 明显的风险是资产所有者可能实际上会使用任何理由来设置阈值。
> 
> 其次，如果设定了新的阈值价格，资产持有人_在事后_可能会改变持有人通过强制结算以市场价格赎回资产的基本能力。
> 
> 第三，这有效地将资产发行人转变为可以有效暂停赎回要求的货币发行局。如果发行人也在借入资产，则可能存在利益冲突。

Yes

bsip-0071.md
clockworkgr: What risk?

tonyniu352:
it is a important reminder for everyone on the core team that you are not authorized to infringe the voting rights of the bts holder. voting power is also property arrribute attached to bts coin, that is,you are not allowed to infringe the property of bts holders. if the voting right adjustment is approved, the call for a vote of no-confidence in the core team is possible.

clockworkgr:
"which shall has no negative impact on the original business logic or function"

So no changing the original business logic?

Stop making silly arguments only when it suits you

bitProfessor:
After price recovery, I no longer support gs, bsip76.  without bitcny, I couldn't find any reason to support it.

pmconrad:
> The development team is not authorized to modify voting power

We developers know that. We do not change consensus without following proper procedures, i. e. creating a BSIP and have it approved by shareholders. Unlike others.

> or discuss how to change such power of bts holder.

We are *also* shareholders, and as such we are *of course* authorized to *discuss* changes.

> you are not allowed to infringe the property of bts holders

You guys had no problems with infringing the property of bitCNY/bitUSD holders when you approved BSIP-76, and now you are complaining that others are discussing the same?

sschiessl-bcp:
Related or not related to this topic, this was posted in Telegram just now.

> cn-vote, [09.10.19 12:39]
> I want to remind the current core development team. Cn-vote has always supported the core development team, please do not live up to everyone's expectations. The core development team should not forget what they are and what they should do. Do your own job, because there are already many excellent and cheaper teams who have repeatedly sought to compete for the core team.
> 
> cn-vote, [09.10.19 12:44]
> BM left, and the bitshares have been running steadily for so many years. Therefore, no one is indispensable. Don't forget who you are and what you should do. Do your job well and everyone will always support you. Otherwise, there will be more suitable teams and people. Thank you all.
> A job like this high salary and not so busy can be difficult to find. Many excellent teams have been coveted for a long time.

bsip-0071.md
```diff
@@ -9,23 +10,43 @@
...
 FP &le; P<sub>gs</sub> &times; MSSR
 
 If this does occur it means that the collateral ratio (CR) of at least one debt position is less than the MSSR.
 
-Global settlement is not a good way to handle bad debt, as can be seen to what happened to bitUSD.  After global settlement was triggered for bitUSD in December 2018:
+Global settlement is not a good way to handle bad debt, as can be seen to what happened to bitUSD.  
+After global settlement was triggered for bitUSD in December 2018:
 
 - traders could no longer borrow bitUSD;
 - the price of bitUSD dropped below USD because of insufficient collateral; and,
```
grctest: The price of bitUSD has already dropped below USD with BSIP 76, [almost 20% by my calculations](https://github.com/bitshares/bsips/issues/221#issuecomment-541351091). What was the max price impact from bitUSD's GS by comparison?

bsip-0071.md
```diff
@@ -9,23 +10,43 @@
...
 - it will take a long time for bitUSD to be revived.
```
grctest: bitUSD was in a GS state from Nov to Feb? BSIP76 has been active for almost a month now and will persist until the price recovers (same conditions for GS recovery - may be a long time).

bsip-0071.md
```diff
@@ -9,23 +10,43 @@
...
 
-The community has had extensive discussions about how to handle the bad debt in a better way in the future.  However an easy way, [BSIP58](https://github.com/bitshares/bsips/blob/master/bsip-0058.md), has been implemented for smartcoins like bitCNY and bitUSD; it has worked successfully to prevent GS from happening with no other obvious impact.  However, BSIP58 has some issues - it is suspected of market manipulation and it risks witnesses independence.  Moving forward, this BSIP will be built based on BSIP58 and eliminate all its disadvantages.
+The community has had extensive discussions about how to handle the bad debt in a better way in the future.
```
grctest: That doesn't sound like a bad consequence of global settlement.

bsip-0071.md
```diff
@@ -9,23 +10,43 @@
...
+The community has had extensive discussions about how to handle the bad debt in a better way in the future.  
+However an easy way, [BSIP58](https://github.com/bitshares/bsips/blob/master/bsip-0058.md), 
+has been implemented for smartcoins like bitCNY and bitUSD; it has worked successfully to prevent GS 
+from happening with no other obvious impact.  
+However, BSIP58 has some issues - it is suspected of market manipulation and it risks witnesses independence.  
+Moving forward, this BSIP will be built based on BSIP58 and eliminate all its disadvantages.
+
+[Shorting attacks and fake volume](https://bitsharestalk.org/index.php?topic=29635.0) have presumably caused massive
```
grctest: Has there been investigations into this? Have the CEX in question commented on the incident? Why couldn't a change to price feed scripts to temporarily filter out sources of suspicious price changes from singular CEX have been the solution instead of fixing the price feeds to an unpegged value?

bsip-0071.md
```diff
@@ -39,9 +60,12 @@ We now have several choices on handling bad debt:
 
 We need to do a deep review on bad debt before evaluating above solutions.
```
grctest: When can we expect this deep review on bad debt? Why can't we work on improving the price feed scripts in parallel?

bsip-0071.md
```diff
@@ -39,9 +60,12 @@ We now have several choices on handling bad debt:
...
 
-Bad debt means in some debt positions the debt cannot be fully paid by selling the collaterals via margin call/force settlement at the market price.  In other words, the relevant smartcoin loses sufficient collateral to back the value and will possibly devalue.
+Bad debt means in some debt positions the debt cannot be fully paid by selling the collaterals via margin 
+call/force settlement at the market price.  In other words, the relevant smartcoin loses sufficient collateral 
+to back the value and will possibly devalue.
 
-At this moment, the system should accept the fact that the smartcoin will possibly devalue, find ways to minimize the impact to different parties in the market and the time to revive.
+At this moment, the system should accept the fact that the smartcoin will possibly devalue,
```
grctest: It's not just a possibility, but now a fact that the CNY & USD assets have been devalued by BSIP 76.

bsip-0071.md
```diff
@@ -51,59 +75,126 @@ In my view, a good way to handle bad debt need to follow below principles:
...
 
-Introduce new parameters: *settlement price* and *call execution price*.  They shall be calculated as:
+While the *threshold price* is a new BitAsset option and defined by the asset owner, 
+the remaining two shall be calculated as:
 
 ```
-if flag "Prevent Global Settlement" is enabled
-	settlement price =  max(FP_M * (1 - settlement_offset), P_gs)
-        call execution price = max(FP_M * (1 + MSSR), P_gs)
+if flag "Prevent Global Settlement and External Attacks" is enabled
```
grctest: > Add flags "Prevent Global Settlement"
> flag "Prevent Global Settlement and External Attacks"

Which is the final name for the flag?

bsip-0071.md
grctest: @bitcrab If you don't own the associated bitasset for your debt, then you effectively sold that 1x CR BTS which the bitasset holder can claim at any moment. Anything above 1x CR would still be yours to vote with, so just maintain a good CR 👍

bsip-0071.md
```diff
@@ -51,59 +75,126 @@ In my view, a good way to handle bad debt need to follow below principles:
...
+introducing a voting procedures where holders of the SmartCoin can participate, with their voting weight reflecting their
+holdings of the SmartCoin in question. This procedure needs precise definition in a new BSIP.
+
+## Voting power
+
+As of right now, a SmartCoin that has BTS as its collateral provides the full amount of collateral as voting power. 
+This is not considered to be fair since it is borrowed voting power and gives incentives to squeeze out as much as 
+possible from a margin position to increase voting weight (borrow, sell bitasset on market, borrow more, and so on). 
+This is adressed by adjusting voting power of a position, and the suggested solution ensures that the voting power of 
+a margin position holders remains unchanged, with the assumption that he sells the newly borrowed bitasset on the
```
grctest: Wouldn't the voting power of margin position holders decrease after implementation since the 1x CR vote weight would be null, only > 1x CR would count as vote weight.

bsip-0071.md
```diff
@@ -51,59 +75,126 @@ In my view, a good way to handle bad debt need to follow below principles:
...
+ Global Settlement would have occured, and are forced to use the open market if they want to avoid holding a 
+ SmartCoin that is eventually undercollaterized
+ 
+ - If the Threshhold price is enabled, the promise of being able to settle for equivalent value of the collateral SmartCoin holders are extradited to the asset's owner ability to set it, with the direct implication on the force settlement price. The Threshhold price is defined to prevent external market manipulation, but ultimately the asset owner may use any justification for setting the threshold.
 
-Variant B would have let bad debt positions matched at their own collateral ratio while matching other margin calls as before at FP<sub>M</sub> &times; MSSR, hoping to improve the overall collateralization. This also means that debt positions with better collateral ratio would have been preferred over bad debt, meaning that bad debt positions will remain sitting on the market for a longer time.
+ - If the Threshhold is set, conflict of interest exists if the asset owner is also borrowing the asset into existence
```
grctest: The asset owner could make an anonymous secondary account so it's safe to assume if the threshold is set there exists a conflict of interest.

bsip-0071.md
```diff
@@ -51,59 +75,126 @@ In my view, a good way to handle bad debt need to follow below principles:
...
+ 
+ - If the Threshhold price is set, margin positions creators essentially have a call price limit that they can set in order to avoid ever being margin called. This means the position carries no risk for the margin position holder anymore, which can be seen as incentive to borrow as much as possible without considering the consequences. This is the reason to adjust the voting weight of a BitAsset position
 
 # Summary for Shareholders
 
-It is important to eliminate concerns about global settlement.  This is currently achieved with the technical options that are available to the witnesses.  A more advanced solution as proposed here is needed.
+It is important to eliminate concerns about global settlement and market manipulation.  This is currently achieved with 
+the technical options that are available to the price feeders (i.e. the witnesses for BitASsets like bitCNY).  
+A more advanced solution as proposed here is needed that allows the price feeders to feed the real external market price.
```
grctest: Price feed publishers can already feed the real external market price, they were pre-bsip76

bsip-0071.md
```diff
@@ -51,59 +75,126 @@ In my view, a good way to handle bad debt need to follow below principles:
...
 
-Variant B would have let bad debt positions matched at their own collateral ratio while matching other margin calls as before at FP<sub>M</sub> &times; MSSR, hoping to improve the overall collateralization. This also means that debt positions with better collateral ratio would have been preferred over bad debt, meaning that bad debt positions will remain sitting on the market for a longer time.
+ - If the Threshhold is set, conflict of interest exists if the asset owner is also borrowing the asset into existence
+ 
+ - If the Threshhold price is set, margin positions creators essentially have a call price limit that they can set in order to avoid ever being margin called. This means the position carries no risk for the margin position holder anymore, which can be seen as incentive to borrow as much as possible without considering the consequences. This is the reason to adjust the voting weight of a BitAsset position
 
 # Summary for Shareholders
```
grctest: impact on smartcoin, debt & bts holders?

bsip-0071.md
sschiessl-bcp: You know why they did not. Please be constructive.

bsip-0071.md
sschiessl-bcp: Yes

bsip-0071.md
sschiessl-bcp: refined the wording

bsip-0071.md
sschiessl-bcp: Differentiate market price and feed price. The promise of force settle for equal value is broken.

bsip-0071.md
sschiessl-bcp: Revival would be in effect only after reaching MCR of the equivalent necessary value.

bsip-0071.md
sschiessl-bcp: Relevance?

bsip-0071.md
sschiessl-bcp: Not by me. No comments, have been asked. Cross-CEX arbitrage would render that useless, which is especially in a short-term scenario still in effect.

bsip-0071.md
sschiessl-bcp: Don't know. Certainly we can, care to join?

sschiessl-bcp:
Fixed the calculation wording when flags are set, so at least the idea is correctly presented.

PR #226, opened 2019-09-30
BSIP77: Require Higher CR When Creating/Adjusting Debt Positions

abitmore:
PR for #161.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Number             | Title                                                    | 
 [74](bsip-0074.md) | Margin Call Fee Ratio | Jerry Liu | Protocol | Draft
 [75](https://github.com/bitshares/bsips/issues/96) | Asset Owner Defines MCR and MSSR Values | John Jones | Protocol | Draft
 [76](https://github.com/bitshares/bsips/issues/221) | Committee-Defined SmartAsset Collateral Threshold | Abit More | Informational | Draft
-[77](https://github.com/bitshares/bsips/issues/161) | Require Higher CR When Creating/Adjusting Debt Positions | Abit More | Protocol | Draft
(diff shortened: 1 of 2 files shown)


bitcrab:
if no more change required, possible to merge this ASAP?

sschiessl-bcp:
I am wondering if this might need to be reconsidered?

`Note: logic around *target collateral ratio* (TCR) is not affected.`

Current BSIP only affects initial positions. It is to be expected that positions live longer such that replenishing CR through margin call would happen more often as compared to creating a new position. Essentially, my suggestion is to also enforce ICR as minimum value for TCR (no new parameter).

It was discussed [here](https://github.com/bitshares/bsips/issues/161#issuecomment-491244109) but not till the end I believe.

abitmore:
> Current BSIP only affects initial positions.

No, it affects updating of existing positions as well.

sschiessl-bcp:
> > Current BSIP only affects initial positions.
> 
> No, it affects updating of existing positions as well.

Yes, of course. I think automatic mechanism of the platform should also honor the asset owner desired buffer.

bitcrab:
> I am wondering if this might need to be reconsidered?
> 
> `Note: logic around *target collateral ratio* (TCR) is not affected.`
> 
> Current BSIP only affects initial positions. It is to be expected that positions live longer such that replenishing CR through margin call would happen more often as compared to creating a new position. Essentially, my suggestion is to also enforce ICR as minimum value for TCR (no new parameter). Could have a boolean flag to switch this on if owner desires it.
> 
> It was discussed [here](https://github.com/bitshares/bsips/issues/161#issuecomment-491244109) but not till the end I believe.

in my view it is good to limit TCR>=ICR, although it is not a must.

PR #233, opened 2019-10-02
Create BSIP78-80

ryanRfox:


ryanRfox:
I'm open to this possible progression:
Discussion 
Draft
Voting
Accepted | Rejected
Installed
Deferred

However, the additional "Discussion" status could be eliminated if we are diligent about requiring the author to produce an actual draft prior to assigning the BSIP number. The draft should be (IMO) text within the Issue's discussion, preferably the description at the top so it's easy to find the latest revision. Once the BSIP Review Team feel the draft text is sufficient, the BSIP number is assigned and the author can make the initial PR with the latest text.

abitmore:
In that sense I think it's better to not reserve 78 nor 79, at least don't put them in the list so far.

ryanRfox:
Shall I close this one, then create a new PR for 80 only? My desire will be to leave the assignments for 78-79 within the Issues, but wait until their respective draft is ready prior to updating each within the README in this case.

ryanRfox:
None of these actually have a draft spec "done" so, I'm going to close this PR for now. Once we get the Issues updated with the draft, we can PR to update the README. We will retain the numbering assignments as they are already on the respective Issues being discussed.

PR #234, opened 2019-10-02
Update status of several old BSIPs

abitmore:


README.md
```diff
@@ -14,9 +14,9 @@ Number             | Title                                                    |
 -------------------|----------------------------------------------------------|-------------------|----------------|--------
 [1](bsip-0001.md)  | BSIP Purpose and Guidelines                              | Fabian Schuh      | Informational  | Draft
 [2](bsip-0002.md)  | Refund Create Order Fees on Cancel                       | Daniel Larimer    | Protocol       | Superseded by [26](bsip-0026.md)
-[3](bsip-0003.md)  | Maker / Taker Market Fees Flag                           | Daniel Larimer    | Protocol       | Deferred
+[3](bsip-0003.md)  | Maker / Taker Market Fees Flag                           | Daniel Larimer    | Protocol       | Superseded by [81](https://github.com/bitshares/bsips/issues/229)
```
pmconrad: Hm, do we make a difference between "Superseded" and "Obsoleted"? Otherwise please make wording consistent throughout the document.

README.md
```diff
@@ -48,7 +48,7 @@ Number             | Title                                                    |
 [38](bsip-0038.md) | Add target collateral ratio option to short positions                       | Abit More | Protocol | Installed
 [39](bsip-0039.md) | Automatically approve proposals by the proposer                             | Fabian Schuh | Protocol | Draft
 [40](bsip-0040.md) | Custom active permission                             | Stefan Schießl | Protocol | Accepted - Milestone 1
-[41](bsip-0041.md) | Reduce MSSR of bitCNY from 1.1 to 1.05               | Jerry Liu      | Protocol | Accepted
+[41](bsip-0041.md) | Reduce MSSR of bitCNY from 1.1 to 1.05               | Jerry Liu      | Informational | Superseded by [59](bsip-0059.md)
```
pmconrad: I would leave the "Accepted" in (but still add "Superseded")

abitmore: IMHO Superseded implies Accepted or Installed.

ryanRfox: I feel BSIP41 is very specific to setting a single value and is now obsoleted by the BSIP59 voting process.  It seems somewhat disingenuous to say BSIP41 is accepted, as the value specified is 1.01 currently.  I also like the succinctness of "Superseded by BSIP". In this case, the reader must resort to another source for the current accepted value.

README.md
abitmore: If something is superseded by something newer, it is replaced because it has become old-fashioned or unacceptable. https://www.collinsdictionary.com/dictionary/english/supersede

I'm an English noob. Does "A superseded by B" implies that B was ever "in use"? Can we just use the word "replaced" which looks simpler?

I feel I'm wasting time discussing this.

README.md
abitmore: Leaving the "Accepted" here would confuse readers, because they may think 1.05 is the final (current) consensus, while it's not.

bsip-0003.md
```diff
@@ -2,7 +2,7 @@
     Title: Maker / Taker Market Fees Flag
     Authors: Daniel Larimer <Dan@cryptonomex.com>
              Fabian Schuh <Fabian@BitShares.eu>
-    Status: Deferred
+    Status: Obsoleted
```
ryanRfox: Suggest to mark it as "Obsoleted by <BSIP>" and include a reference to the current BSIP/discussion.

README.md
MichelSantos: I think that there can be a benefit to distinguishing "superceded" and "accepted" because "Version 2" of a BSIP might supercede "Version 1" but either one may or may not be accepted or installed.

abitmore: So what to do?

bsip-0003.md
abitmore: There is one.

README.md
MichelSantos: I think that two or more descriptions can be applied to a BSIP if it is appropriate.

Issue #71, opened 2018-03-23
About BSIP 39: make the approval optional


abitmore:
Since sometimes the proposer don't want to approve immediately, it's better if she has an option to decide whether approve.

abitmore:
> You recommend to extend "proposal_create" with another extension to add a "flag" for this? Something like, "add_proposers_approval"?

Yes. IMHO it's not a security related change, changing default behavior is not the best option. It's better to maintain backward compatibility, avoid breaking existing applications.

abitmore:
> How about an extension to the `proposal_create_operation` that let's you define whether to auto-approve or not?

That's the idea..

syalon:
Is this BSIP currently suspended? I see that the initial version of the [PR](https://github.com/bitshares/bitshares-core/pull/1368) has been submitted and is also considered backward compatible. We should probably complete the final draft of BSIP and conduct voting. What do you think? @xeroc

ryanRfox:
No @syalon this BSIP is still in draft state. 

@xeroc I feel this BSIP should be cleaned up slightly before we put it up for a vote:

- [ ] Discussion 
  - [ ] Add this Issue's URL along with https://github.com/bitshares/bitshares-core/issues/138
  - [ ] Add summary of of both threads at the bottom of the document
- [ ]  Abstract, Motivation and Rational - refactor these to include the discussed topics and outcome. They seem lacking depth.
- [ ] Specification - more detailed design
- [ ] Summary for Shareholders - include scope of proposals

In my opinion it was merged prior to "done" but that's just an opportunity to revise it now. The concept is good and should be put to vote.

abitmore:
Please review PR #235.

ryanRfox:
Thank you @abitmore for addressing the items in the list above. I ticked them off as complete. Let's proceed within the PR for review, merge and ultimately voting.

abitmore:
Done.

PR #235, opened 2019-10-02
BSIP39 update: optionally auto-approve the proposal

abitmore:
PR for #71.

### DIFF
--- a/bsip-0039.md
+++ b/bsip-0039.md
@@ -34,6 +34,16 @@ itself, and two approvals.
 The inconvenience and additional fee hinders adoption of this scheme and
 makes it unnecessary complicated.
 
+Due to lacking of an auto-approval feature, an ignorant user might fire a
+`proposal_create` operation to create a proposal and then immediately fire
+a `proposal_update` (i.e. approve) operation to approve the proposal.
+However, the final proposal ID is not known before the `proposal_create`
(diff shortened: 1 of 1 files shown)


bsip-0039.md
```diff
@@ -35,28 +36,33 @@ makes it unnecessary complicated.
...
+Giving the proposer an option to automatically approve the proposal
+after creation solves the issue.
 
 If the proposer is not part of the multisig-setup, having him approve
-the proposal automatically does affect the validity of the proposal
+the proposal automatically does not affect the validity of the proposal
 itself.
 
-# Specifications
+It's not a security related change, changing the default behavior is not
```
pmconrad: Actually the change to improve security indirectly.

A naive user might bundle a `proposal_create` and a `proposal_update` (i. e. approve) operation into one transaction to achieve the same effect. However, the final proposal ID is not known before the `proposal_create` is beyond the last irreversible block. So the naive user might inadvertently approve the wrong proposal.

ryanRfox:
The status is currently marked as "draft" and it **not** on the list I forwarded to BBF for voting. What does the code say? If the current implementation should be improved, I feel we can use BSIP39 to propose the desired specification and then post for voting.

bsip-0039.md
abitmore: Thanks. Updated.

PR #237, opened 2019-10-03
Create bsip-0077.md

ioBanker:


bsip-0078.md
```diff
@@ -0,0 +1,66 @@
...
+=================
+BitShares is aiming to become bridge of assets for market makers providing cross-currency liquidity as a software to individuals and corporate, supporting the ed25519-sha-256 pre-image hashing algorithm would enable BitShares to become compatible with the financial system transmission and it's future functions.
+
+Today banks are deploying solutions for Cross-Currency Settlement and FX Market Making.
+
+When those solutions are ILP-enabled, banks won’t have to change much – market makers will still hold accounts at their banks and they'll establish relationships with the other banks and market makers with whom they want to transact. ILP enables transactions across any number of different ledgers and exchanges using cryptographic escrow and a specialized two-phase commit protocol.
+
+Risks
+===============
+**Note** below is a discussion of HTLC risks in general, not risks inherent in implementing this BSIP.
```
pmconrad: This is duplicated from elsewhere, right? Please remove and leave only those risks that are specific to this proposal (if any).

bsip-0078.md
```diff
@@ -0,0 +1,66 @@
...
+This becomes even more important with atomic swaps. The shorter duration (a.k.a “inner”) contract should allow time to achieve its own irreversibility. And the longer duration (a.k.a. “outer”) contract must allow time for irreversibility of both.
+
+In addition, with an atomic swap both sides must consider the redemption time necessary. The creator of the inner contract must decide that should the redeemer redeem at the last moment, is there enough time to redeem the outer contract before the timelock expires.
+
+One final consideration of the timelock portion of the contract is the use of capital. Should the other party not accept, the funds in the contract are held until the timelock expires. Long expiration times could result in missed opportunity costs.
+
+Specifications
+=========
+
+**ed25519-sha-256**: Include this additional hashing algorithm to the list of hashes available to be used.
```
pmconrad: Please explain - is the addition of this hash function **necessary** and **sufficient** for participation in ILP? In other words, does ILP support other hashing functions like those that we already have, and does ILP demand other things as well ("crypto-conditions")?

bsip-0078.md
```diff
@@ -0,0 +1,66 @@
...
+
+**ILP PREIMAGE-SHA-256 Crypto-Conditions**: To calculate a condition and fulfillment in the proper format, you should use a [Crypto-Conditions](https://tools.ietf.org/html/draft-thomas-crypto-conditions-03) library such as [five-bells-condition](https://github.com/interledgerjs/five-bells-condition). For fulfillments, ILP recommends using one of the following methods to generate the fulfillment:
```
pmconrad: Above your proposal sounds like it is only about adding a hashing function. Now you say that we must "calculate a condition and fulfillment in the proper format". Whatever that means, it would probably be a much bigger change than the mere addition of a hash function.

bsip-0078.md
```diff
@@ -0,0 +1,66 @@
...
+
+Use a cryptographically secure source of randomness to generate at least 32 random bytes.
```
pmconrad: Generating preimage is client responsibility, not an on-chain protocol issue. It should not part of the specification.

bsip-0078.md
```diff
@@ -0,0 +1,66 @@
+```
+BSIP: 77
```
ryanRfox: Please change to BSIP82

ioBanker: Changed.

bsip-0078.md
ioBanker: Yes, it's duplicated from BSIP: 64, It has same risks of HASH160 algorithm.

bsip-0078.md
ioBanker: To satisfy a full atomic swap transactions between BitShares and XRP (ILP Blockchain) for example, A crypto condition will be communicated from "ILP offChain" user to "onChain" user, so user "onChain" will create HTLC using that condition with "ed25519-sha-256" algorithm, when the fulfillment appears at "offChain" side, the "onChain" user will be using that fulfillment to release the transaction "onChain", BitShares blockchain suppose to be validating if that fulfillment is true for the earlier provided condition in that HTLC.

bsip-0078.md
ioBanker: The hash function ed25519-sha-256 is not a single equation.

bsip-0078.md
ioBanker: It's a part of specification since it's not only used by client.

sschiessl-bcp:
What is the motivation of this BSIP in terms of additionally available use-cases?

ioBanker:
> What is the motivation of this BSIP in terms of additionally available use-cases?

I believe supporting XRP Escrow encryption algorithms within BitShares HTLC algorithms is a good step to the future of having atomic swaps between BitShares and CMC Ranked #3 XRP networks.

Also XRP has built in API which is supported by Blockchain coders themselves, this would help us to integrate some standard API calls to any XRP node and give users the ability to trust any XRP node to execute the escrow transactions.

I say if it's doable, lets do it, if it's complex and hard to do now, lets keep it in mind.

pmconrad:
If the goal is to connect our chain to XRP then we should start from there and see what's necessary to achieve it. Like I said above, the mere addition of a new hash function is not it.

PR #238, opened 2019-10-10
BSIP83: Decouple BitAssets from Platform Governance Process

jmjatlanta:
Update: The current text is being revised based on discussions with the community including the Committee, and will be updated in the future

### DIFF
--- a/bsip-0083.md
+++ b/bsip-0083.md
@@ -1,0 +1,74 @@
+```
+BSIP: 0083
+Title: Decouple BitAssets from Platform Governance Process
+Authors:    
+   Alfredo Garcia
+   Christopher Sanborn
+   John M. Jones
+   Michel Santos
+   Nathan Hourt
+   Peter Conrad
+   Ryan R. Fox
+Status: Draft
+Type: Protocol
+Obsoletes: BSIP-0016, BSIP-0058, BSIP-0059, BSIP-0075, BSIP-0076
+Created: 2019-10-07
+Discussion: https://github.com/bitshares/bsips/issues/239
+Worker: <TBD>
+```
+
+# Abstract
+This BSIP separates the governance of the technical security of the blockchain from the governance of BitAssets.  BitAssets are a collection of special case SmartCoin which conflate blockchain governance and market effectiveness. To best serve the BitShares protocol, the existing set of BitAssets must be reconfigured to separate the concerns of block production, price feeds, and market operations; that is to say: deterministic consensus, non-deterministic external data, and smart contract settings.
+
+Therefore, this BSIP defines a process for BitAssets management to ensure protocol security remains the singular focus of the block producer role, and give BitAssets the best chance for long-term viability by nominating a new custodian to quickly react to market conditions.  Furthermore, the portion of BTS collateral below the full valuation of a debt asset, as defined by its price feed, that is borrowed from the blockchain, shall not contribute to the voting power of the debtor.
+
+# Motivation
+Block production is an essential component of the BitShares platform itself, and as such should be expected to take an agnostic and neutral view towards the products deployed within the platform. Therefore, this BSIP will define a process for BitAsset management to ensure protocol security remains the singular focus of the block producer role and give BitAssets the best chance for long-term stability by nominating a new custodian to quickly react to market conditions. In order for the distributed consensus mechanism of DPOS to achieve adequate decentralization, it is imperative that block producers maintain strict independence from the politics of individual products deployed on the platform. 
+
+Governance is an important component of the core BTS token and effective application of governance throughout the BitShares protocol is imperative to the long-term success of the project.  A number of BSIPs have addressed the shortcomings of the existing vote calculation mechanism. This BSIP changes vote calculation which currently confers governance rights which are intended for assignment to the protocol.
+
+In short, this BSIP will put the BitAssets under the control of a designated user account. With this change, the newly appointed BitAssets manager will be able to independently maintain the parameters of their derivative asset without interference from the platform’s block producers and committee members. At the same time, the platform itself and its committee will not be unduly influenced by the demands of private product management. An advantage gained by this is that the BitAsset manager will be able to react appropriately to rapidly changing market conditions, as may be required to properly maintain a successful derivative asset.
+
+# Rationale
+Because the BitAssets are currently controlled by platform-wide governance mechanisms, and because these are not well suited to the management of smart coins, which may at times need to respond quickly to market conditions, the authors propose to reorganize these assets to utilize a directly appointed multisignature authority and thereby enable these assets to update in synchronization with market ebbs and flows.
+
+The authors recommend reassigning the blockchain-controlled BitAssets’ issuer field from committee-account to cn-vote, which is a multisignature account composed of seven different parties and having numerous proxy followers, and setting cn-vote as the sole feed producer, until cn-vote can delegate this role to accounts of its own appointment. The authors selected cn-vote for this role because cn-vote has demonstrated an active and vested interest in the stability and viability of the BitAsset products. Should cn-vote accept this appointment, the specification of the BitAsset parameters and feed production will become the responsibility of the cn-vote multisignature account.
+
+As a result of this assignment, the block producers will no longer provide feeds for BitAssets, and the new set of feed producers will no longer be affected by direct stakeholder voting.
+
+Because BitAsset management will no longer be affected by stakeholder voting, the vote-weighting calculations can be amended without detriment so that accounts holding BTS collateral positions will count only collateral above the value of their debt at the feed price. Collateral up to a 1.0 collateral-to-debt ratio is under lien to the BitAsset holder, and therefore does not represent a stake in the blockchain, and will no longer be counted as voting stake. Collateral in excess of a 1.0 collateral-to-debt ratio will continue to be counted in the debt holder’s voting stake.
+
+# Specification
+The blockchain software is designed to support assets with multisignature issuer authorities; therefore, the technical changes required for this proposal are straightforward. The following technical changes are required:
+
+**Assignment of Accumulated Fees**
+
+At the hardfork date, the accumulated fees of the existing BitAssets will be transferred to the committee-account.
+
+**Update to Committee-Controlled BitAsset Issuer Fields**
+
+At the hardfork date, the committee-controlled BitAsset objects' issuer fields will be updated to specify the cn-vote account (ID 1.2.1148184). The existing price feeds will be cleared, and the cn-vote account will be specified as the feed producer. After the hardfork, the cn-vote account will be able to assign other feed producers at its discretion.
+
+**Disable the witness_fed_asset and committee_fed_asset Flags**
+
+At the hardfork date, the `witness_fed_asset` and `committee_fed_asset` flags will be unset from the BitAssets, and will be disabled from use in future assets.
+
+**Amend total_core_in_orders to Track Limit and Collateral Orders Separately**
+
+The `total_core_in_orders` field of the `account_statistics_object` will be deprecated, and assets in limit orders will be tallied in a new `total_core_in_limit_orders` field while collateral will be tracked in a new `total_core_in_collateral_orders` map, keyed on the collateralized asset ID. During vote weight calculation, the `total_core_in_limit_orders` will add directly to the account’s voting weight; additionally, the `total_core_in_collateral_orders` which exceeds the feed price value of the collateralized debt will be added to the account’s voting weight as well.
+
+**Enable Remittance of Worker Proposals in Core Asset**
+
+Escrow agents administering existing worker proposals presently denominated in BitAssets shall be authorized to remit worker payments in BitShares core asset, or in a BitAsset selected at the discretion of the workers and escrow agents.
+
+# Discussion
+### Voting Power
+The voting power of debt holders will be reduced to only the portion of collateral that exceeds full collateral valuation of the debt assets that are borrowed from the blockchain. It is arguable who “owns” BTS that are locked up as collateral in BitAsset smart contracts. The authors of this BSIP hold the view that debt holders have received BitAssets in return for their collateral (which they may have sold on the market for more BTS), and thus have given up rights to the collateral itself, at least to the portion of collateral that represents the value of the BitAssets they have received. It is only logical that by exchanging the collateral for BitAssets they have given up the rights associated with the collateral, including its voting power.
+### Worker Proposals
+Worker proposals were originally denominated in BitAssets with the intent to add liquidity in their respective markets. With BitAssets now being placed under user control, liquidity is no longer the concern of the platform. A further implication of placing BitAssets under custodian control is that workers being paid in BitAssets would be exposed to the risk of devaluation and/or de-pegging at the custodian's discretion. This BSIP aims to remedy this by allowing different payout methods deemed less risky.
+
+# Summary
+In this BSIP the authors provide a unified solution to advance the project forward most efficiently. This proposal will decouple BitAssets from the rest of the BitShares platform with regard to governance. Instead, BitAssets can be controlled independently by a new BitAsset manager. The platform will be governed by the elected witnesses and committee, as before. This BSIP does not impact non-BitAsset Smart Coins.
+
+# Copyright
+This document is placed into the public domain.
--- a/bsip-0083.md
+++ b/bsip-0083.md
@@ -48,11 +48,11 @@ Subsequently, tokens of this new asset can be freely transferred or traded.
 
 **Add a feeding_authority extension to BitAssets**
 
-An extension will be added to `asset_update_bitasset_operation` that allows specifying an elected authority (see BSIP-0084) for determining the list of feed producers. `asset_publish_feed_operation` will be adapted to check that authority if the extension is present on a BitAsset.
+An extension will be added to `asset_update_bitasset_operation` that allows specifying an elected authority ([see BSIP-0084](bsip-0084.md)) for determining the list of feed producers. `asset_publish_feed_operation` will be adapted to check that authority if the extension is present on a BitAsset.
 
 **Addition of voting process through BITASSET.MANAGEMENT**
 
-Two new elected authorities (see BSIP-0084) will be created based on the BITASSET.MANAGEMENT asset with settings that match the current voting mechanism (regarding min/max_number and proxy_allowed). The first one is for controlling the new bitasset-manager account (see below), the second one for selecting price feeders.
+Two new elected authorities ([see BSIP-0084](bsip-0084.md)) will be created based on the BITASSET.MANAGEMENT asset with settings that match the current voting mechanism (regarding min/max_number and proxy_allowed). The first one is for controlling the new bitasset-manager account (see below), the second one for selecting price feeders.
 
 **Update to Committee-Controlled BitAsset Issuer Fields**
 
@@ -68,15 +68,13 @@ At the hardfork date, the `witness_fed_asset` and `committee_fed_asset` flags wi
 
 By approving this BSIP, the shareholders instruct the committee to set the active authority of the bitasset manager account to the entity elected to act as the bitasset-manager account, and to set the feeding_authority extension of all BitAssets controlled by the bitasset-manager account to the entity elected for the curation of the price feed list, after sufficient shareholders have voted on the new entities, but no later than two weeks after the hardfork date.
 
-**Enable Remittance of Worker Proposals in Core Asset**
-
-Escrow agents administering existing worker proposals presently denominated in BitAssets shall be authorized to remit worker payments in BitShares core asset, or in a BitAsset selected at the discretion of the workers and escrow agents.
-
 # Discussion
 
 ## Worker Proposals
 
-Worker proposals were originally denominated in BitAssets with the intent to add liquidity in their respective markets. With BitAssets now being placed under user control, liquidity is no longer the concern of the platform. A further implication of placing BitAssets under custodian control is that workers being paid in BitAssets would be exposed to the risk of devaluation and/or de-pegging at the custodian's discretion. This BSIP aims to remedy this by allowing different payout methods deemed less risky.
+Worker proposals were originally denominated in BitAssets with the intent to add liquidity in their respective markets, and advance them as a favored platform product. With BitAssets now being placed under the control of an independent manager, liquidity of a single particular product line is no longer the concern of the platform. A further implication of placing BitAssets under custodian control is that workers being paid in BitAssets would be exposed to the risk of devaluation and/or de-pegging at the custodian's discretion. Since this BSIP separates the BitAsset products from the platform, it may also sunset the tradition of preferentially denominating worker proposals in BitAssets, and free worker proposal authors to choose from a wider range of payout options, whether BitAsset or otherwise.
+
+ The authors of this BSIP recommend that worker proposals be open in selecting payout methods.
 
 ## Comparison with Recent Committee Proposal
 
--- a/bsip-0083.md
+++ b/bsip-0083.md
@@ -11,7 +11,7 @@ Authors:
    Ryan R. Fox
 Status: Draft
 Type: Protocol
-Obsoletes: BSIP-0016, BSIP-0058, BSIP-0059, BSIP-0075, BSIP-0076
+Obsoletes: BSIP-0016, BSIP-0058, BSIP-0059, BSIP-0076
 Created: 2019-10-07
 Discussion: https://github.com/bitshares/bsips/issues/239
 Worker: <TBD>
@@ -20,55 +20,61 @@ Worker: <TBD>
 # Abstract
 This BSIP separates the governance of the technical security of the blockchain from the governance of BitAssets.  BitAssets are a collection of special case SmartCoin which conflate blockchain governance and market effectiveness. To best serve the BitShares protocol, the existing set of BitAssets must be reconfigured to separate the concerns of block production, price feeds, and market operations; that is to say: deterministic consensus, non-deterministic external data, and smart contract settings.
 
-Therefore, this BSIP defines a process for BitAssets management to ensure protocol security remains the singular focus of the block producer role, and give BitAssets the best chance for long-term viability by nominating a new custodian to quickly react to market conditions.  Furthermore, the portion of BTS collateral below the full valuation of a debt asset, as defined by its price feed, that is borrowed from the blockchain, shall not contribute to the voting power of the debtor.
+Therefore, this BSIP defines a process for BitAssets management to ensure protocol security remains the singular focus of the block producer role, and give BitAssets the best chance for long-term viability by nominating a new custodian to quickly react to market conditions.
 
 # Motivation
 Block production is an essential component of the BitShares platform itself, and as such should be expected to take an agnostic and neutral view towards the products deployed within the platform. Therefore, this BSIP will define a process for BitAsset management to ensure protocol security remains the singular focus of the block producer role and give BitAssets the best chance for long-term stability by nominating a new custodian to quickly react to market conditions. In order for the distributed consensus mechanism of DPOS to achieve adequate decentralization, it is imperative that block producers maintain strict independence from the politics of individual products deployed on the platform. 
 
-Governance is an important component of the core BTS token and effective application of governance throughout the BitShares protocol is imperative to the long-term success of the project.  A number of BSIPs have addressed the shortcomings of the existing vote calculation mechanism. This BSIP changes vote calculation which currently confers governance rights which are intended for assignment to the protocol.
-
-In short, this BSIP will put the BitAssets under the control of a designated user account. With this change, the newly appointed BitAssets manager will be able to independently maintain the parameters of their derivative asset without interference from the platform’s block producers and committee members. At the same time, the platform itself and its committee will not be unduly influenced by the demands of private product management. An advantage gained by this is that the BitAsset manager will be able to react appropriately to rapidly changing market conditions, as may be required to properly maintain a successful derivative asset.
+In short, this BSIP will put the BitAssets under the control of a new user account controlled by the stakeholders through a new type of voting. With this change, the newly appointed BitAssets manager will be able to independently maintain the parameters of their derivative asset without interference from the platform’s block producers and committee members. At the same time, the platform itself and its committee will not be unduly influenced by the demands of private product management. An advantage gained by this is that the BitAsset manager will be able to react appropriately to rapidly changing market conditions, as may be required to properly maintain a successful derivative asset.
 
 # Rationale
-Because the BitAssets are currently controlled by platform-wide governance mechanisms, and because these are not well suited to the management of smart coins, which may at times need to respond quickly to market conditions, the authors propose to reorganize these assets to utilize a directly appointed multisignature authority and thereby enable these assets to update in synchronization with market ebbs and flows.
+Because the BitAssets are currently controlled by platform-wide governance mechanisms, and because these are not well suited to the management of smart coins, which may at times need to respond quickly to market conditions, the authors propose to reorganize these assets to utilize a dedicated stakeholder-voted multisignature authority and thereby enable these assets to update in synchronization with market ebbs and flows.
 
-The authors recommend reassigning the blockchain-controlled BitAssets’ issuer field from committee-account to cn-vote, which is a multisignature account composed of seven different parties and having numerous proxy followers, and setting cn-vote as the sole feed producer, until cn-vote can delegate this role to accounts of its own appointment. The authors selected cn-vote for this role because cn-vote has demonstrated an active and vested interest in the stability and viability of the BitAsset products. Should cn-vote accept this appointment, the specification of the BitAsset parameters and feed production will become the responsibility of the cn-vote multisignature account.
-
-As a result of this assignment, the block producers will no longer provide feeds for BitAssets, and the new set of feed producers will no longer be affected by direct stakeholder voting.
-
-Because BitAsset management will no longer be affected by stakeholder voting, the vote-weighting calculations can be amended without detriment so that accounts holding BTS collateral positions will count only collateral above the value of their debt at the feed price. Collateral up to a 1.0 collateral-to-debt ratio is under lien to the BitAsset holder, and therefore does not represent a stake in the blockchain, and will no longer be counted as voting stake. Collateral in excess of a 1.0 collateral-to-debt ratio will continue to be counted in the debt holder’s voting stake.
+The block producers will no longer provide feeds for BitAssets.
 
 # Specification
-The blockchain software is designed to support assets with multisignature issuer authorities; therefore, the technical changes required for this proposal are straightforward. The following technical changes are required:
+The following technical changes are required:
 
 **Assignment of Accumulated Fees**
 
 At the hardfork date, the accumulated fees of the existing BitAssets will be transferred to the committee-account.
 
+**Creation of BITASSET.MANAGER asset**
+
+At the hardfork date, a new BITASSET.MANAGER asset will be created with the committee account as its owner and all flags and permissions disabled. The initial distribution will be made equal to the current BTS distribution. BTS that are locked up in open orders, in collateral positions, in HTLC objects or in vesting balances will be assigned to the owners of the respective positions. Other BTS positions are out of scope (e. g. stealth).
+Subsequently, tokens of this new asset can be freely transferred or traded.
+
+**Add a feeding_authority extension to BitAssets**
+
+An extension will be added to `asset_update_bitasset_operation` that allows specifying an elected authority (see BSIP-0084) for determining the list of feed producers. `asset_publish_feed_operation` will be adapted to check that authority if the extension is present on a BitAsset.
+
+**Addition of voting process through BITASSET.MANAGER**
+
+Two new elected authorities (see BSIP-0084) will be created based on the BITASSET.MANAGER asset with settings that match the current voting mechanism (regarding min/max_number and proxy_allowed). The first one is for controlling the new bitasset-manager account (see below), the second one for selecting price feeders.
+
 **Update to Committee-Controlled BitAsset Issuer Fields**
 
-At the hardfork date, the committee-controlled BitAsset objects' issuer fields will be updated to specify the cn-vote account (ID 1.2.1148184). The existing price feeds will be cleared, and the cn-vote account will be specified as the feed producer. After the hardfork, the cn-vote account will be able to assign other feed producers at its discretion.
+At the hardfork date, a new bitasset-manager account will be created with its active authority set to the first of the new elected authorities, and with the owner authority controlled by the committee.
+The committee-controlled BitAsset objects' issuer fields will be updated to specify this new account.
+After the hardfork, the new account will be able to assign other feed producers at its discretion.
 
 **Disable the witness_fed_asset and committee_fed_asset Flags**
 
-At the hardfork date, the `witness_fed_asset` and `committee_fed_asset` flags will be unset from the BitAssets, and will be disabled from use in future assets.
-
-**Amend total_core_in_orders to Track Limit and Collateral Orders Separately**
-
-The `total_core_in_orders` field of the `account_statistics_object` will be deprecated, and assets in limit orders will be tallied in a new `total_core_in_limit_orders` field while collateral will be tracked in a new `total_core_in_collateral_orders` map, keyed on the collateralized asset ID. During vote weight calculation, the `total_core_in_limit_orders` will add directly to the account’s voting weight; additionally, the `total_core_in_collateral_orders` which exceeds the feed price value of the collateralized debt will be added to the account’s voting weight as well.
+At the hardfork date, the `witness_fed_asset` and `committee_fed_asset` flags will be unset from the committee-owned BitAssets, and will be disabled from use in future assets. The existing price feeds will be cleared. The BitAssets will have the feeding_authority extension set to the second of the new elected authorities.
 
 **Enable Remittance of Worker Proposals in Core Asset**
 
 Escrow agents administering existing worker proposals presently denominated in BitAssets shall be authorized to remit worker payments in BitShares core asset, or in a BitAsset selected at the discretion of the workers and escrow agents.
 
 # Discussion
-### Voting Power
-The voting power of debt holders will be reduced to only the portion of collateral that exceeds full collateral valuation of the debt assets that are borrowed from the blockchain. It is arguable who “owns” BTS that are locked up as collateral in BitAsset smart contracts. The authors of this BSIP hold the view that debt holders have received BitAssets in return for their collateral (which they may have sold on the market for more BTS), and thus have given up rights to the collateral itself, at least to the portion of collateral that represents the value of the BitAssets they have received. It is only logical that by exchanging the collateral for BitAssets they have given up the rights associated with the collateral, including its voting power.
+
 ### Worker Proposals
+
 Worker proposals were originally denominated in BitAssets with the intent to add liquidity in their respective markets. With BitAssets now being placed under user control, liquidity is no longer the concern of the platform. A further implication of placing BitAssets under custodian control is that workers being paid in BitAssets would be exposed to the risk of devaluation and/or de-pegging at the custodian's discretion. This BSIP aims to remedy this by allowing different payout methods deemed less risky.
 
 # Summary
-In this BSIP the authors provide a unified solution to advance the project forward most efficiently. This proposal will decouple BitAssets from the rest of the BitShares platform with regard to governance. Instead, BitAssets can be controlled independently by a new BitAsset manager. The platform will be governed by the elected witnesses and committee, as before. This BSIP does not impact non-BitAsset Smart Coins.
+
+In this BSIP the authors provide a unified solution to advance the project forward most efficiently. This proposal will decouple BitAssets from the rest of the BitShares platform with regard to governance. Instead, BitAssets can be controlled independently by a new BitAsset manager elected by interested parties. The platform will be governed by the elected witnesses and committee, as before. This BSIP does not impact non-BitAsset Smart Coins.
 
 # Copyright
 This document is placed into the public domain.
--- a/bsip-0083.md
+++ b/bsip-0083.md
@@ -74,8 +74,6 @@ By approving this BSIP, the shareholders instruct the committee to set the activ
 
 Worker proposals were originally denominated in BitAssets with the intent to add liquidity in their respective markets, and advance them as a favored platform product. With BitAssets now being placed under the control of an independent manager, liquidity of a single particular product line is no longer the concern of the platform. A further implication of placing BitAssets under custodian control is that workers being paid in BitAssets would be exposed to the risk of devaluation and/or de-pegging at the custodian's discretion. Since this BSIP separates the BitAsset products from the platform, it may also sunset the tradition of preferentially denominating worker proposals in BitAssets, and free worker proposal authors to choose from a wider range of payout options, whether BitAsset or otherwise.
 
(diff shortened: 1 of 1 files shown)


ryanRfox:
ACK

bsip-0083.md
```diff
@@ -0,0 +1,68 @@
...
+    John M. Jones
+    Michel Santos
+    Nathan Hourt
+    Peter Conrad
+    Ryan R. Fox
+Status: Draft
+Type: Protocol
+Obsoletes: BSIP-0016, BSIP-0058, BSIP-0059, BSIP-0075, BSIP-0076
+Created: 2019-10-07
+Discussion: https://github.com/bitshares/bitshares-core/issues/tbd
```
pmconrad: Use PR 238 or create new issue?

bsip-0083.md
```diff
@@ -0,0 +1,68 @@
...
+
+Governance is an important component of the core BTS token and effective application of governance throughout the BitShares protocol is imperative to the long-term success of the project.  A number of BSIPs have addressed the shortcomings of the existing vote calculation mechanism. This BSIP changes vote calculation which currently confers governance rights which are intended for assignment to the protocol.
+
+In short, this BSIP will privatize the BitAssets. With this change, the newly appointed BitAssets manager will be able to independently maintain the parameters of their derivative asset without interference from the platform’s block producers and committee members. At the same time, the platform itself and its committee will not be unduly influenced by the demands of private product management. An advantage gained by this is that the BitAsset manager will be able to react appropriately to rapidly changing market conditions, as may be required to properly maintain a successful derivative asset.
```
pmconrad: Is "privatize" a good term?

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
+```
+BSIP: 83
```
pmconrad: Make it 0083 please

bsip-0083.md
pmconrad: Created https://github.com/bitshares/bsips/issues/239

bitcrab:
this is a complete joking, right?

TechsUsInc:
In reference to voting;
In my opinion this is also a perfect opportunity to include valuable BTS that is either locked up in Open Orders and/or staking and apply it towards voting with the hourly maintenance. 

In reference to BP's;
I suggest each BP submit substantial collateral on regular intervals which both helps the development of the network, and it forces them to be vested in the platform. Failure to produce satisfactory from the community results in forfeiture of vesting.

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Enable Remittance of Worker Proposals in Core Asset**
+
+Escrow agents administering existing worker proposals presently denominated in BitAssets shall be authorized to remit worker payments in BitShares core asset, or in a BitAsset selected at the discretion of the workers and escrow agents.
+
+# Discussion
+### Voting Power
+The voting power of debt holders will be reduced to only the portion of collateral that exceeds full collateral valuation of the debt assets that are borrowed from the blockchain. It is arguable who “owns” BTS that are locked up as collateral in BitAsset smart contracts. The authors of this BSIP hold the view that debt holders have received BitAssets in return for their collateral (which they may have sold on the market for more BTS), and thus have given up rights to the collateral itself, at least to the portion of collateral that represents the value of the BitAssets they have received. It is only logical that by exchanging the collateral for BitAssets they have given up the rights associated with the collateral, including its voting power.
```
abitmore: I still think the changes about voting should not be in this BSIP. Better discuss it in the scope of [BSIP-0024 - Locking Bitshares away as 'Bitshares Influence' for voting privileges on the BTS DEX](https://github.com/bitshares/bsips/blob/master/bsip-0024.md) and/or https://github.com/bitshares/bsips/issues/83.

thul3:
The most stupid idea i have read about giving power of price feeds to a small group.

"It is arguable who “owns” BTS"

Is it really ?Because i thought bitshares is being run like a business and for businesses and the current law and business practise is cleary adding voting rights of bought stocks to the debtor.
Who is taking the risk ?The debtor ?So bitshares is borrowing me money but don't allow me to buy full worth BTS ?

Just BTS without voting rights ?
If i buy with the borrowed money EOS coins will my EOS coins also have no voting rights ?

thul3:
The most stupid idea i have read about giving power of price feeds to a small group.

"It is arguable who “owns” BTS"

Is it really ?Because i thought bitshares is being run like a business and for businesses and the current commercial law and business practise is cleary adding voting rights of bought stocks via margin to the debtor.
Who is taking the risk to lose BTS?The debtor ?Or the simple holder ?
Biitshares is borrowing me money but don't allow me to buy fully worth BTS,
but just BTS without voting rights ?

If i buy with the my borrowed money EOS coins will my EOS coins also have no voting rights ?

Defenition of Collateral:
Collateral is an asset that a lender accepts as security for a loan. If the borrower defaults on the loan payments, the lender can seize the collateral and resell it to recoup the losses.

So till the collateral is not being seized the ownership of that collateral is clearly owned by the debtor

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+The authors recommend reassigning the blockchain-controlled BitAssets’ issuer field from committee-account to cn-vote, which is a multisignature account composed of seven different parties and having numerous proxy followers, and setting cn-vote as the sole feed producer, until cn-vote can delegate this role to accounts of its own appointment. The authors selected cn-vote for this role because cn-vote has demonstrated an active and vested interest in the stability and viability of the BitAsset products. Should cn-vote accept this appointment, the specification of the BitAsset parameters and feed production will become the responsibility of the cn-vote multisignature account.
+
+As a result of this assignment, the block producers will no longer provide feeds for BitAssets, and the new set of feed producers will no longer be affected by direct stakeholder voting.
```
grctest: Which bitassets are in scope of this BSIP? Just bitCNY? What about the following assets owned by the committee-account? GAS, DIESEL, KRW, BTC, OIL, SILVER, GOLD, TRY, SGD, HKD, RUB, SEK, NZD, CNY, MXN, CAD, CHF, AUD, GBP, JPY, EUR, USD, CROWDFUN, DRAGON, SERVICE, BTCSHA, TESTME, BANCOR, NASDAQC, HANGSENG, NIKKEI, SHENZHEN, SHANGHAI, ARS, RUBLE, XCD, RUPEES, FORINT

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+Because BitAsset management will no longer be affected by stakeholder voting, the vote-weighting calculations can be amended without detriment so that accounts holding BTS collateral positions will count only collateral above the value of their debt at the feed price. Collateral up to a 1.0 collateral-to-debt ratio is under lien to the BitAsset holder, and therefore does not represent a stake in the blockchain, and will no longer be counted as voting stake. Collateral in excess of a 1.0 collateral-to-debt ratio will continue to be counted in the debt holder’s voting stake.
```
grctest: This should be a separate BSIP

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+At the hardfork date, the accumulated fees of the existing BitAssets will be transferred to the committee-account.
+
+**Update to Committee-Controlled BitAsset Issuer Fields**
+
+At the hardfork date, the committee-controlled BitAsset objects' issuer fields will be updated to specify the cn-vote account (ID 1.2.1148184). The existing price feeds will be cleared, and the cn-vote account will be specified as the feed producer. After the hardfork, the cn-vote account will be able to assign other feed producers at its discretion.
+
+**Disable the witness_fed_asset and committee_fed_asset Flags**
+
+At the hardfork date, the `witness_fed_asset` and `committee_fed_asset` flags will be unset from the BitAssets, and will be disabled from use in future assets.
```
grctest: By "and will be disabled from use in future assets." you mean assets issued by cn-vote, not the whole bitshares network right? Because witnesses should still be free to feed 'private' smartcoins on the bitshares network in the future.

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Amend total_core_in_orders to Track Limit and Collateral Orders Separately**
+
+The `total_core_in_orders` field of the `account_statistics_object` will be deprecated, and assets in limit orders will be tallied in a new `total_core_in_limit_orders` field while collateral will be tracked in a new `total_core_in_collateral_orders` map, keyed on the collateralized asset ID. During vote weight calculation, the `total_core_in_limit_orders` will add directly to the account’s voting weight; additionally, the `total_core_in_collateral_orders` which exceeds the feed price value of the collateralized debt will be added to the account’s voting weight as well.
+
+**Enable Remittance of Worker Proposals in Core Asset**
+
+Escrow agents administering existing worker proposals presently denominated in BitAssets shall be authorized to remit worker payments in BitShares core asset, or in a BitAsset selected at the discretion of the workers and escrow agents.
```
grctest: What does this have to do with your proposal? They were always free to do this, no?

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+### Worker Proposals
+Worker proposals were originally denominated in BitAssets with the intent to add liquidity in their respective markets. With BitAssets now being placed under user control, liquidity is no longer the concern of the platform. A further implication of placing BitAssets under custodian control is that workers being paid in BitAssets would be exposed to the risk of devaluation and/or de-pegging at the custodian's discretion. This BSIP aims to remedy this by allowing different payout methods deemed less risky.
```
grctest: * Originally denominated in bitassets to offer workers a stable source of income, not just liquidity reasons.
* Payment in BTS is more risky than holding bitassets; bitassets have decreased in value less this year than BTS has.
* I don't think this is neccessary to include in this BSIP.

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Amend total_core_in_orders to Track Limit and Collateral Orders Separately**
+
+The `total_core_in_orders` field of the `account_statistics_object` will be deprecated, and assets in limit orders will be tallied in a new `total_core_in_limit_orders` field while collateral will be tracked in a new `total_core_in_collateral_orders` map, keyed on the collateralized asset ID. During vote weight calculation, the `total_core_in_limit_orders` will add directly to the account’s voting weight; additionally, the `total_core_in_collateral_orders` which exceeds the feed price value of the collateralized debt will be added to the account’s voting weight as well.
+
+**Enable Remittance of Worker Proposals in Core Asset**
+
+Escrow agents administering existing worker proposals presently denominated in BitAssets shall be authorized to remit worker payments in BitShares core asset, or in a BitAsset selected at the discretion of the workers and escrow agents.
+
+# Discussion
```
grctest: Would be great to add:
* Pro/cons for changing asset ownership
* Risks (regulatory responsibilities, asset flag misuse risk)
* Stakeholder impact (bitasset holder, debtor, bts holder, exchanges that integrated bitassets)

bsip-0083.md
ryanRfox: Yes, all BitAssets are in scope of this BSIP; all other SmartCoins are out of scope. The specification section must be updated to note all of them. @oxarbitrage I believe you have the list and can update this?

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+# Specification
+The blockchain software is designed to support assets with multisignature issuer authorities; therefore, the technical changes required for this proposal are straightforward. The following technical changes are required:
+
+**Assignment of Accumulated Fees**
+
+At the hardfork date, the accumulated fees of the existing BitAssets will be transferred to the committee-account.
+
+**Update to Committee-Controlled BitAsset Issuer Fields**
+
+At the hardfork date, the committee-controlled BitAsset objects' issuer fields will be updated to specify the cn-vote account (ID 1.2.1148184). The existing price feeds will be cleared, and the cn-vote account will be specified as the feed producer. After the hardfork, the cn-vote account will be able to assign other feed producers at its discretion.
```
grctest: It doesn't require a hardfork to change asset owner nor to specify cn-vote as the sole price feeder.

grctest: If the asset ownership is transferred from the committee-account to cn-vote, these assets would lose their 'bit' prefix & return to being simply smartcoins not bitassets. 

Why doesn't cn-vote instead create their own set of smartcoins to prove changes work before requesting implementation on the committee owned bitassets?

bsip-0083.md
ryanRfox: Ultimately each escrow agent has final authority over remittances Andy hey typically desire to remain outside of governance issues. The intent here is to allow the Community to signal to escrow agents that changes to BItAssets mechanics after the initial worker proposal agreements were formed should be considered for future remittances.

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+In short, this BSIP will put the BitAssets under the control of a designated user account. With this change, the newly appointed BitAssets manager will be able to independently maintain the parameters of their derivative asset without interference from the platform’s block producers and committee members. At the same time, the platform itself and its committee will not be unduly influenced by the demands of private product management. An advantage gained by this is that the BitAsset manager will be able to react appropriately to rapidly changing market conditions, as may be required to properly maintain a successful derivative asset.
+
+# Rationale
+Because the BitAssets are currently controlled by platform-wide governance mechanisms, and because these are not well suited to the management of smart coins, which may at times need to respond quickly to market conditions, the authors propose to reorganize these assets to utilize a directly appointed multisignature authority and thereby enable these assets to update in synchronization with market ebbs and flows.
```
grctest: They're owned by the committee and fed by the witnesses. Ideally change requests would be sent to the committee, who would then issue agreed upon changes to the price feed publishers directly, instead of BSIPs attempting to directly influence/coerce witnesses to make changes (potentially against the wishes of the committee & to the detriment of network stability).

Not long ago I requested the committee to change the max price feed publisher quantity to 25, this was done via bitsharestalk after which a proposal was created and voted upon by the committee. This didn't require a BSIP, nor should bitasset settings/policies require BSIP votes, instead the committee could vote on a committee proposal for changing the asset (voting on proposed asset description to indicate to feeders the economic policies).

Why doesn't cn-vote join the committee for influence, rather than request full control?

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+Governance is an important component of the core BTS token and effective application of governance throughout the BitShares protocol is imperative to the long-term success of the project.  A number of BSIPs have addressed the shortcomings of the existing vote calculation mechanism. This BSIP changes vote calculation which currently confers governance rights which are intended for assignment to the protocol.
+
+In short, this BSIP will put the BitAssets under the control of a designated user account. With this change, the newly appointed BitAssets manager will be able to independently maintain the parameters of their derivative asset without interference from the platform’s block producers and committee members. At the same time, the platform itself and its committee will not be unduly influenced by the demands of private product management. An advantage gained by this is that the BitAsset manager will be able to react appropriately to rapidly changing market conditions, as may be required to properly maintain a successful derivative asset.
```
grctest: > At the same time, the platform itself and its committee will not be unduly influenced by the demands of private product management.

This will still remain a risk, as the committee is at the controls of the tune-able network settings & witnesses/committee members would still be able to feed non-bitassets in the future which could run into similar demanding scenarios.

Is cn-vote prepared to become the new focus of unduly influence by demanding private product managers?

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Motivation
+Block production is an essential component of the BitShares platform itself, and as such should be expected to take an agnostic and neutral view towards the products deployed within the platform. Therefore, this BSIP will define a process for BitAsset management to ensure protocol security remains the singular focus of the block producer role and give BitAssets the best chance for long-term stability by nominating a new custodian to quickly react to market conditions. In order for the distributed consensus mechanism of DPOS to achieve adequate decentralization, it is imperative that block producers maintain strict independence from the politics of individual products deployed on the platform. 
+
+Governance is an important component of the core BTS token and effective application of governance throughout the BitShares protocol is imperative to the long-term success of the project.  A number of BSIPs have addressed the shortcomings of the existing vote calculation mechanism. This BSIP changes vote calculation which currently confers governance rights which are intended for assignment to the protocol.
```
grctest: I really like the idea of removing the 1x CR from active vote weight & allowing the remaining surplus BTS as debtor vote weight, but I think it'd be best in its own BSIP since this would require a hardfork change where as the other requested changes do not and have fairly different focuses.

thul3:
"If given the choice, I don't believe that voting power correction for margin positions will ever make it."

"I believe it won't pass the voting, so why we waste time/efforts/funds including it in the BSIP? Having it in the BSIP means other features in the BSIP won't pass as well."

"That's the whole point. Including it in the BSIP forces the affected parties to compromise."

"Why to force?"

"because as you said...it would never get voted on its own."


At least people should know what it really is about

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Motivation
+Block production is an essential component of the BitShares platform itself, and as such should be expected to take an agnostic and neutral view towards the products deployed within the platform. Therefore, this BSIP will define a process for BitAsset management to ensure protocol security remains the singular focus of the block producer role and give BitAssets the best chance for long-term stability by nominating a new custodian to quickly react to market conditions. In order for the distributed consensus mechanism of DPOS to achieve adequate decentralization, it is imperative that block producers maintain strict independence from the politics of individual products deployed on the platform.
```
grctest: If your proposed asset owner change for all bitassets passed, the witnesses wouldn't have a singular focus on block production as there are many user created smartcoins which can be fed by witnesses. I have 4 myself fed by several witnesses, configured in a private price feed publisher list.

How was witness decentralization put at threat by the politics of bitassets? By having votes withdrawn if they didn't support rushed BSIPs targeting bitassets? Witnesses are an inherently political role, they have to campaign for their positions in the first place, so politics will always be at the forefront of their minds.

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+Obsoletes: BSIP-0016, BSIP-0058, BSIP-0059, BSIP-0075, BSIP-0076
+Created: 2019-10-07
+Discussion: https://github.com/bitshares/bsips/issues/239
+Worker: <TBD>
+```
+
+# Abstract
+This BSIP separates the governance of the technical security of the blockchain from the governance of BitAssets.  BitAssets are a collection of special case SmartCoin which conflate blockchain governance and market effectiveness. To best serve the BitShares protocol, the existing set of BitAssets must be reconfigured to separate the concerns of block production, price feeds, and market operations; that is to say: deterministic consensus, non-deterministic external data, and smart contract settings.
+
+Therefore, this BSIP defines a process for BitAssets management to ensure protocol security remains the singular focus of the block producer role, and give BitAssets the best chance for long-term viability by nominating a new custodian to quickly react to market conditions.  Furthermore, the portion of BTS collateral below the full valuation of a debt asset, as defined by its price feed, that is borrowed from the blockchain, shall not contribute to the voting power of the debtor.
```
grctest: As mentioned below, after such a change the witnesses and potentially the committee would still have other focuses than their block producer role, they're still able to feed many user created smartcoins and run non block producing services/api's.

bsip-0083.md
```diff
@@ -0,0 +1,74 @@
...
+Status: Draft
+Type: Protocol
+Obsoletes: BSIP-0016, BSIP-0058, BSIP-0059, BSIP-0075, BSIP-0076
+Created: 2019-10-07
+Discussion: https://github.com/bitshares/bsips/issues/239
+Worker: <TBD>
+```
+
+# Abstract
+This BSIP separates the governance of the technical security of the blockchain from the governance of BitAssets.  BitAssets are a collection of special case SmartCoin which conflate blockchain governance and market effectiveness. To best serve the BitShares protocol, the existing set of BitAssets must be reconfigured to separate the concerns of block production, price feeds, and market operations; that is to say: deterministic consensus, non-deterministic external data, and smart contract settings.
```
grctest: The abstract does not mention the primary focus of this BSIP which is requesting an asset ownership transfer from the committee-account to the cn-vote account.

What are the 'must have' bitasset reconfigurations which the committee is unable to perform?

Is the committee unable to achieve deterministic consensus on bitasset settings? What proof is there for this?

Why can't non-deterministic external data be a pull request for witness price feeds instead of requesting asset ownership transfer?

bsip-0083.md
grctest: I don't think it falls under 24/83 since they propose a bitshares 'stake power' like in Steem instead of eliminating the vote weight of the <=1x CR. I do agree that a separate BSIP for this idea would be appropriate, I like it.

bitProfessor:
1、I agree that witnesses are only responsible for production block and no longer responsible for feeding price.
2、i disagree giving power of price feeds to a small group.This is the stupidest thing.

bitcrab:
it make sense to do the decouple, however we need be careful to select the new manager and price feeder, the new smartcoin manager should be similar to committee-account, a multisig account with members be voted by stakeholders, and the new price feeders should also be voted by stakeholders.

the adjustment of voting power should not be included in this BSIP.

and I suggest to add one more content -  to pay salary to committee members as well as the new smartcoin manager.

bitcrab:
after communicating with some people, I feel maybe BSIP management is the key problem? this BSIP comes out because BSIP76 comes out not perfectly following the core team process and make core team blamed, although it get enough support in voting?

but, even we do the decouple, are we sure that there will be no BSIP any more in smartcoin governance?

I don't think so. maybe we need to review the BSIP management process.

bsip-0083.md
pmconrad: The account will continue to exist and will continue to be the owner of the SmartCoins. It is in the responsibility of the asset owner to care for the asset, which can include handing it over to someone else.

bsip-0083.md
pmconrad: Every smartcoin can define a list of accounts that should be price feeders. There is nothing that stops anyone from listing witness accounts there.
What we propose here is to remove the flag that connects the witness *job* to the pricefeeder job.

bsip-0083.md
pmconrad: > They were always free to do this, no?

All (to my knowledge) *existing* worker proposals specify payout in BitAssets. Some escrows have refused payout in BTS because it contradicts the wording of the proposals.

bsip-0083.md
pmconrad: The stability promise of BitAssets has been destroyed by BSIP-76. bitCNY is effectively pegged to 0.22 BTS/bitCNY now, with the same volatility as BTS.

The goal of this BSIP is to decouple platform governance (including worker proposals) from BitAsset governance. Demanding payout in BitAssets for platform workers means they remain coupled.

btsfav:
I think it's a good idea to add specialized groups / break down the management. However, there will never be a "Maker" success, that train is long gone, so don't expect magic. Also, adding China to this BSIP makes it completely biased and worthless. Not the best move, I can not vote for it.

bsip-0083.md
pmconrad: The "Discussion" section is meant for including a condensed version of discussion that *has taken place* elsewhere. As in "We have also considered X but decided to go with Y because Z."
So feel free to discuss, the outcome can than be added here if it is deemed important.

bsip-0083.md
shulthz: > 
> 
> The account will continue to exist and will continue to be the owner of the SmartCoins. It is in the responsibility of the asset owner to care for the asset, which can include handing it over to someone else.

What you said is meaningless, and what you said is shirking the responsibility to a third party.

bsip-0083.md
shulthz: > 
> 
> Every smartcoin can define a list of accounts that should be price feeders. There is nothing that stops anyone from listing witness accounts there.
> What we propose here is to remove the flag that connects the witness _job_ to the pricefeeder job.

This is not a correct solution, and you didn't give a clear governance programmes.

bsip-0083.md
shulthz: When the MSSR=110%, the bitasset still have the same volatility as BTS, but i didn't find you say that, why?

pmconrad:
We have suggested the cn-vote account because it was our impression that cn-vote represents a large majority of the interests of debt holders. It seems this was a misjudgement on our part. We are of course open for other solutions. We are in no way connected with the cn-vote account, nor have we approached cn-vote wrt this proposal before we announced it yesterday.

Handing the BitAssets over to a normal account has the advantage that it is a comparatively simple change and could be executed quickly when approved. The same goes with assigning cn-vote as the sole price feeder - it is a simple first step, and afterwards the new BitAsset manager can assign whoever he wishes.

bsip-0083.md
pmconrad: We have had block producers being voted out after reliably doing their job for *years*, because they didn't respond to price feed demands from debt holders within *hours*. This is endangering the stability of the blockchain.
By decoupling BitAssets from blockchain governance we hope to prevent witnesses being targeted because of the pricefeed, simply because they will not be feed providers anymore.

bsip-0083.md
pmconrad: This BSIP is not the result of a request from cn-vote. Also see my remarks [below](https://github.com/bitshares/bsips/pull/238#issuecomment-540940222).

bsip-0083.md
pmconrad: See my remarks [below](https://github.com/bitshares/bsips/pull/238#issuecomment-540940222) wrt our proposing the cn-vote account.

sschiessl-bcp:
> We have suggested the cn-vote account because it was our impression that cn-vote represents a large majority of the interests of debt holders. It seems this was a misjudgement on our part. We are of course open for other solutions. We are in no way connected with the cn-vote account, nor have we approached cn-vote wrt this proposal before we announced it yesterday.
> 
> Handing the BitAssets over to a normal account has the advantage that it is a comparatively simple change and could be executed quickly when approved. The same goes with assigning cn-vote as the sole price feeder - it is a simple first step, and afterwards the new BitAsset manager can assign whoever he wishes.

For a more complex solution, the committee-account could remain the owner, while being able to appoint a SmartCoin manager that is allowed to set certain parameters / options, couldn't?

bsip-0083.md
pmconrad: That is not true. Only the spread was higher.

shulthz:
> 
> 
> We have suggested the cn-vote account because it was our impression that cn-vote represents a large majority of the interests of debt holders. It seems this was a misjudgement on our part. We are of course open for other solutions. We are in no way connected with the cn-vote account, nor have we approached cn-vote wrt this proposal before we announced it yesterday.
> 
> Handing the BitAssets over to a normal account has the advantage that it is a comparatively simple change and could be executed quickly when approved. The same goes with assigning cn-vote as the sole price feeder - it is a simple first step, and afterwards the new BitAsset manager can assign whoever he wishes.

This bsip didn't clear something and not give a clear management of the programme.

What it talk about is the asset of community, not belong to the cn-vote or someone.
If the cn-vote didn't accept that, and what did you want to do, force the cn-vote or destroy these assets? We need to know what's your thought clearly for that.

and didn't you consider or clealy talk about the fault of sole price feeder?

pmconrad:
> What it talk about is the asset of community
> If the cn-vote didn't accept that,

I think it is clear that this proposal will not pass without the approval of the cn-vote *and* the majority of the community. Like I said, cn-vote may have been a poor choice due to a misjudgement on our side. We are open for other suggestions.

> and didn't you consider or clealy talk about the fault of sole price feeder?

I don't understand the question, sorry.

bangzi1001:
> 
> 
> after communicating with some people, I feel maybe BSIP management is the key problem? this BSIP comes out because BSIP76 comes out not perfectly following the core team process and make core team blamed, although it get enough support in voting?
> 
> but, even we do the decouple, are we sure that there will be no BSIP any more in smartcoin governance?
> 
> I don't think so. maybe we need to review the BSIP management process.

BSIP76 is relevant to committee and witlessness but not relevant to core team. I suggest any changes on smartcoin parameter or price feed should post at committee's Github:
https://github.com/bitshares/committee-tools/

shulthz:
> Like I said, cn-vote may have been a poor choice due to a misjudgement on our side. 
> 
Who is " our"?

> > and didn't you consider or clealy talk about the fault of sole price feeder?
> 
> I don't understand the question, sorry.
The question is very simple: The danger of the sole price feeder.

bsip-0083.md
shulthz: > 
> because they didn't respond to price feed demands from debt holders within _hours_. 
> By decoupling BitAssets from blockchain governance we hope to prevent witnesses being targeted because of the pricefeed, simply because they will not be feed providers anymore.

I think you need carefully say that “demands from debt holders”. This is vote result.

If your want to prevent witnesses being targeted because of the pricefeed, design a witnees of price feed mechanism.
1.The price feeder didn't need to produce block and just need to provide the price feed, 
2.The witnees of price feed will get the reward from the bitasset which he feed.
3.The witnees of price feed should vote by the stakeholder.
...
I found we learned nothing from the MAKERDAI.

bsip-0083.md
shulthz: Why is not true, when MSSR=110%, we have 20% volatility in bitasset.
When GS happen, this also destroyed the stability promise of BitAssets which has happend in bitusd.

pmconrad:
> Who is " our"?

The authors of the BSIP.

> The question is very simple: The danger of the sole price feeder.

The new manager can appoint more than one price feeder if he wants. The single feeder is only a simple solution until the new manager takes up his duties.

shulthz:
> 
> 
> > Who is " our"?
> 
> The authors of the BSIP.

So the authors of the BSIP is not very rigorous, this should not happen.
cn-vote only is a proxy like baozi or beos.

> 
> > The question is very simple: The danger of the sole price feeder.
> 
> The new manager can appoint more than one price feeder if he wants. The single feeder is only a simple solution until the new manager takes up his duties.

This is not the solution, just push the responsibility to the new manager and want the new manager to solve this problem. It is very irresponsible.

I can easily got the information from other program for the price feed: 
https://blog.makerdao.com/introducing-oracles-v2-and-defi-feeds/

bsip-0083.md
shulthz: The debt holders still should have the full voting power even if  full or not full collateral.

The reason is very simple, when another DEX have collateral mechanism of bts, and pledge a lot of bts, when price drop down, all the debt didn't get margin called in the market asap, so you tell me this DEX didn't have the voting power?

Agorise:
Relying on algos to humans is our preference. 
 
We're also the largest holder of bitgold and bitsilver so of course we don't want our family's life savings to be put in jeopardy either, which it clearly is right now. 
 
If bitassets (mpa's) are to be taken out of the control of the chain itself, a couple thoughts: 
1. Think of each mpa as a product that a business is selling. Competition is good. 
2. Each business that feeds an mpa competes to have people use their mpa. Fees, spreads and other things could be used to throw a few sats at the people who host the feed for that mpa too, and let those businesses decide what kind of fees they want to charge. 
3. NOT a fan at all of salaries as bitcrab mentioned above. Salaries most of the time, over time, will breed apathy, laziness and an un-competitive attitude towards the product or service that you business sells. Think more in terms of co-ops. The more successful your business is, the more fees/income will be generated, and hence distributed to the feed producers/publishers. 
4. Create more competing mpa's and have those businesses feeds/pegs/accuracy/use compete with one another. bitWheat, bitCattle, bitCorn, bitPlatinum
 
If for example Bitshares has 20 mpa's, then Bitshares is also large enough now to have 20 competing committees who run those 20 products. Competition breeds innovation, efficiency and many times, success. 
 
Oh, and reducing the world's "whale problem" a bit is probably also of benefit.

TechsUsInc:
How does the terminology of this BSIP not infer a replace-as-fee or extended block header hard fork?

bsip-0083.md
grctest: >We have had block producers being voted out after reliably doing their job for years, because they didn't respond to price feed demands from debt holders within hours. This is endangering the stability of the blockchain.

Indeed the act of intimidating/coercing witnesses into implementing a BSIP prior to its WP activation (and before the BSIP left draft status) is deplorable ([related bitsharestalk thread](https://bitsharestalk.org/index.php?topic=29659.msg336083#msg336083)).

Should the new asset owner position have been put to public bid, rather than nomination? Given the following most recent public cn-vote rep message on telegram threatening to replace core around the same time this bsip came out raises concerns.

> [09.10.19 11:47 cn-vote]
> The core development team should not forget what they are and what they should do. Do your own job, because there are already many excellent and cheaper teams who have repeatedly sought to compete for the core team.
> BM left, and the bitshares have been running steadily for so many years. Therefore, no one is indispensable. Don't forget who you are and what you should do. Do your job well and everyone will always support you. Otherwise, there will be more suitable teams and people. Thank you all. A job like this high salary and not so busy can be difficult to find. Many excellent teams have been coveted for a long time.

---

> I think you need carefully say that “demands from debt holders”. This is vote result.

It was activated in production prior to any WP activating: https://bitsharestalk.org/index.php?topic=29653.msg335831#msg335831

bsip-0083.md
grctest: >  What we propose here is to remove the flag that connects the witness job to the pricefeeder job.
Given the committee never provides price feeds themselves, why offer committee as an option either?
An advantage of using the witness fed setting is that it's a permanent setting, so if I got hit by a bus and all witnesses switched for new witnesses in 10 years time it would still be fully operational.

bsip-0083.md
grctest: The risks and stakeholder impact can be estimated without a prior discussion having taken place.

In some of my past BSIPs I've used the discussions section to preemptively cover both for/against perspectives for issues informally raised outwith Github.

bsip-0083.md
grctest: > so you tell me this DEX or CEX didn't have the full voting power?

Perhaps we could encourage users to return BTS from CEX to the DEX? CEX voting with user funds (even proxying their vote weight) is not appropriate.

froooze:
I see the need for a new group/token called price-feeder. 
Asset owner needs to buy price-feed-tokens and creates a whitelist for price-feeder.  
Price-feeder gets selected and paid by asset owner to provide price feeds.

MichelSantos:
The current text is being revised based on discussions with the Committee, and will be updated in the future

bsip-0083.md
pmconrad: Removed this point.

bsip-0083.md
pmconrad: Replaced cn-vote

bsip-0083.md
pmconrad: Replaced cn-vote

bsip-0083.md
pmconrad: Removed.

pmconrad:
After much discussion we have made two significant amendments to the proposal:
1. The voting weight change has been removed.
2. We have replaced the suggestion to appoint "cn-vote" as the bitasset manager with a voting mechanism. The committee will retain full control over the manager account, but can hand over day-to-day operations to a different entity.

bsip-0083.md
```diff
@@ -0,0 +1,80 @@
...
+
+**Update to Committee-Controlled BitAsset Issuer Fields**
+
+At the hardfork date, a new bitasset-manager account will be created with its active authority set to the first of the new elected authorities, and with the owner authority controlled by the committee.
+The committee-controlled BitAsset objects' issuer fields will be updated to specify this new account.
+After the hardfork, the new account will be able to assign other feed producers at its discretion.
+
+**Disable the witness_fed_asset and committee_fed_asset Flags**
+
+At the hardfork date, the `witness_fed_asset` and `committee_fed_asset` flags will be unset from the committee-owned BitAssets, and will be disabled from use in future assets. The existing price feeds will be cleared. The BitAssets will have the feeding_authority extension set to the second of the new elected authorities.
```
abitmore: It's dangerous to clear existing price feeds. It may lead to unexpected consequences.

bsip-0083.md
```diff
@@ -0,0 +1,80 @@
...
+
+An extension will be added to `asset_update_bitasset_operation` that allows specifying an elected authority (see BSIP-0084) for determining the list of feed producers. `asset_publish_feed_operation` will be adapted to check that authority if the extension is present on a BitAsset.
+
+**Addition of voting process through BITASSET.MANAGER**
+
+Two new elected authorities (see BSIP-0084) will be created based on the BITASSET.MANAGER asset with settings that match the current voting mechanism (regarding min/max_number and proxy_allowed). The first one is for controlling the new bitasset-manager account (see below), the second one for selecting price feeders.
+
+**Update to Committee-Controlled BitAsset Issuer Fields**
+
+At the hardfork date, a new bitasset-manager account will be created with its active authority set to the first of the new elected authorities, and with the owner authority controlled by the committee.
```
abitmore: Since people need time to vote, it would be a better idea to transfer the assets at a later date. IMHO we should not transfer the assets until there is a good-to-use UI to update votes.

bsip-0083.md
```diff
@@ -0,0 +1,80 @@
...
+The following technical changes are required:
+
+**Assignment of Accumulated Fees**
+
+At the hardfork date, the accumulated fees of the existing BitAssets will be transferred to the committee-account.
+
+**Creation of BITASSET.MANAGER asset**
+
+At the hardfork date, a new BITASSET.MANAGER asset will be created with the committee account as its owner and all flags and permissions disabled. The initial distribution will be made equal to the current BTS distribution. BTS that are locked up in open orders, in collateral positions, in HTLC objects or in vesting balances will be assigned to the owners of the respective positions. Other BTS positions are out of scope (e. g. stealth).
+Subsequently, tokens of this new asset can be freely transferred or traded.
```
abitmore: Also air drop to centralized exchanges, BEOS custodian accounts and etc? Great efforts on communications may be needed to make this protocol upgrade going smoothly.

bsip-0083.md
pmconrad: Without a valid feed neither forced settlement nor borrowing will be possible, until the new feeders take up their duties. This I think will happen quickly, because it is known in advance, and interested parties have already proven that they can act quickly when they see the need.

Thanks to the threshold price I wouldn't expect any forced settlements, and stopping the borrowing briefly shouldn't have any negative consequences.

Anything else?

bsip-0083.md
pmconrad: This BSIP is about *what* to do, not *when*.
Voting is largely controlled by proxies. Would it be helpful to create and pre-set proxies for these new elected authorities?

bsip-0083.md
pmconrad: Yes. It is up to the users who hold their BTS on CEXs or with BEOS to care about their voting rights. Most probably don't *want* to vote anyway, because otherwise they wouldn't do this.

Also, this will not be the first time that users will be pre-warned to hold funds in their own account if they want to participate in a sharedrop.

bsip-0083.md
pmconrad: Also note that both the new bitasset manager and the committee will have full control over the bitassets, i. e. if any problems arise the committee can set a fixed list of price feeders until voting kicks in.

shulthz:
Some bitasset didn't have the threshold price, what will happen to them?

bsip-0083.md
abitmore: It doesn't sound like you're serious.

It's possible that at the hf time the price is above the threshold, or the threshold BSIP has been voted out.

bsip-0083.md
abitmore: It does define when. "At the hardfork date" implies there is only one time defined for the hard fork.

bsip-0083.md
pmconrad: Even then, a brief interruption with ample pre-warning time is unlikely to cause much damage.

bsip-0083.md
pmconrad: The important part about this proposal is that it clearly defines that bitasset management should be handed over from the committee to an independent asset manager. Allowing to postpone this for an indefinite amount of time is counterproductive.

pmconrad:
> What's that mean? "interested parties""?

Whoever is interested in voting on this. And still has the voting tokens, obviously.

> Some bitasset didn't have the threshold price, what will happen to them?

This BSIP applies to all committee-owned bitassets, regardless of threshold price.

shulthz:
> 
> 
> > What's that mean? "interested parties""?
> 
> Whoever is interested in voting on this. And still has the voting tokens, obviously.
> 

This should be the stakeholder of BTS, not the "interested parties", please make the words more accurate.

> > Some bitasset didn't have the threshold price, what will happen to them?
> 
> This BSIP applies to all committee-owned bitassets, regardless of threshold price.

It's very dangerous like what @abitmore said.

Need more serious and specific action.

MichelSantos:
I, as an author, support this proposal

MichelSantos:
> > Creation of BITASSET.MANAGER asset
> 
> If this creation is necessary? please explain.

This is the new voting token for the new BitAsset manager.

"The initial distribution will be made equal to the current BTS distribution."
"Subsequently, tokens of this new asset can be freely transferred or traded."

shulthz:
> 
> 
> > > Creation of BITASSET.MANAGER asset
> > 
> > 
> > If this creation is necessary? please explain.
> 
> This is the new voting token for the new BitAsset manager.
> 

What's that mean of “new voting token”？

bsip-0083.md
shulthz: I don't think the “Creation of BITASSET.MANAGER asset” is necessary.

This will also push the bitasset to be a third party, it's no different with before words.

What's the right method is which the stakeholder of BTS vote the price feeder and committee of bitasset.

shulthz:
I don't think the “Creation of BITASSET.MANAGER asset” is necessary.

This will also push the bitasset to be a third party, it's no different with before words.

What's the right method is which the stakeholder of BTS vote the price feeder and committee of bitasset.

pmconrad:
> This will also push the bitasset to be a third party, it's no different with before words.
> What's the right method is which the stakeholder of BTS vote the price feeder and committee of bitasset.

As the title of the BSIP implies, this is exactly the point. We suggest to decouple the bitAssets from the platform. Initially, the BITASSET.MANAGER distribution will match BTS distribution. This can and will change, when the new asset starts being traded independently from BTS. "interested parties" can buy more of these tokens, and those who aren't interested can sell them.

> This should be the stakeholder of BTS, not the "interested parties"

You are of course entitled to have a different opinion, but in the opinion of the authors what should happen is what this BSIP proposes.

bsip-0083.md
```diff
@@ -0,0 +1,80 @@
...
+
+In short, this BSIP will put the BitAssets under the control of a new user account controlled by the stakeholders through a new type of voting. With this change, the newly appointed BitAssets manager will be able to independently maintain the parameters of their derivative asset without interference from the platform’s block producers and committee members. At the same time, the platform itself and its committee will not be unduly influenced by the demands of private product management. An advantage gained by this is that the BitAsset manager will be able to react appropriately to rapidly changing market conditions, as may be required to properly maintain a successful derivative asset.
```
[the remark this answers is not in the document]
pmconrad: Please note that the "Motivation" sections is meant to explain why the BSIP was created. There is really no point in arguing about the wording.

bsip-0083.md
pmconrad: How about this:
1. We remove the automatic clearing and replacement of the price feeds from the BSIP.
2. We mandate that the committee switch to manual feeder selection for all committee-owner bitassets within two weeks after the BSIP has been approced.
3. We mandate that the committee (or bitasset manager) switch the price feed to the new elected authority within two weeks after the hardfork activation.

abitmore: I think we can use 2 or 3 timestamps for the protocol upgrade.

thul3:
""interested parties" can buy more of these tokens, and those who aren't interested can sell them."

Is this a joke ?

thul3:
or the second trojan horse from dev team trying to create a possibility to give control over bitassets to private group ?

thul3:
Update: The current text is being revised based on discussions with the community including the Committee, and will be updated in the future

" As the title of the BSIP implies, this is exactly the point. We suggest to decouple the bitAssets from the platform. Initially, the BITASSET.MANAGER distribution will match BTS distribution. This can and will change, when the new asset starts being traded independently from BTS. "interested parties" can buy more of these tokens, and those who aren't interested can sell them."

COMMITTEE said cleary that it do not want to give away bitassets management to third parties.

bsip-0083.md
pmconrad: Replaced cn-vote.

bsip-0083.md
pmconrad: Removed.

pmconrad:
> COMMITTEE said cleary that it do not want to give away bitassets management to third parties.

This is reflected in the current text. The committee account will be the owner of the new manager account, which gives them full control.

shulthz:
> It is also our opinion that currently BitAssets _are_ effectively controlled by a small group of debt holders. The fact that witnesses have been coerced into providing a fake price feed clearly proves this.

This is not the fact, BSIP76 is voted in by the stakeholder of bts, if someone disagree it, he needs to bid for votes, this is very fair for everyone.
No body can coerce witnesses, but the witnesses need to follow the vote.
And don't call "fake price feed", if the authors always have this thought in the price feed, this will can't make this BSIP better, and this BSIP will can't make somthing better.

> This BSIP attempts to clean up the current governance mess by creating a clear separation of duties.

Whether you admit it or not, the debt holders has a big Voting weights in the governance now.
The debt holder is tightly connected with the community and a part of community, not "interested parties" or "private group", they are inalienable part to the platform and community.
If the authors have the hostile attitude to the debt holder, can't make thing better.

pmconrad:
> No body can coerce witnesses, but the witnesses need to follow the vote.

Fact is that witnesses were voted out for not providing a fake feed even before the vote was active.

shulthz:
>Fact is that witnesses were voted out for not providing a fake feed even before the vote was active.

This is a preconceived idea, every wittness will vote out for something, it was decided by the stakeholder, and the stakeholder can't forecast the behavior of the wittness which will be actived.

bsip-0083.md
pmconrad: Resolved through elected entities.

bsip-0083.md
pmconrad: Changed text for a smoother transition:
* no removal of feeders, only of flags
* instruct committee to enable new elected entities after voting has kicked in

bsip-0083.md
pmconrad: IMO it is easier to let the committee activate the elected entities manually. It also gives us more flexibility if things don't quite work out as planned. And it is less work than implementing another hardfork.

pmconrad:
That is not for you to decide single-handedly!

sschiessl-bcp:
> Transferring voting rights from BTS to another token is not acceptable and should never be an option. PR reverted.

BTS holders may decide it, unlikely but who knows. Given the controversy this bsip might still have been merged prematurely. 

What process do we have for conflicts like this?

jmjatlanta:
Evidently, the current process is if Abit likes it, he pushes it through without discussion. If he doesn't like it, he reverts it even if there was thorough discussion. I think the process needs improvement.

thul3:
Who liked it outside of the core team ?

Am asking cause i didn't saw the committee supporting it or any proxy.

pmconrad:
> Who liked it outside of the core team ?

That is beside the point. No single member has a veto right to anything.

thul3:
Would be nice if core could also focus on non LTM refferals to get bitshares real growth instead discussing on something which clearly has no chance to pass.

I asked core to have a look at it how it can be done since i'm not a tech guy and i guess this is currently the most important part for bitshares to get growth.Would be nice to see Core focusing on that the same way as you focus on this one.

shulthz:
Transferring voting rights from BTS to another token is not acceptable and should never be an option.
We shouldn't approved any PR which clearly has no chance to pass, this is a waste of time and money.

MichelSantos:
@shulthz I was unaware that anything could be considered "unacceptable" given the BSIPs that have have been presented to voters during the past year

Issue #81, opened 2018-05-26
New BSIP: voting with non-core asset, and asset-specific committee


abitmore:
Currently only CORE can vote for certain things E.G. witnesses, workers and the committee.

It's been demanded by the community for a long time that, holders of other assets can vote with their assets, for things related to those assets.

Another thing is asset-specific committee. Currently it's able to set committee of an asset to top-N holders. I think it would be useful if the committee can be decided by voting.

jiefangqian:
support

bangzi1001:
Support! Very useful and very powerful feature.

xeroc:
Would that be feasible computationally? It needs to go through all the holders of different assets and accumulate the respective stake.

abitmore:
> Would that be feasible computationally? It needs to go through all the holders of different assets and accumulate the respective stake.

Sure. There are fees related to balance the situation.

abitmore:
Just think out loud: it's possible that the blockchain charge a maintenance fee from the asset's fee pool for every vote re-tallying, and the fee can be related to quantity of holders of the asset. If no enough BTS in the fee pool, deactivate voting feature for the asset, can even deduct some BTS from fee pool in this case for the efforts done. To be fair, the asset committee can decide when to re-tally votes, but not re-tally at every maintenance interval.

shulthz:
maybe we need set a threshold for the DAC to open this function.

TheTaconator:
Is the idea to permit bitCNY holders to be permitted to vote on what? The market fee for bitCNY?

If they are permitted to vote on witnesses, workers, and committee, how should the their holdings be valued relative to BTS stake held by others?

Does the idea permit holders of GATEWAY.BTC be permitted to vote on witnesses, workers, and committee?

abitmore:
Witnesses, workers, and committee, global fee schedule and etc are CORE token's business, not others'.

This proposal is about UIA business. For example, OBITS holders may like to vote for the percentage market fee rate on OPEN.X assets.

grctest:
I like the idea - could encourage UIA usage more 👍

sschiessl-bcp:
This would come in handy with allowing users (for a hefty fee) to create accounts that have dynamic permissions like committee, but linked to an UIA. Thoughts?

thontron:
This is a good way to make businesses more autonomous/limit trust in single party.

We should identify the resource costs and if a fee can offer value for both businesses and stakeholders.

cloud-8:
[...]

PR #242, opened 2019-10-14
BSIP-84: Token-based voting

pmconrad:
For #81

### DIFF
--- a/bsip-0084.md
+++ b/bsip-0084.md
@@ -1,4 +1,4 @@
-    BSIP: 
+    BSIP: 0084
     Title: Elections based on non-core asset
     Authors: Peter Conrad
     Status: Draft
@@ -177,14 +177,22 @@ The chain logic for adjusting account balances is modified as follows:
 
 After the time of the hardfork, a new type of `special_authority` is allowed. This new type wraps an `elected_authority_object`.
 
+If the wrapped `elected_authority` has been deleted, the `special_authority` will block, i. e. it cannot authorize anything anymore.
 
 # Discussion
 
+## Deletion
+
+Deleting an authority that is still being used, e. g. as the active authority of an account, can cause problems.
+
+The problem is, if we disallow deletion if the authority is in use anywhere, then it is easy for someone to block an authority from being deleted.
+
+OTOH someone who uses an elected authority that is not under his control can be expected to coordinate this with the authority owner. Setting an elected authority as owner on an account is strongly discouraged, because authority deletion would block the account. Use as an active authority (or feed producer) can easily be repaired by the account owner (or asset owner).
 
 # Risks
 
-This BSIP has an impact on the performance of balance-changing operations. It is believed that the overall impact will be low, because only few assets will be affected.
-
+* This BSIP has an impact on the performance of balance-changing operations. It is believed that the overall impact will be low, because only few assets will be affected.
+* Deleting an authority that is still being used, e. g. as the active authority of an account, can cause problems.
 
 # Summary for Shareholders
 
--- a/bsip-0084.md
+++ b/bsip-0084.md
@@ -39,7 +39,7 @@ Because maintaining elected authorities is expensive (in terms of node resources
 
 ## 1. New asset flag "voting_allowed"
 
-A new asset flag/permission "voting_allowed" will be introduced. At the time of the hardfork, all existing UIAs will have the corresponding permission set. Future assets can have the permission set only if they are UIAs not MPAs nor PMs.
+A new asset flag/permission "voting_allowed" will be introduced. At the time of the hardfork, all existing assets except BTS will have the corresponding permission set.
 
 As usual with flags, the flag can be changed only if the permission is set. The permission can be unset any time, but can be set only when supply is zero.
 
--- a/bsip-0084.md
+++ b/bsip-0084.md
@@ -1,0 +1,195 @@
+    BSIP: 
+    Title: Elections based on non-core asset
+    Authors: Peter Conrad
+    Status: Draft
+    Type: Protocol
+    Created: 2019-10-12
+    Discussion: https://github.com/bitshares/bsips/issues/81
+
+# Abstract
+
+From the beginning, the BitShares blockchain has offered the ability to vote with the core token, and to have elected accounts govern the blockchain.
+
+For specific use-cases it can be desirable to have a similar mechanism for voting and elections where the votes are counted in relation not to BTS but to some other asset. An example might be a community of people who are interested in a specific topic.
+
+This BSIP proposes changes that will enable elections based on dedicated assets.
+
+
+# Motivation
+
+The feature has been requested from independent businesses as well as from within the community. In addition, it paves the way for a to-be-proposed change to BitAsset governance, i. e. the decoupling of BitAsset management from blockchain governance.
+
+
+# Rationale
+
+There are fundamental differences between the mechanics proposed here and the mechanics already in place for BTS-based voting.
+
+BTS balances are affected by almost every operation, due to transaction fees. The voting tokens will only be affected when they are being used.
+
+BTS is used in a multitude of ways, e. g. as collateral, as the counterpart in most active markets, as payment for workers and witnesses, as cashback for fees and so on. Contrarily, it is assumed that the primary purpose of the voting tokens will be voting. They are unlikely to be used as collateral.
+
+These differences allow for various simplifications and optimizations. In particular, we propose to allow only liquid balances for voting. Because these presumably change rarely in comparison to the number of distinct balances, it is more efficient to recalculate votes on the fly instead of once per maintenance interval (see STEEM for comparison).
+
+Furthermore, we make no distinction between voting and elections. Voting (as in making a yes/no decision) can be emulated with an election where only the votes for two designated candidates are counted and compared to each other. Depending on voting rules (to be defined externally on a case-by-case basis), the one with more votes wins, or perhaps the one with an absolute majority of eligible votes.
+
+
+# Specifications
+
+## 1. New asset flag "voting_allowed"
+
+A new asset flag/permission "voting_allowed" will be introduced. At the time of the hardfork, all existing UIAs will have the corresponding permission set. Future assets can have the permission set only if they are UIAs not MPAs nor PMs.
+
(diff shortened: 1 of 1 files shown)


bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+
+**Fields:**
+
+* `account_id_type creator` - the account to pay the fee, also the only one who can delete the authority
+* `asset_id_type voting_asset` - the asset on which to base the voting
+* `unsigned_int num_members` - the number of authority members to vote on if fixed, or 0 otherwise
+* `unsigned_int min_members` - the minimum number of authority members to elect, or 0 if fixed
+* `unsigned_int max_members` - the maximum number of authority members that can be elected, or 0 if fixed
+* `flat_set<account_id_type> candidates` - a list of candidates eligible for voting, or empty
+* `optional<asset_id_type> candidate_holders` - and asset that candidates must hold to be eligible for voting
```
MichelSantos: Should "and asset" be "any asset"?

pmconrad: Fixed "an asset". Thanks!

bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+As usual with flags, the flag can be changed only if the permission is set. The permission can be unset any time, but can be set only when supply is zero.
+
+The flag must not be used before the time of the hardfork.
+
+
+## 2. New operation "create_elected_authority"
+
+**Note 1:** Since the overall computational overhead for this voting mechanism is significant, the height of the fee should reflect this.
+
+**Note 2:** The new asset flag `voting_allowed` is only checked for this operation. If the flag is removed from an asset, any existing elected authorities are unaffected.
```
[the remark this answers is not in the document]
pmconrad: Yes.

bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+The operation must not be used before the time of the hardfork.
+
+* `authority` must exist.
+* `owner == authority.creator`
+
+**Evaluation:**
+
+* The fee stored in `authority` is returned to `owner`.
+* The `elected_authority_object` is deleted.
+* All related objects, including the `authority` as well as all user votes, are also deleted.
```
abitmore: I think this gives the creator too big power in some scenarios, it seems too risky for the accounts that has the elected authority as owner/active, as well as other entities that utilize the elected authority.

bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+
+**Note:** Afterwards, accounts that have the `authority` still referenced in its `owner` authority will be unable to ever change their owner again. Accounts that have the `authority` referenced in their `active` authority will be usable only by their `owner` authority until the `active` authority has been changed.
+
+
+## 4. New operation "update_asset_vote"
```
abitmore: What is this operation for? The name looks confusing.

bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+## 1. New asset flag "voting_allowed"
+
+A new asset flag/permission "voting_allowed" will be introduced. At the time of the hardfork, all existing UIAs will have the corresponding permission set. Future assets can have the permission set only if they are UIAs not MPAs nor PMs.
+
+As usual with flags, the flag can be changed only if the permission is set. The permission can be unset any time, but can be set only when supply is zero.
+
+The flag must not be used before the time of the hardfork.
+
+
+## 2. New operation "create_elected_authority"
```
abitmore: We usually put the verb at the end, no?

bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+* If `candidate_holders` is present the `candidates` must be empty.
+* `vote_until`, if present, must be in the future
+
+**Evaluation:**
+
+A new object type `elected_authority_object` is introduced. The operation creates such an object using the fields from the operation.
+
+A new, empty authority is created. The desired number of members for the authority is set to `max(num_members,min_members)`.
+
+Half of the operation fee is set aside and stored in the `elected_authority_object`.
```
abitmore: What will this half be used? Do we need new operations to deposit to it and/or withdraw from it?

bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+
+The chain logic for adjusting account balances is modified as follows:
+
+* For each voting slate of the balance's owner with the same asset type as is being updated:
+    * If the voting slate points to a proxy, adjust the proxy's proxy token count according to the balance delta.
+    * Apply the balance delta to the vote tally according to the user's (or proxy's if set) voting slate.
+    * Reflect the resulting differences to the election outcome in the respective `authority` object.
+
+
+## 6. `account_update_operation`
```
abitmore: How about account_create_operation?

bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+    * `votes_to_add`, `votes_to_remove` and `number` are applied to the user's voting slate.
+    * Let `voting_balance` be the sum of the `voter`'s token balance of the asset plus his proxy token balance.
+    * The voting delta is the `voting_balance` for each vote to be added and for the new `number`, the negative `voting_balance` for each vote to be removed and the previous `number`.
+    * The voting delta in votes is applied to the vote tally of `authority`.
+    * The resulting differences in the election outcome (if any) are reflected in the respective `authority` object.
+
+**Note:** the intent is to have vote tallying work in the same way it is currently performed when voting with BTS. The same goes for determining the number of accounts that make up the authority, unless it is fixed.
+
+
+## 5. `adjust_balance`
```
abitmore: Does it imply that only balance will be counted for voting? How about vesting balances, orders, collateral and etc?

pmconrad: Yes, only liquid balances count. See "Rationale" for an explanation.

bsip-0084.md
pmconrad: I wanted to keep it simple.

bsip-0084.md
pmconrad: It will be returned upon deletion.
Elected authorities are expensive for the chain. This mechanism is meant as an incentive to actually delete elected authorities that are no longer in use.

bsip-0084.md
pmconrad: Yes, I have thought about how to handle this. The problem is, if we disallow deletion if the authority is in use anywhere, then it is easy for someone to block an authority from being deleted. 

OTOH I think someone who uses an elected authority that is not under his control can be expected to coordinate this with the authority owner. Setting an elected authority as owner on an account is a bad idea, because as you say, authority deletion would block the account. Use as an active authority (or feed producer) can easily be repaired by the account owner (or asset owner).

I therefore think it is better to delete the authority anyway, and let those who are using it deal with it.

bsip-0084.md
abitmore: I guessed it. Please add the rationale to the document.

bsip-0084.md
abitmore: Please add these to the discussions section and/or a risks section. Thanks.

bsip-0084.md
abitmore: Won't support? I'm fine with it. Perhaps add a note.

bsip-0084.md
```diff
@@ -0,0 +1,195 @@
...
+The feature has been requested from independent businesses as well as from within the community. In addition, it paves the way for a to-be-proposed change to BitAsset governance, i. e. the decoupling of BitAsset management from blockchain governance.
+
+
+# Rationale
+
+There are fundamental differences between the mechanics proposed here and the mechanics already in place for BTS-based voting.
+
+BTS balances are affected by almost every operation, due to transaction fees. The voting tokens will only be affected when they are being used.
+
+BTS is used in a multitude of ways, e. g. as collateral, as the counterpart in most active markets, as payment for workers and witnesses, as cashback for fees and so on. Contrarily, it is assumed that the primary purpose of the voting tokens will be voting. They are unlikely to be used as collateral.
```
abitmore: > Contrarily, it is assumed that the primary purpose of the voting tokens will be voting.  They are unlikely to be used as collateral.

Personally I won't assume this, but the opposite.

bsip-0084.md
abitmore: Understood. I think it's better to note this in the discussions section.

bsip-0084.md
pmconrad: Hm, interesting. Anything specific you have in mind?

IMO in order to be used as collateral, a token would need to have some intrinsic value, which I don't see in a voting token.

bsip-0084.md
pmconrad: Done.

bsip-0084.md
pmconrad: Renamed to `elected_authority_vote`

bsip-0084.md
pmconrad: Renamed

bsip-0084.md
pmconrad: Done.

bsip-0084.md
pmconrad: Done.

bsip-0084.md
pmconrad: Done.

bsip-0084.md
abitmore: A voting token would not have value if it tracks the voting weight of another token well (which is hard), in that case it is a simple utility token carries one role for another token. When we have this BSIP (token-based voting), I'd expect that people would directly vote with the tokens they already have but not create new tokens only for voting, because it would be economically much cheaper.

This is also the reason that I don't agree with the MANAGER token idea proposed in BSIP83, because the new token can't track the distribution of BTS well in the long run. That idea means to privatize committee-owned smartcoins, which is something similar to STEALTH, and would divide the community.

On the other hand, there would also be use cases that a voting token is needed, e.g. if the voting weights are time-based or derived from another token.

bsip-0084.md
```diff
@@ -0,0 +1,213 @@
...
+
+These differences allow for various simplifications and optimizations. In particular, we propose to allow only liquid balances for voting. Because these presumably change rarely in comparison to the number of distinct balances, it is more efficient to recalculate votes on the fly instead of once per maintenance interval (see STEEM for comparison).
```
abitmore: On the other hand, the simplification that only counts liquid balances does limit the potential of the feature. There could be use cases that require tokens in other forms to be counted in voting as well. I do agree to implement the simplified version first though.

Issue #240, opened 2019-10-13
BSIP 85: Maker order creation fee discount


abitmore:
BSIP: 0085
    Title: Maker order creation fee discount
    Author: Abit More <https://github.com/abitmore>
    Status: Draft
    Type: Protocol
    Created: 2019-10-13
    Discussion: https://github.com/bitshares/bsips/issues/240
    Worker: TBD

# Abstract

This BSIP proposes a protocol change so that the committee can define
different fee rates for maker orders and taker orders.

# Motivation

To improve liquidity of the BitShares DEX.

# Rationale

The maker-taker fee model is adopted widely in centralized exchanges and
helped them to attract trading activities.

In the BitShares DEX, for each filled order, the owner of the order need to
pay an order creation fee and conditionally a market fee.
[BSIP 81](https://github.com/bitshares/bsips/blob/master/bsip-0081.md) described
a way to apply the maker-taker fee model to the market fee. This BSIP
describes a way to apply the maker-taker fee model to the order creation fee.

# Specification

Add a new global parameter `maker_fee_discount_percent` which can
be updated by the committee only after the protocol upgrade.
Initial value of that parameter is `0`.
Valid range of that parameter is `[0, 100%]`.

When a limit order is filled or partially filled for the first time,
technically, when its `deferred_fee` field is non-zero,
* if the order is a take order, process the deferred fee as before;
* if the order is a maker order, 
  * if the order creation fee was paid in BTS, return
  `round_down(deferred_fee * maker_fee_discount_percent)` to the owner,
  then process the remaining deferred fee as before;
  * if the order creation fee was paid in another asset, return
  `round_down(deferred_paid_fee * maker_fee_discount_percent)`
  to the owner, return
  `round_down(deferred_fee * maker_fee_discount_percent)` to
  the fee pool of the asset, then process the remaining deferred fee
  and deferred paid fee as before.

# Discussion

As of writing, to incentivize market making, the `limit_order_create`
operation fee is very small. However, since there is only one fee,
the fee for consuming liqudity is also very small.
This BSIP gives the committee a tool to increase fee for consuming
liquidity while still keeping a low cost for market makers to provide
liquidity.

# Copyright
This document is placed in the public domain.

abitmore:
@froooze

abitmore:
@ryanRfox I guess this BSIP can be assigned with number 84?

ryanRfox:
[...]

PR #243, opened 2019-10-14
Add BSIP85: Maker order creation fee discount

abitmore:
PR for #240.

### DIFF
--- a/bsip-0085.md
+++ b/bsip-0085.md
@@ -1,7 +1,7 @@
     BSIP: 0085
     Title: Maker order creation fee discount
     Author: Abit More <https://github.com/abitmore>
-    Status: Draft
+    Status: Accepted
     Type: Protocol
     Created: 2019-10-13
     Discussion: https://github.com/bitshares/bsips/issues/240
--- a/bsip-0085.md
+++ b/bsip-0085.md
@@ -10,11 +10,11 @@
 # Abstract
 
 This BSIP proposes a protocol change so that the committee can define
-different fee rates for maker orders and taker orders.
+different transaction fee rates for maker orders and taker orders.
 
 # Motivation
 
-To improve liquidity of the BitShares DEX.
+To improve liquidity of the BitShares DEX by creating the ability to incentivize market making.
 
 # Rationale
 
--- a/bsip-0085.md
+++ b/bsip-0085.md
@@ -1,0 +1,49 @@
+    BSIP: 0085
+    Title: Maker order creation fee discount
+    Author: Abit More <https://github.com/abitmore>
+    Status: Draft
+    Type: Protocol
+    Created: 2019-10-13
+    Discussion: https://github.com/bitshares/bsips/issues/240
+    Worker: TBD
+
+# Abstract
+
+This BSIP proposes a protocol change so that the committee can define
+different fee rates for maker orders and taker orders.
+
+# Motivation
+
+To improve liquidity of the BitShares DEX.
+
+# Rationale
+
+The maker-taker fee model is adopted widely in centralized exchanges and
+helped them to attract trading activities.
+
+In the BitShares DEX, for each filled order, the owner of the order need to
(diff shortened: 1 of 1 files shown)


bsip-0085.md
```diff
@@ -0,0 +1,49 @@
...
+
+When a limit order with a non-zero pending fee is filled or partially filled,
+* if the order is a take order, process the pending fee as before;
+* if the order is a maker order, return
+`pending_fee * maker_fee_discount_percent` (round down) to the owner,
+then process the remaining pending fee as before.
+ Note: similar to [BSIP 26](
+https://github.com/bitshares/bsips/blob/master/bsip-0026.md), in case when
+ the order creation fee was paid in an asset which was not BTS,
+ return an amount of BTS to the asset's fee pool and
```
MichelSantos: Is this from the `deferred_fee` field?
What is the amount?

bsip-0085.md
```diff
@@ -0,0 +1,49 @@
...
+ return an amount of that asset to the order owner.
```
MichelSantos: Is this from the `deferred_paid_fee` field?
How much is the amount?

pmconrad:
Is this worth it? `limit_order_create` fee is currently 0.0253 BTS (for non-lifetime members), which is equivalent to less than 0.0001 USD. Are makers seriously complaining about this?

I think BSIP-61 is a better option because it also requires less processing.

bsip-0085.md
abitmore: I didn't want to add the details into the specification. It's similar to BSIP26. If you think it's necessary, perhaps create a PR? Thanks.

abitmore:
> limit_order_create fee is currently 0.0253 BTS

We are seeking for significantly increasing the taker fee while keeping the maker fee low enough.

abitmore:
> I think BSIP-61 is a better option

They can both exist at the same time.

pmconrad:
> We are seeking

Who is "we"?

abitmore:
> Who is "we"?

E.G. @froooze

bsip-0085.md
MichelSantos: It isn't clear to me which is why I asked.  Therefore I cannot make a PR.  Without the clarity it will be left up to the code reviewer to interpret whether the new code is similar enough to BSIP26.

bsip-0085.md
MichelSantos: This is left to the code reviewer to interpret the similarity to BSIP26

bsip-0085.md
abitmore: I will update.

abitmore: Updated, please review.

MichelSantos: I greatly appreciate the elaboration @abitmore .  The update is clear and helpful.

abitmore:
@sschiessl-bcp please check this when you have time, I think it can be merged. Thanks.

bsip-0085.md
```diff
@@ -0,0 +1,61 @@
...
+Valid range of that parameter is `[0, 100%]`.
+
+When a limit order is filled or partially filled for the first time,
+technically, when its `deferred_fee` field is non-zero,
+* if the order is a take order, process the deferred fee as before;
+* if the order is a maker order, 
+  * if the order creation fee was paid in BTS, return
+  `round_down(deferred_fee * maker_fee_discount_percent)` to the owner,
+  then process the remaining deferred fee as before;
+  * if the order creation fee was paid in another asset, return
```
sschiessl-bcp: Can you please elaborate why this has two parts? (deferred paid fee and deferred fee)? @abitmore

abitmore: It's explained in [BSIP 26](https://github.com/bitshares/bsips/blob/master/bsip-0026.md).

sschiessl-bcp: Thx

Issue #194, opened 2019-08-17
BSIP 86: Share market fees to the network


abitmore:
BSIP: 0086
    Title: Share market fees to the network
    Author: Abit More <https://github.com/abitmore>
    Status: Draft
    Type: Protocol
    Created: 2019-11-29
    Discussion: https://github.com/bitshares/bsips/issues/194
    Worker: TBD

# Abstract

This BSIP proposes a protocol change: when every trade happens, if there are
market fees generated, a portion (E.G. 5% or 10%) of the market fees goes to
the network.

# Motivation

The network needs more tools to generate income to support its development.

# Rationale

A major activity in the network is trading assets. The asset owners make
profits in the form of market trading fees. Currently, the main tools for the
network to generate income from asset trading activities are order creation
fee and order cancellation fee. It seems reasonable that the asset owners
share some profits (a part of market fees) to the network to support its
development.

The cut of market fees can go to committee-account's vesting balances.
As a supporting measure, committee-account should be exempted from
white-listing restrictions, so that it is able to claim the vesting balances
and sell them for core token or other tokens to pay for development later.

# Specifications

## Protocol upgrade

A time will need to be scheduled for applying the change. In this document,
terms "before the protocol upgrade", "at the protocol upgrade" and "after
the protocol upgrade" may or may not be used to indicate things happen before
the scheduled time, at the scheduled time and after the scheduled time.

## New global parameter

Add a new global parameter `market_fee_network_percent` which can be updated
by the committee only after the protocol upgrade.
Initial value of that parameter is `0`.
Valid range of that parameter is `[0, 100%]`.

## When processing market fees

After the protocol upgrade, when splitting a non-zero market fee, firstly
split `amount * market_fee_network_percent` (round down) to committee-account,
then process the remainder as before. The amount split to committee-account
should go to the vesting balance object whose type is `market_fee_sharing`.

## When checking asset authorities

After the protocol upgrade, when checking authorities (E.G. white-lists) of an
asset on an account, if the account is committee-account, let it pass.

# Discussion

* Other than [the "coin-days as market fees" proposal](
 https://github.com/bitshares/bsips/issues/191) which applies
 a discount to market fees to benefit core token holders directly, this BSIP
 seeks for a mechanism to increase the network's income and benefit core token
 holders indirectly.

* Adjusting the fee schedule and network percent in the referral program is
 another option to increase the network's income, and is probably sufficient
 for funding development.

* In the original design, the purpose of market fees is to reward asset
[...]

PR #253, opened 2019-12-03
BSIP 86: Share market fees to the network

abitmore:
PR for #194.

### DIFF
--- a/bsip-0086.md
+++ b/bsip-0086.md
@@ -1,4 +1,4 @@
-    BSIP: TBD
+    BSIP: 0086
     Title: Share market fees to the network
     Author: Abit More <https://github.com/abitmore>
     Status: Draft
--- a/bsip-0086.md
+++ b/bsip-0086.md
@@ -10,8 +10,8 @@
 # Abstract
 
 This BSIP proposes a protocol change: when every trade happens, if there are
-market fees generated, a portion (E.G. 5% or 10%) of the market fees goes to
-the network.
+market fees generated (E.G. 0.2% of trading volume), a portion of the market
+fees (E.G. 0.01% of trading volume) goes to the network.
 
 # Motivation
 
@@ -104,10 +104,10 @@ asset on an account, if the account is committee-account, let it pass.
 
 # Non-Technical Summary
 
-This BSIP adds a tool for the committee to impose a tax on market trading
+This BSIP adds a tool for the committee to impose a fee on market trading
 fees for all assets which enabled market trading fee, thus would potentially
 increase income for the network to support its development. It may probably
-damage businesses relying on market fees if the committee-decided tax is high.
+damage businesses relying on market fees if the committee-decided fee is high.
 The committee may need some efforts to manage the income.
 
 # Copyright
--- a/bsip-0086.md
+++ b/bsip-0086.md
@@ -83,8 +83,8 @@ asset on an account, if the account is committee-account, let it pass.
  to get involved in the parameter governance process. The percentage can be
  `0` if it is more appropriate.
 
-* Asset owners can specify a zero market fee percentage to get around the
- "tax", while still profiting by setting a higher deposit/withdrawal fee.
+* Asset owners can specify a zero market fee percentage to get around the fee
+ sharing, while still profiting by setting a higher deposit/withdrawal fee.
  It is uncommon nowadays probably because it is less convenient or attractive
  for their customers.
 
--- a/bsip-0086.md
+++ b/bsip-0086.md
@@ -81,11 +81,11 @@ asset on an account, if the account is committee-account, let it pass.
 
 * This would be another motivation for businesses building on top of BitShares
  to get involved in the parameter governance process. The percentage can be
- `0` if it's more appropriate.
+ `0` if it is more appropriate.
 
 * Asset owners can specify a zero market fee percentage to get around the
  "tax", while still profiting by setting a higher deposit/withdrawal fee.
- It is uncommon nowadays probably because it's less convenient or attractive
+ It is uncommon nowadays probably because it is less convenient or attractive
  for their customers.
 
 * Another strategy is to charge a network market fee globally, independent of
@@ -93,10 +93,14 @@ asset on an account, if the account is committee-account, let it pass.
  a fee from asset owners, which kills some freedom and would hinder non-profit
  use of the platform.
 
-* It's possible that committe-account would accumulate a lot of tokens with no
+* It is possible that committee-account would accumulate a lot of tokens with no
  value and tokens which are unable or hard to be used to pay for development.
- It's a side effect. Hopefully there will be quite some usable tokens.
+ It is a side effect. Hopefully there will be quite some usable tokens.
  The committee will have to do some work to manage the tokens anyway.
+
+* The committee will be accumulating tokens which could potentially be used or
+ not used for any purpose. In other words, this BSIP as currently written does
+ not stipulate how those tokens will or will not be used by the committee.
 
 # Non-Technical Summary
 
--- a/bsip-0086.md
+++ b/bsip-0086.md
@@ -2,16 +2,18 @@
     Title: Share market fees to the network
     Author: Abit More <https://github.com/abitmore>
     Status: Draft
-    Type: Protocol
+    Type: Accepted
     Created: 2019-11-29
     Discussion: https://github.com/bitshares/bsips/issues/194
     Worker: TBD
 
 # Abstract
 
-This BSIP proposes a protocol change: when every trade happens, if there are
-market fees generated (E.G. 0.2% of trading volume), a portion of the market
-fees (E.G. 0.01% of trading volume) goes to the network.
+This BSIP proposes a protocol change: Every time a trade is executed (limit order is filled) and
+if there are market fees generated, a portion of the market fees goes to the network.
+
+Example: The network percent of market fees is set to 50%, and 0.2% of trading volume is the set as market fee. 
+Then 0.1% of trading volume goes to the network.
 
 # Motivation
 
@@ -44,8 +46,8 @@ the scheduled time, at the scheduled time and after the scheduled time.
 
 Add a new global parameter `market_fee_network_percent` which can be updated
 by the committee only after the protocol upgrade.
-Initial value of that parameter is `0`.
-Valid range of that parameter is `[0, 100%]`.
+Initial value of that parameter is `0%`.
+Valid range of that parameter is `[0%, 100%]`.
 
 ## When processing market fees
 
@@ -105,10 +107,10 @@ asset on an account, if the account is committee-account, let it pass.
 # Non-Technical Summary
 
 This BSIP adds a tool for the committee to impose a fee on market trading
-fees for all assets which enabled market trading fee, thus would potentially
-increase income for the network to support its development. It may probably
-damage businesses relying on market fees if the committee-decided fee is high.
-The committee may need some efforts to manage the income.
+fees for all assets which enabled market trading fee, thus it can increase income 
+for the network to support its development. It may 
+hinder businesses relying on market fees if the committee-decided fee is high.
+The committee may need some efforts to decide on an appropriate value of the parameter and manage the income.
 
 # Copyright
 
--- a/bsip-new.md
+++ b/bsip-new.md
@@ -1,0 +1,115 @@
+    BSIP: TBD
+    Title: Share market fees to the network
+    Author: Abit More <https://github.com/abitmore>
+    Status: Draft
+    Type: Protocol
+    Created: 2019-11-29
+    Discussion: https://github.com/bitshares/bsips/issues/194
+    Worker: TBD
+
+# Abstract
+
+This BSIP proposes a protocol change: when every trade happens, if there are
+market fees generated, a portion (E.G. 5% or 10%) of the market fees goes to
+the network.
+
+# Motivation
+
+The network needs more tools to generate income to support its development.
+
+# Rationale
+
+A major activity in the network is trading assets. The asset owners make
+profits in the form of market trading fees. Currently, the main tools for the
+network to generate income from asset trading activities are order creation
+fee and order cancellation fee. It seems reasonable that the asset owners
+share some profits (a part of market fees) to the network to support its
+development.
+
+The cut of market fees can go to committee-account's vesting balances.
+As a supporting measure, committee-account should be exempted from
+white-listing restrictions, so that it is able to claim the vesting balances
+and sell them for core token or other tokens to pay for development later.
+
+# Specifications
+
+## Protocol upgrade
+
+A time will need to be scheduled for applying the change. In this document,
+terms "before the protocol upgrade", "at the protocol upgrade" and "after
+the protocol upgrade" may or may not be used to indicate things happen before
+the scheduled time, at the scheduled time and after the scheduled time.
+
+## New global parameter
+
+Add a new global parameter `market_fee_network_percent` which can be updated
+by the committee only after the protocol upgrade.
+Initial value of that parameter is `0`.
+Valid range of that parameter is `[0, 100%]`.
+
+## When processing market fees
+
+After the protocol upgrade, when splitting a non-zero market fee, firstly
+split `amount * market_fee_network_percent` (round down) to committee-account,
+then process the remainder as before. The amount split to committee-account
+should go to the vesting balance object whose type is `market_fee_sharing`.
+
+## When checking asset authorities
+
+After the protocol upgrade, when checking authorities (E.G. white-lists) of an
+asset on an account, if the account is committee-account, let it pass.
+
+# Discussion
+
+* Other than [the "coin-days as market fees" proposal](
+ https://github.com/bitshares/bsips/issues/191) which applies
+ a discount to market fees to benefit core token holders directly, this BSIP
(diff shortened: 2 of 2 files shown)


bsip-0086.md
```diff
@@ -0,0 +1,115 @@
...
+white-listing restrictions, so that it is able to claim the vesting balances
+and sell them for core token or other tokens to pay for development later.
+
+# Specifications
+
+## Protocol upgrade
+
+A time will need to be scheduled for applying the change. In this document,
+terms "before the protocol upgrade", "at the protocol upgrade" and "after
+the protocol upgrade" may or may not be used to indicate things happen before
```
MichelSantos: > may or may not be

I don't see how "may not" is helpful for specifications.  Might it be better to change this phrase to "will be"?

MichelSantos:
The discussion is very good

My personal suggestion, which should be considered completely optional, is to consider adding a discussion point that says that the committee will be accumulating tokens which could potentially be used or not used for any purpose.  In other words, this BSIP as currently written does not stipulate how those tokens will or will not be used by the committee.

bsip-0086.md
```diff
@@ -0,0 +1,115 @@
...
+ issuers for their work (e.g. gateway providers, or businesses built around
+ PMs), while operation fees (`limit_order_create/cancel` in particular) reward
+ the network for performing its work. The separation of market/operation fees
+ matches the separation of roles. Changing this may damage existing businesses.
+ Thus this BSIP could be a bad idea from this perspective. Also note that the
+ network can profit from market fees by setting them on committee-owned tokens.
+
+* This would be another motivation for businesses building on top of BitShares
+ to get involved in the parameter governance process. The percentage can be
+ `0` if it's more appropriate.
```
MichelSantos: > "it's"
There are 4 usages of "it's" which you might consider changing to "it is"

bsip-0086.md
```diff
@@ -0,0 +1,115 @@
...
+ "tax", while still profiting by setting a higher deposit/withdrawal fee.
+ It is uncommon nowadays probably because it's less convenient or attractive
+ for their customers.
+
+* Another strategy is to charge a network market fee globally, independent of
+ what the asset owner decides. It essentially removes the option to not set
+ a fee from asset owners, which kills some freedom and would hinder non-profit
+ use of the platform.
+
+* It's possible that committe-account would accumulate a lot of tokens with no
```
MichelSantos: > "committe-account"

Typo: committee-account

bsip-0086.md
sschiessl-bcp: Then also check that the wording is consistently used.

bsip-0086.md
```diff
@@ -0,0 +1,115 @@
...
+profits in the form of market trading fees. Currently, the main tools for the
+network to generate income from asset trading activities are order creation
+fee and order cancellation fee. It seems reasonable that the asset owners
+share some profits (a part of market fees) to the network to support its
+development.
+
+The cut of market fees can go to committee-account's vesting balances.
+As a supporting measure, committee-account should be exempted from
+white-listing restrictions, so that it is able to claim the vesting balances
+and sell them for core token or other tokens to pay for development later.
```
sschiessl-bcp: I think it would make sense to add the following:

"The committee is instructed to regularly (at least once every quarter) convert incoming market fees into BitShares native assets: BTS (favorable) or any BitAsset."

bsip-0086.md
```diff
@@ -0,0 +1,115 @@
...
+terms "before the protocol upgrade", "at the protocol upgrade" and "after
+the protocol upgrade" may or may not be used to indicate things happen before
+the scheduled time, at the scheduled time and after the scheduled time.
+
+## New global parameter
+
+Add a new global parameter `market_fee_network_percent` which can be updated
+by the committee only after the protocol upgrade.
+Initial value of that parameter is `0`.
+Valid range of that parameter is `[0, 100%]`.
```
sschiessl-bcp: Here is a thought: Does it make sense to couple this into strengthening the referral rewards program.

What I mean: Whatever comes in as `market_fee_network_percent` gets funneled into the referral rewards program, and is split according to the `network_fee` (with current values that means that 20% goes to committee, and 80% into referral rewards.

Downside is that this also potentially clutters account's balances with useless assets, but it would give a massive boost to the rewards program.

bsip-0086.md
abitmore: It's for easier to copy & paste.

bsip-0086.md
abitmore: I don't think it's good to say that the committee should do this or that in the document.

bsip-0086.md
abitmore: Another side effect is the referrers may get tokens they can't use due to white-lists and etc, it could also be a headache for asset owners. Although the idea is interesting, I personally prefer not adding it to this BSIP (maybe create a new BSIP).

bsip-0086.md
sschiessl-bcp: I don't think it's good to give an entity a role if that entity does not accept those responsibilities explicitly. Then we should think of a new entity to be the receiver.

bsip-0086.md
sschiessl-bcp: Allow asset onwers to turn off rewards funneling? (then all goes to committee)?

bsip-0086.md
abitmore: ?

bsip-0086.md
abitmore: Sounds complicated.

bsip-0086.md
abitmore: IMO the scope of a BSIP should be as small as needed. It's hard to define how to efficiently use the potential funds right now. Selling the funds for BTS or bitAsset is not the only way to use the funds. For example, it could be easier for the committee to directly pay gateway.BTC for development. I'd rather let the committee decide when they see fit.

sschiessl-bcp: > IMO the scope of a BSIP should be as small as needed. It's hard to define how to efficiently use the potential funds right now. Selling the funds for BTS or bitAsset is not the only way to use the funds. For example, it could be easier for the committee to directly pay gateway.BTC for development. I'd rather let the committee decide when they see fit.

Considering the debacle with CryptoBridge holding any value in native assets should be preferrable. I see the point that a BSIP can't and shouldn't tell committee what to do. The very least ALL current active committee members should confirm their willingness for taking on that responsibility. This might be considered unfair, with the information available to me my personal trust regarding committee towards workers and transparent payments is not very high at the moment, considering what has happened in the past with worker proposals governed by committee.

bsip-0086.md
sschiessl-bcp: Ok, I guess if asset owners what to take part in rewards program to boost their adoption they can set it themselves.

bsip-0086.md
sschiessl-bcp: Another thought: 
How about the following

a) let the committee decide who will be the `market_fee_network_budget_account` of the incoming budget via committee parameter (value is an account name). This may be very well just the committee-account in the beginning.
b) if desirable, allow a fee for the one that handles the budget. To do that, define additionally a `market_fee_network_budget_handler` (value is an account name) and a `market_fee_network_budget_handler_fee` (value is percent). Then the incoming budget goes `1-market_fee_network_budget_handler_fee`% to the `market_fee_network_budget_account` and `market_fee_network_budget_handler_fee`% to the  `market_fee_network_budget_handler`. The handler should have access to the `market_fee_network_budget_account` account of course, but this is external configuration of account permissions and not a core requirement (would e.g. allow use of BSIP 40 permissions)

sschiessl-bcp: @bitcrab see here

bsip-0086.md
MichelSantos: Although I think it would be better to remove the "or may not", I think that the idea is clear enough in its current form.

bsip-0086.md
abitmore: "market_fee_network_budget_handler" complicates things related to white-listing.

sschiessl-bcp: > "market_fee_network_budget_handler" complicates things related to white-listing. IMO it's unnecessary.

Not more than if the asset owner would use market fee sharing, same logic.

bitcrab: > "market_fee_network_budget_handler" complicates things related to white-listing. IMO it's unnecessary.

agree, it bring more complex unnecessarily.

MichelSantos:
I conclude that the proposal is technically complete

bsip-0086.md
```diff
@@ -0,0 +1,119 @@
...
+    Status: Draft
+    Type: Protocol
+    Created: 2019-11-29
+    Discussion: https://github.com/bitshares/bsips/issues/194
+    Worker: TBD
+
+# Abstract
+
+This BSIP proposes a protocol change: when every trade happens, if there are
+market fees generated, a portion (E.G. 5% or 10%) of the market fees goes to
```
dls-cipher: e.g. 5% or 10% from 0.1%

bsip-0086.md
```diff
@@ -0,0 +1,119 @@
...
+terms "before the protocol upgrade", "at the protocol upgrade" and "after
+the protocol upgrade" may or may not be used to indicate things happen before
+the scheduled time, at the scheduled time and after the scheduled time.
+
+## New global parameter
+
+Add a new global parameter `market_fee_network_percent` which can be updated
+by the committee only after the protocol upgrade.
+Initial value of that parameter is `0`.
+Valid range of that parameter is `[0, 100%]`.
```
dls-cipher: Parameter should be `[0, 25%]`

a) we cant force against private business or even be able to take their entire income.
b) existing 3rd parties and new on-boarding businesses/asset owners will feel more safe knowing that maximum cut they can pay is 25% from their market fee.

bsip-0086.md
```diff
@@ -0,0 +1,119 @@
...
+ matches the separation of roles. Changing this may damage existing businesses.
+ Thus this BSIP could be a bad idea from this perspective. Also note that the
+ network can profit from market fees by setting them on committee-owned tokens.
+
+* This would be another motivation for businesses building on top of BitShares
+ to get involved in the parameter governance process. The percentage can be
+ `0` if it is more appropriate.
+
+* Asset owners can specify a zero market fee percentage to get around the
+ "tax", while still profiting by setting a higher deposit/withdrawal fee.
```
dls-cipher: replace "tax" with "newly introduced network market fee for asset owners".

a) blockchain lose legally freedom as entire MIT and free concept tech the moment when use the word tax.
b) to be able to "tax" blockchain must have an owner (business registered entity)

Solution: Try to replace tax with something as `network default market fee`

bsip-0086.md
```diff
@@ -0,0 +1,119 @@
...
+ It is a side effect. Hopefully there will be quite some usable tokens.
+ The committee will have to do some work to manage the tokens anyway.
+
+* The committee will be accumulating tokens which could potentially be used or
+ not used for any purpose. In other words, this BSIP as currently written does
+ not stipulate how those tokens will or will not be used by the committee.
+
+# Non-Technical Summary
+
+This BSIP adds a tool for the committee to impose a tax on market trading
```
dls-cipher: See [here](https://github.com/bitshares/bsips/pull/253/files#r360287036) for entire paragraph where contains word "tax".

bsip-0086.md
abitmore: The whole sentence looks strange after the change. Any better idea?

bsip-0086.md
abitmore: I'd like to hear others' opinions. @sschiessl-bcp @MichelSantos @bitcrab please?

bsip-0086.md
abitmore: I need help here. Too hard for me to reword this paragraph :(

bsip-0086.md
abitmore: Updated to "get around fee sharing".

bsip-0086.md
abitmore: > Not more than if the asset owner would use market fee sharing, same logic. Or do I miss something?

If an asset owner would use market fee sharing, she will consider the matters about whitelisting. The situation here is different, here the committee is doing something but not the asset owners. By using the committee-account directly, we only need to exclude that one account from whitelisting restrictions; by using a dynamic account we need more complicated logic.

bsip-0086.md
MichelSantos: Well since you asked I added [my opinion](https://github.com/bitshares/bsips/issues/194#issuecomment-568225147) into the issue to avoid its getting lost

bsip-0086.md
MichelSantos: How about this?

"This BSIP adds a tool for the committee to impose the sharing of market trading fees for every asset that has enabled a market trading fee. ..."

bsip-0086.md
MichelSantos: How about

"This BSIP proposes a protocol change: a portion (e.g. 5% or 10%) of every market trading fee (e.g. 0.1% of a filled trade) shall be shared with the network."

bsip-0086.md
bitcrab: I agree to limit the parameter to ` [0, 25%]`, 25% is high enough.

sschiessl-bcp: Does the network percentage of transaction fees have any limit?

The one thing that came to mind is that a fixed upper limit might gateways to do worst case planning, instead of opening communication with committee to agree on feasible values.

bsip-0086.md
sschiessl-bcp: "This BSIP proposes that a portion of every market trading fee shall be shared with the network."

Then add an example at the end of this paragraph.

bsip-0086.md
bitcrab: @sschiessl-bcp what kind of worst case planning? upper is just to tell public that system at most can charge so much on market fees.

As GDEX founder to me a 5%-10% market fee sharing to system is totally acceptable, even 25% is discussible，not aware of other gateways' opinions.

sschiessl-bcp: With worst case planning I mean: If the fee sharing limit is set to 20%, then gateways could simply adjust the fee such that the remaining 80% reflect the actually intended fee, and if the actual fee sharing % is set lower than the limit, gateways are simply making a bonus. In other words: The limit would be to the downside of the user.

This is certainly a thoughtplay, but our gateways have shown worst-case behavior lately.

bsip-0086.md
sschiessl-bcp: True, complexity would be increased. Still, a strong referral rewards program is exactly what other exchanges have that makes them grow massively.

sschiessl-bcp: Okay, if the committee is accepting this role pro-bono ...

bsip-0086.md
sschiessl-bcp: Just one last I want to add: A strong referral rewards program is exactly what other exchanges have that makes them grow massively. So we are putting the success in the hands of the gateways. Just FYI.

bsip-0086.md
bitcrab: this is just as the tax logic, the tax paid by service provider to goverment always come from the service buyer. no problem.

abitmore:
@sschiessl-bcp: yes, I think so.

sschiessl-bcp:
Status set to accepted. Merge at your own convenience

shulthz:
I suggest to pause this BSIP until solved the legal risk.

This BSIP should make a statement for this kind lawsuit.

>   This BSIP86 has legal risk, i suggest everyone need to consider it very carefully and seek a legal consultancy.
> 
>     When BSIP86 charges the market fees from the GATEWAY.assets, the BTS system will have joint liability and must be responsible for these GATEWAY.assets, but these GATEWAY.assets were without supervision and control by the BTS system.
> 
>     Gateway just likes the DApp of BTS, BTS just provides the service and takes the transaction fee of blockchain，DApp do what is not the business of BTS, but when BTS charges the market fees of GATEWAY.assets, the nature of things will become different.
>

sschiessl-bcp:
> I suggest to pause this BSIP until solved the legal risk.
> 
> This BSIP must make a statement for this kind lawsuit，how to deal? who will be responsible for these lawsuit?
> This BSIP must make a statement for this kind lawsuit to BTS holders.
> 
> This BSIP86 has legal risk, i suggest everyone need to consider it very carefully and seek a legal consultancy.
> 
> When BSIP86 charges the market fees from the GATEWAY.assets, the BTS system will have joint liability and must be responsible for these GATEWAY.assets, but these GATEWAY.assets were without supervision and control by the BTS system.
> 
> Gateway just likes the DApp of BTS, BTS just provides the service and takes the transaction fee of blockchain，DApp do what is not the business of BTS, but when BTS charges the market fees of GATEWAY.assets, the nature of things will become different.

The BSIP creates the technical framework. Without any active action, nothing will be activated. That will need to be done by committee setting the fee. Please direct your efforts towards committee members and make them aware.

shulthz:
>The BSIP creates the technical framework. Without any active action, nothing will be activated. That will need to be done by committee setting the fee. Please direct your efforts towards committee members and make them aware.

Yes, it is a technical framework, but it didn't indicate it has legal risk clearly in the MD, it makes a trap to bts holders and committee.

Committees have staff changes, at last, the bts holders will face the lawsuit.

Even it is a technical framework, but it must make a statement for so clearly legal risk to bts holders, let the bts holders clearly know what they will face in the future.

PR #257, opened 2019-12-30
1000:1 Reverse Split of Core BTS Token

litepresence:
The motivation of this split is to increase share price immediately to prevent delisting risk, encourage listing acceptance at new exchanges, and more generally encourage higher trade volume at centralized exchanges by removing the delisting threat.

### DIFF
--- a/bsip-0086.md
+++ b/bsip-0086.md
@@ -1,0 +1,82 @@
+    BSIP: BSIP 86
+    Title: 1000:1 Reverse Split of Core BTS Token
+    Authors: litepresence finitestate@tutamail.com
+    Status: Draft
+    Type: Protocol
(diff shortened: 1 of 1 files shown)


abitmore:
Just FYI, BSIP number 86 has already been assigned to https://github.com/bitshares/bsips/pull/253 (Share market fees to the network).

sschiessl-bcp:
This would require a massive effort to synchronize that all software is updated to reflect the new precision, or face immediate exploitation by clever traders / bots / other systems. I imagine that many integrations (also from CEXes) have hardcoded precision for assets, thus they all need to change fundamental parameters.

While this might have positive effect simply by being mentioned everywhere, I don't see that this effort can be pulled off without a budget.

PR #265, opened 2020-02-26
Create bsip-0087.md

bitcrab:
create a new PR for BSIP87 to replace the old messy one.

### DIFF
--- a/bsip-0087.md
+++ b/bsip-0087.md
@@ -1,0 +1,36 @@
+    BSIP: 0087
+    Title: Force Settlement Fee Ratio
+    Authors:
+      Jerry Liu bitcrab@qq.com
+    Status: Draft
+    Type: Protocol
+    Created: 2020-02-23
+    Discussion: https://github.com/bitshares/bsips/issues/260
+    Worker: TBD
+
+# Abstract
+This BSIP provide a solution to charge fee from force settlement.
+
+# Motivation
+Force settlement is an important part in smartcoin design, it provides the power to smartcoin holders to ask for collaterals with reference to feed price, and then guarantees the peg of smartcoin.
+
+The community continuously optimize the rules of smartcoin to balance the interests of relevant parties, taking bitCNY as an example, at the beginning the rules focus more on smart coin peg but care less on debt position owners’ interests, change are done to give debt position owners more protection, the force settlement offset has been changed from 0 to 2%, the max settlement  volume per maintenance period has been changed from no limit to 0.5% of smartcoin supply.  
+
+In the past several weeks, force settlement happened more intensively than before, the background is that the crypto market seemly switched from bear to bull, and BTS price went above 0.22CNY - which is the voted threshold per BSIP76, at some time the frequently happened force settlements made the lowest debt position collateral ratio even higher than 3, this phenomenon triggered warm discussion, almost all the users agree that the settler need to pay more cost while executing force settlement, and the cost will be paid to system as fee.
+
+# Rationale
+Force settlement fee can be seen as to pay part of the cost for smartcoin supply, stabilization and also the liquidation of collaterals, it is irrelevant to market fee sharing.
+To ensure the debt positions can always be closed with suffcient smartcoin, it is more feasible to cut off part of the traded collaterals as fee, instead of smartcoin.
+
+# Specification
+Add one new parameter Force Settlement Fee Ratio(FSFR) for each smartcoin, which is controlled by the smartcoin owner.
+`Force settlement order price = settlement price*(1+FSO)/(1-FSFR)`
+
+Here settlement price = feed price.
+
+FSO is Force Settlement Offset.
+
+When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collaterals in quantity `X*(1-FSFR)/[settlement price*(1+FSO)]`, the settled debt position owner sells collaterals in quantity `X/[settlement price*(1+FSO)]` and get smartcoin in quantity `X`, the delta in paid and received collaterals in quantity `X*FSFR/[settlement price*(1+FSO)]` will be paid to the owner of the smartcoin as force settlement fee.
+
+# Copyright
+This document is placed in the public domain.
--- a/bsip-0087.md
+++ b/bsip-0087.md
@@ -33,9 +33,9 @@ Add one new parameter Force Settlement Fee Percentage(FSFP) for each smartcoin, 
 
 FSO is Force Settlement Offset of the smartcoin.
 
-When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collaterals in quantity `X*(1-FSO)*(1-FSFP)/feed price price*`, the settled debt position owner sells collaterals in quantity `X*(1-FSO)/feed price*` and get smartcoin in quantity `X`, the delta in paid and received collaterals in quantity `X*FSFP*(1-FSO)/feed price*` will be paid to the owner of the smartcoin as force settlement fee.
+When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collateral in quantity `X*(1-FSO)*(1-FSFP)/feed price price*`, the settled debt position owner sells collateral in quantity `X*(1-FSO)/feed price*` and get smartcoin in quantity `X`, the delta in paid and received collateral in quantity `X*FSFP*(1-FSO)/feed price*` will be paid to the owner of the smartcoin as force settlement fee.
 
-In this scenario, `force settlement order price = feed price/(1-FSO)` is the price to the settled debt position owners(maker). and the force settlement fee is calculated out based on this order price and FSFP.
+In this scenario, `force settlement order price = feed price/(1-FSO)` is the price to the settled debt position owners(maker). and the force settlement fee is paid in collateral and the amount is calculated out based on this order price and FSFP.
 
 # Copyright
 This document is placed in the public domain.
--- a/bsip-0087.md
+++ b/bsip-0087.md
@@ -27,7 +27,7 @@ Add one new parameter Force Settlement Fee Percentage(FSFP) for each smartcoin, 
 
 FSO is Force Settlement Offset.
 
-When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collaterals in quantity `X*(1-FSFP)/[settlement price*(1+FSO)]`, the settled debt position owner sells collaterals in quantity `X/[settlement price*(1+FSO)]` and get smartcoin in quantity `X`, the delta in paid and received collaterals in quantity `X*FSFP/[settlement price*(1+FSO)]` will be paid to the owner of the smartcoin as force settlement fee.
+When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collaterals in quantity `X*(1-FSFP)/[feed price price*(1+FSO)]`, the settled debt position owner sells collaterals in quantity `X/[feed price*(1+FSO)]` and get smartcoin in quantity `X`, the delta in paid and received collaterals in quantity `X*FSFP/[feed price*(1+FSO)]` will be paid to the owner of the smartcoin as force settlement fee.
 
 In this scenario, `force settlement order price = feed price*(1+FSO)` is the price to the settled debt position owners(maker). 
 
--- a/bsip-0087.md
+++ b/bsip-0087.md
@@ -35,7 +35,7 @@ FSO is Force Settlement Offset of the smartcoin.
 
 When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collateral in quantity `X*(1-FSO)*(1-FSFP)/feed price price*`, the settled debt position owner sells collateral in quantity `X*(1-FSO)/feed price*` and get smartcoin in quantity `X`, the delta in paid and received collateral in quantity `X*FSFP*(1-FSO)/feed price*` will be paid to the owner of the smartcoin as force settlement fee.
 
-In this scenario, `force settlement order price = feed price/(1-FSO)` is the price to the settled debt position owners(maker). and the force settlement fee is paid in collateral and the amount is calculated out based on this order price and FSFP.
+In this scenario, as a price offset parameter, FSO is irrelevant to fee charging, it just define `force settlement price = feed price/(1-FSO)`. and the force settlement fee is paid in collateral and the amount is calculated out based on `force settlement price` and `FSFP`.
 
 # Copyright
 This document is placed in the public domain.
--- a/bsip-0087.md
+++ b/bsip-0087.md
@@ -1,5 +1,5 @@
     BSIP: 0087
-    Title: Force Settlement Fee Ratio
+    Title: Force Settlement Fee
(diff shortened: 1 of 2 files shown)


abitmore:
Thanks.

bsip-0087.md
```diff
@@ -0,0 +1,36 @@
...
+    Type: Protocol
+    Created: 2020-02-23
+    Discussion: https://github.com/bitshares/bsips/issues/260
+    Worker: TBD
+
+# Abstract
+This BSIP provide a solution to charge fee from force settlement.
+
+# Motivation
+Force settlement is an important part in smartcoin design, it provides the power to smartcoin holders to ask for collaterals with reference to feed price, and then guarantees the peg of smartcoin.
```
sschiessl-bcp: Force settlement can't guarantee anything. It is merely a tool that may be used in a away that may strengthen the peg.

bsip-0087.md
```diff
@@ -0,0 +1,36 @@
...
+
+In the past several weeks, force settlement happened more intensively than before, the background is that the crypto market seemly switched from bear to bull, and BTS price went above 0.22CNY - which is the voted threshold per BSIP76, at some time the frequently happened force settlements made the lowest debt position collateral ratio even higher than 3, this phenomenon triggered warm discussion, almost all the users agree that the settler need to pay more cost while executing force settlement, and the cost will be paid to system as fee.
+
+# Rationale
+Force settlement fee can be seen as to pay part of the cost for smartcoin supply, stabilization and also the liquidation of collaterals, it is irrelevant to market fee sharing.
+To ensure the debt positions can always be closed with suffcient smartcoin, it is more feasible to cut off part of the traded collaterals as fee, instead of smartcoin.
+
+# Specification
+Add one new parameter Force Settlement Fee Ratio(FSFR) for each smartcoin, which is controlled by the smartcoin owner.
```
sschiessl-bcp: A ratio should be defined as the fraction of two values. Is this new fee a ratio?

bsip-0087.md
```diff
@@ -0,0 +1,36 @@
...
+`Force settlement order price = settlement price*(1+FSO)/(1-FSFR)`
+
+Here settlement price = feed price.
```
sschiessl-bcp: Let's use feed price instead then. This wording is buried within core and just confusing.

bsip-0087.md
```diff
@@ -0,0 +1,36 @@
...
+
+In the past several weeks, force settlement happened more intensively than before, the background is that the crypto market seemly switched from bear to bull, and BTS price went above 0.22CNY - which is the voted threshold per BSIP76, at some time the frequently happened force settlements made the lowest debt position collateral ratio even higher than 3, this phenomenon triggered warm discussion, almost all the users agree that the settler need to pay more cost while executing force settlement, and the cost will be paid to system as fee.
+
+# Rationale
+Force settlement fee can be seen as to pay part of the cost for smartcoin supply, stabilization and also the liquidation of collaterals, it is irrelevant to market fee sharing.
+To ensure the debt positions can always be closed with suffcient smartcoin, it is more feasible to cut off part of the traded collaterals as fee, instead of smartcoin.
+
+# Specification
+Add one new parameter Force Settlement Fee Ratio(FSFR) for each smartcoin, which is controlled by the smartcoin owner.
+`Force settlement order price = settlement price*(1+FSO)/(1-FSFR)`
```
sschiessl-bcp: There is a distinct price for the person doing the settling (taker), and another one for the person being settled (maker). 

The `force settlement order price` as indicated above would then be the one for the taker.

bsip-0087.md
```diff
@@ -0,0 +1,36 @@
...
+
+Here settlement price = feed price.
+
+FSO is Force Settlement Offset.
+
+When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collaterals in quantity `X*(1-FSFR)/[settlement price*(1+FSO)]`, the settled debt position owner sells collaterals in quantity `X/[settlement price*(1+FSO)]` and get smartcoin in quantity `X`, the delta in paid and received collaterals in quantity `X*FSFR/[settlement price*(1+FSO)]` will be paid to the owner of the smartcoin as force settlement fee.
```
sschiessl-bcp: Can we add an example below where you use CNY ?

bsip-0087.md
abitmore: I think it should be a percentage.

bsip-0087.md
abitmore: To be clear, `settlement_price` is the variable name used in core and API, and often be referenced as "feed price".

bsip-0087.md
bitcrab: > Can we add an example below where you use CNY ?

currently bitCNY is mainly used as one trading asset/derivative in DEX, it is very seldom used in scenarios other than this, I don't think we need to add CNY usage example here as this is just a BSIP.

bitcrab:
modified the changed file according to your suggestion, please check again.

bsip-0087.md
shulthz: > 
> 
> > Can we add an example below where you use CNY ?
> 
> currently bitCNY is mainly used as one trading asset/derivative in DEX, it is very seldom used in scenarios other than this, I don't think we need to add CNY usage example here as this is just a BSIP.

This is just like a joke, if bitCNY is mainly used as one trading asset/derivative in DEX even currently, so why bitcny need it?

Why not use bitEUR add an example?

bitcrab: > > > Can we add an example below where you use CNY ?
> > 
> > 
> > currently bitCNY is mainly used as one trading asset/derivative in DEX, it is very seldom used in scenarios other than this, I don't think we need to add CNY usage example here as this is just a BSIP.
> 
> This is just like a joke, if bitCNY is mainly used as one trading asset/derivative in DEX even currently, so why bitcny need it?
> 
> Why not use bitEUR to add an example? and if bitEUR is mainly used as one trading asset/derivative in DEX, so why we need this BSIP?

because Bitshares is now finding ways to get more income to pay necessary payments and buy back BTS. either force settlement or margin call is suitable for being charged.

and because bitEUR is in too low supply and too low trading volume to be a good example, if bitEUR is widely used somewhere outside BTS-DEX, please provide the details.

however, we need this BSIP even if bitEUR is widely used somewhere, this BSIP is to enable smartcoin owners to charge force settlement, whether to charge depend on smartcoin owner's choice.

sschiessl-bcp: It's not about whether the chosen asset is a good cancidate. Call it bitABC if you want. It is about illustrating with actual number to give better understanding.

bsip-0087.md
```diff
@@ -0,0 +1,35 @@
...
+# Rationale
+Force settlement fee can be seen as to pay part of the cost for smartcoin supply, stabilization and also the liquidation of collaterals, it is irrelevant to market fee sharing.
+To ensure the debt positions can always be closed with suffcient smartcoin, it is more feasible to cut off part of the traded collaterals as fee, instead of smartcoin.
+
+# Specification
+Add one new parameter Force Settlement Fee Percentage(FSFP) for each smartcoin, which is controlled by the smartcoin owner.
+
+FSO is Force Settlement Offset.
+
+When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collaterals in quantity `X*(1-FSFP)/[feed price price*(1+FSO)]`, the settled debt position owner sells collaterals in quantity `X/[feed price*(1+FSO)]` and get smartcoin in quantity `X`, the delta in paid and received collaterals in quantity `X*FSFP/[feed price*(1+FSO)]` will be paid to the owner of the smartcoin as force settlement fee.
```
sschiessl-bcp: The current definition of FSFP is confusing imo.

FSO is a percent, and FSFP is a percent. For the sake of clarity of my explanation, let's have both be denominated as values in `[0,1]` where `1` is `100%`. I would suggest to refer them both to the same basis, namely the feed price.

Current feed price is `Z BTS/bitABC`. I want to settle `X bitABC`. Now the formula to calculate the collateral `Y BTS` I receive should follow `Y = X*Z*(1-FSFP-FSO)`, i.e. the percentages are both in reference to `X*Z`. 

Now to coin the wording: The force settlement order price is `Z*(1-FSFP-FSO)` (maker price).
Then the debt position holder sells collateral for `Z*(1-FSO)` (i.e. he gets to keep `Z*FSO` collateral as compensation), and the platform receives `Z*FSFP` as fee.

bsip-0087.md
bitcrab: > It's not about whether the chosen asset is a good cancidate. Call it bitABC if you want. It is about illustrating with actual number to give better understanding.

Thanks, updated the #Motivation part, please review.

bsip-0087.md
bitcrab: @abitmore please advise, I feel the important thing is to keep it compatible to core code and BSIP74, which is already voted up.

sschiessl-bcp:
> @abitmore please advise, I feel the important thing is to keep it compatible to core code and BSIP74, which is already voted up.

`1.14.239 | Poll - BSIP74 - Market Fee Ratiobitcrab` currently has `464,918,785` votes, and to become active `715,885,028` are needed. 

In order to use `1.14.158 | threshold-bsip` as the intended threshold for BSIPs to become active it would need to either be mentioned in the BSIP-74 specifically, or the better option a BSIP needs to be written establishing a set of rules for when BSIPs become active that include `1.14.158` as the threshold. This new BSIP must then be passed with the old rules (become active by receiving part of the daily pay).

If you do not agree with me there, please elaborate why.

bitcrab:
> > @abitmore please advise, I feel the important thing is to keep it compatible to core code and BSIP74, which is already voted up.
> 
> `1.14.239 | Poll - BSIP74 - Market Fee Ratiobitcrab` currently has `464,918,785` votes, and to become active `715,885,028` are needed.
> 
> In order to establish `1.14.158 | threshold-bsip` as the threshold for BSIPs it would need to either be mentioned in the BSIP-74 specifically, or the better option a BSIP needs to be written establishing a set of rules for when BSIPs become active that include `1.14.158` as the threshold. This new BSIP must then be passed with the old rules (become active by receiving part of the daily pay).
> 
> If you do not agree with me there, please elaborate why.

the poll BSIP worker voting is just to ensure that the community agree to implement it, without considering budget.

1.14.158 is effective for all the BSIPs.

after that, one more voting for budget is needed, one BSIP can be included in a big budget worker or as a separate budget worker to accept approval, either is OK, IIRC we have experienced both.

I hope at this moment we can firstly focus on making the BSIP document ready and then discuss how to vote. 

Chinese can build one hospital in 10 days, I don't think it's good to cost more than 10 days to prepare one document.

sschiessl-bcp:
> the poll BSIP worker voting is just to ensure that the community agree to implement it, without considering budget.
> 
> 1.14.158 is effective for all the BSIPs.

I understand the intent. But this BSIP has not been approved as the new BSIP threshold by BTS holders, not for BSIP-74 nor for all BSIPs. It has only been created by committee. Should that be considered enough? And if so, why?

sschiessl-bcp:
> the poll BSIP worker voting is just to ensure that the community agree to implement it, without considering budget.

I have opened a new issue for this, let's keep this one on topic
https://github.com/bitshares/bsips/issues/267

bsip-0087.md
abitmore: I still think it's a UI thing. `Y=X*Z*(1-FSFP)*(1-FSO)` is as clear as `Y=X*Z*(1-FSFP-FSO)`, and the former is even clearer when implementing in the core.

sschiessl-bcp: With `Y=X*Z*(1-FSFP)*(1-FSO)` the two parameters are dependent.

e.g. `X=100 bitABC` `Z=10 BTS/bitABC` `FSFP=0.5` (50%) `FSO=0.1`. Then we have
`Y = 1000*0.5*0.9=450`. It is also not explicitly clear how the distribution works (How much does the network get, how much does debt position holder get?). I need to know in what order the offset/fee are applied. In your case you define it that first the offset is applied, and then the fee. In the above example the debt position holder would receive `100 BTS` and the platform `450 BTS`. I think that is counterintuitive, at least I assumed that the platform would get `500 BTS` and the owner `100 BTS`.

To make a drastic example, if `FSO=0.5` holds then `FSFP` only gets its percent of half the settled value. If you really intend this kind of dependency please highlight it in a clear way in a remark. Asset management is already highly complicated and this introduces and unnecessary additional complication (imho).

bitcrab: either formula is acceptable in math, if you are not satisfied with 100/450, you can get 100/500 easily by setting a higher FSFP, here we need to put higher priority on the compatibility to core code.

so just let core team member decide, OK? I don't think we need to waste so much time on such a thing.

sschiessl-bcp: > either formula is acceptable in math, if you are not satisfied with 100/450, you can get 100/500 easily by setting a higher FSFP, here we need to put higher priority on the compatibility to core code.
> 
> so just let core team member decide, OK? I don't think we need to waste so much time on such a thing.

Certainly, if its better for performance reason. 

The meaning of the parameters needs to be explained properly, especially if they have a dependence in the displayed formula. A BSIP is the reference for the common BTS holders to decide if they want to vote for it or not, and needs to be understandable.

> either formula is acceptable in math, if you are not satisfied with 100/450, you can get 100/500 easily by setting a higher FSFP

This is just an example highlighting that the parameters can be easily misunderstood if not explained properly.

So keeping your formula is fine, but please add an explanation of the dependency of the two parameters on each other.

bitcrab: update #Specification.

actually in my draft I suppose the current `force settlement price = feed price*(1+FSO)`, as what is shown in GUI, however it seems in core code the `settlement order price = feed price/(1-FSO)`, example is that the current bitCNY force settlement price  shown at top of GUI is 0.2244, however the order price = 0.22449.

I feel it's better to let force settlement price = feed price*(1+FSO), however, seems if we follow the current core code, the specification should be modified as:

_When a force settlement is executed, the buyer sells smartcoin with quantity `X` and get collaterals in quantity `X*(1-FSO)*(1-FSFP)/feed price`, the settled debt position owner sells collaterals in quantity `X*(1-FSO)/feed price` and get smartcoin in quantity `X`, the delta in paid and received collaterals in quantity `X*FSFP*(1-FSO)/feed price` will be paid to the owner of the smartcoin as force settlement fee._

_In this scenario, `force settlement order price = feed price/(1-FSO)` is the price to the settled debt position owners(maker). and the force settlement fee is calculated out based on this order price and FSFP._

@abitmore please advise whether to do this modification.

abitmore: I think the modified text is clear. Thanks.

bitcrab: > In summary, keeping your formula is fine, but please add an explanation of the dependency of the two parameters on each other.

I think now it's clearly explained in updated specification:

_In this scenario, as a price offset parameter, FSO is irrelevant to fee charging, it just define `force settlement price = feed price/(1-FSO)`. and the force settlement fee is paid in collateral and the amount is calculated out based on `force settlement price` and `FSFP`._

PR #268, opened 2020-03-16
Update bsip-0072.md

nathanhourt:
This updates the Tanks and Taps description and specification for clarity and correctness based on changes made to the protocol during implementation.

### DIFF
--- a/bsip-0072.md
+++ b/bsip-0072.md
@@ -1,7 +1,7 @@
     BSIP: 72
     Title: Tanks and Taps: A General Solution for Smart Contract Asset Handling
     Authors: Nathan Hourt <nat.hourt@gmail.com>
-    Status: Draft
+    Status: Accepted
     Type: Protocol
     Created: 2019-08-27
     Updated: 2020-03-16
--- a/bsip-0072.md
+++ b/bsip-0072.md
@@ -4,19 +4,20 @@
     Status: Draft
     Type: Protocol
     Created: 2019-08-27
+    Updated: 2020-03-16
     Discussion: https://github.com/bitshares/bsips/issues/178
 
 # Abstract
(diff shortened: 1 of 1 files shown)


sschiessl-bcp:
What is the gist of the changes?

nathanhourt:
Most of the changes are technical, but there are also some alterations in the abstract and motivation sections to make clearer what Tanks and Taps is and is not. I realized that readers might misunderstand that TNT adds turing complete functionality, so I added a statement that it does not. Changes like that. The [rich diff](https://github.com/bitshares/bsips/pull/268/files?short_path=427ca2c#diff-427ca2cfd57e4e479c2108a1429a70ab) should make those alterations readily evident.

The bulk of the changes is in the specification section, correcting the spec to account for things added and removed during implementation. Most notably, the concept of the `sink` was dropped in favor of the `connection`, and the `deposit_source_restrictor` was removed. These changes were made because I realized during implementation that the TNT structures were too trusting by default and it was easy to create structures that work in the intended fashion, but also in unintended fashions. The updates adjust the defaults so that security is built throughout the structures and not left as an add-on at the end. Created structures should now only work in the intended fashion, no surprises.

sschiessl-bcp:
Looks good to me. Since this BSIP is already created on-chain to be voted on, I would like to ask @abitmore to have a look. I think the changes are not significant enough to warrant a re-voting.

nathanhourt:
Yeah, that was something I considered I'm not sure how much editing BSIPs is a thing... Is there some other way I should go about publishing this update?

abitmore:
Sorry, I haven't made chance to thoroughly review the original BSIP yet, but trusted other reviewers, so please don't expect my opinions on this PR soon.

MichelSantos:
> I think the changes are not significant enough to warrant a re-voting.

For whatever it's worth, I agree that the changes are technical improvements that do not alter the spirit of the specifications.

PR #269, opened 2020-04-07
Update for [0,100%] to [0,30%]

dls-cipher:
As per discussion in legal and PR terms.

### DIFF
--- a/bsip-0086.md
+++ b/bsip-0086.md
@@ -47,7 +47,7 @@ the scheduled time, at the scheduled time and after the scheduled time.
 Add a new global parameter `market_fee_network_percent` which can be updated
 by the committee only after the protocol upgrade.
 Initial value of that parameter is `0%`.
-Valid range of that parameter is `[0%, 100%]`.
+Valid range of that parameter is `[0%, 30%]`.
 
 ## When processing market fees
 


sschiessl-bcp:
Can you point to the dixussion please?

sschiessl-bcp:
I disagree with this change, as no one has adressed  my concern.

abitmore:
@sschiessl-bcp I as the original author have agreed with the change. @bitcrab as representative of a large part of voters (imo) has agreed with the change. The PR concerns are clear: a platform usually won't charge a 100% fee. Nevertheless I'd like @dls-cipher to give more info about the legal concerns he mentioned too though.

sschiessl-bcp:
> @sschiessl-bcp I as the original author have agreed with the change. @bitcrab as representative of a large part of voters (imo) has agreed with the change. The PR concerns are clear: a platform usually won't charge a 100% fee. Nevertheless I'd like @dls-cipher to give more info about the legal concerns he mentioned too though.

Neither the author or a representative can decide for the voters, simple as that. Especially, since this upper bound has been addressed in review and decided for value 100%. If @bitcrab has enough influence he could provide proof on-chain for that.

Either way, for me this is a principle question, but I guess it's fine since this will likely receive "post-launch-approval" when the change is going live on mainnet simply by the fact that witnesses that support it won't be voted out.

abitmore:
@sschiessl-bcp as I understand that you seek for procedural justice or even perfectness, in comparison, I concern more about whether the overall outcome is positive, which is another sort of perfectness. Since many voters aren't voting thoughtfully, in addition they make few efforts on improving the voting process (E.G. little feedback to changes), and education is hard, IMO focusing too much on the process rather than on the result may negatively impact overall progress, which would then negatively impact the value of the platform. It's just my thoughts though.

PR #270, opened 2020-04-12
Update bsip-0086.md clarity, destination of funds to committee

winston-1984:
BSIP0086 proposing funding the committee vesting account; not network through the reserve pool.   

QUOTE:
`The cut of market fees can go to committee-account's vesting balances.`

Throughout the document however there are several mentions of the network, which implies reserve pool.  Updated document to clarify by ensuring all mention of network was changed to committee. 

EXAMPLE:
`this BSIP seeks for a mechanism to increase the network's income `

SHOULD BE:
`this BSIP seeks for a mechanism to increase the committee's income `

I also suggest updating the variable

`market_fee_network_percent`

to 

`market_fee_committee_percent`

There is substantive difference between funding the network through reserve pool and funding the committee through committee account vesting balances.   

There was only one mention of where the funds went and that was "committee account vesting balances".    All other mentions were of "network income".   

This PR and the final resting place of BSIP86 funds might be a miscommunication error and should be discussed with voters before this implementation detail is finalized.

### DIFF
--- a/bsip-0086.md
+++ b/bsip-0086.md
@@ -10,26 +10,27 @@
 # Abstract
 
 This BSIP proposes a protocol change: Every time a trade is executed (limit order is filled) and
-if there are market fees generated, a portion of the market fees goes to the network.
+if there are market fees generated, a portion of the market fees goes to the `committee-account`.
 
-Example: The network percent of market fees is set to 50%, and 0.2% of trading volume is the set as market fee. 
-Then 0.1% of trading volume goes to the Committee.
+Example: The committee percent of market fees is set to 50%, and 0.2% of trading volume is the set as market fee. 
+Then 0.1% of trading volume goes to the `committee-account`.
 
 # Motivation
 
-The network needs more tools to generate income to support its development.
+The BitShares Committee needs more tools to generate income to support development.
 
 # Rationale
 
-A major activity in the network is trading assets. The asset owners make
-profits in the form of market trading fees. Currently, the main tools for the
-network to generate income from asset trading activities are order creation
-fee and order cancellation fee. It seems reasonable that the asset owners
-share some profits (a part of market fees) to the network to support its
-development.
+A major activity in the **network** is trading assets. The asset owners make
+profits in the form of market trading fees. 
 
-The cut of market fees can go to committee-account's vesting balances.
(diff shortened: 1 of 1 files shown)


bsip-0086.md
```diff
@@ -44,42 +45,46 @@ the scheduled time, at the scheduled time and after the scheduled time.
 
 ## New global parameter
 
-Add a new global parameter `market_fee_network_percent` which can be updated
+Add a new global parameter `market_fee_committee_percent` which can be updated
 by the committee only after the protocol upgrade.
 Initial value of that parameter is `0%`.
 Valid range of that parameter is `[0%, 100%]`.
 
+UPDATE: limit `[0,100%]` to `[0,30%]` see: https://github.com/bitshares/bsips/pull/269
```
sschiessl-bcp, line 53: The wording change to make it clear it goes to committee seems reasonable.

On the marked change here: https://github.com/bitshares/bsips/issues/194#issuecomment-607404101

sschiessl-bcp (changes requested):

sschiessl-bcp:
Does this need clarification @abitmore ?

abitmore:
I don't think it's good to change the wording from "to the network" to "to the committee". The final goal is for the network anyway, it's the spirit of the BSIP. Changing it to "for the committee" will make it a much worse taste. Due to technical difficulties we decided that it's best the committee handles the funds for the network, but it's implementation details, the funds are definitely not owned by the committee.

abitmore (changes requested):
As last message.

shulthz:
I suggest to pause this BSIP until the resolve the legal risk.

> This BSIP86 has legal risk, i suggest everyone need to consider it very carefully and seek a legal consultancy.

When BSIP86 charges the market fees from the GATEWAY.assets, the BTS system will have joint liability and must be responsible for these GATEWAY.assets, but these GATEWAY.assets were without supervision and control by the BTS system.

Gateway just likes the DApp of BTS, BTS just provides the service and takes the transaction fee of blockchain，DApp do what is not the business of BTS, but when BTS charges the market fees of GATEWAY.assets, the nature of things will become different.

sschiessl-bcp:
stale

PR #271, opened 2020-04-15
Update specification and wording of BSIP 81

abitmore:
It's found that the specification was not clear enough.
See also: https://github.com/bitshares/bitshares-core/pull/2136#issuecomment-613768577

### DIFF
--- a/bsip-0081.md
+++ b/bsip-0081.md
@@ -28,23 +28,28 @@ them to attract trading activities.
 
 There is already a flag `market_fee_percent` in asset options.
 
-Add a new flag `taker_fee_percent` into asset options, assign
-`market_fee_percent` to `taker_fee_percentage` at the consensus
-upgrade time for existing assets. The new flag can only be set
+Add a new optional flag `taker_fee_percent` into asset options, 
+which can only be set
 or updated by asset owners after the consensus upgrade.
+The valid range of the new flag is `[0%, 100%]`.
 
 Before the consensus upgrade, when an order buying that asset
 got filled, the amount `bought_amount * market_fee_percent`
-will go to the asset's accumulated fees.
+(capped by `max_market_fee`)
+will enter the asset's market fee sharing process.
 
 After the consensus upgrade, when an order buying the asset
 got filled,
-* if the order is a *maker*, the amount
-  `bought_amount * market_fee_percent` will go to the
-  asset's accumulated fees;
-* otherwise (if the order is a *taker*), the amount
-  `bought_amount * taker_fee_percent` will go to the
-  asset's accumulated fees.
+* if the order is a *maker*, or the `taker_fee_percent` flag
+  is not set in the asset, the amount
+  `bought_amount * market_fee_percent` 
+  (capped by `max_market_fee`)
+  will enter the asset's market fee sharing process;
+* otherwise (if the order is a *taker* and the
+  `taker_fee_percent` is set in the asset), the amount
+  `bought_amount * taker_fee_percent`
+  (capped by `max_market_fee`)
+  will enter the asset's market fee sharing process.
 
 # Copyright
 This document is placed in the public domain.


abitmore:
This pull request changes only implementation details, I think it doesn't require a re-vote.

bsip-0081.md
```diff
@@ -28,23 +28,28 @@ them to attract trading activities.
...
 
 After the consensus upgrade, when an order buying the asset
 got filled,
-* if the order is a *maker*, the amount
-  `bought_amount * market_fee_percent` will go to the
-  asset's accumulated fees;
-* otherwise (if the order is a *taker*), the amount
-  `bought_amount * taker_fee_percent` will go to the
-  asset's accumulated fees.
+* if the order is a *maker*, or the `taker_fee_percent` flag
```
sschiessl-bcp, line 43: "..., or the `taker_fee_percent` flag is not set in the asset ..."

Is this extra condition necessary? I think "... if the order is a *maker* ..." is already precise and whether the `taker_fee_percent` is set does not affect this anymore.

You could add in the first sentences
 - "Before the consensus upgrade, when market fee is enabled and an order buying that asset got filled ..."
 - "After the consensus upgrade, when market fee is enabled and order buying the asset got filled, ..." 
to clarify that both fees will only be collected if market fee is enabled.

bsip-0081.md
```diff
@@ -28,23 +28,28 @@ them to attract trading activities.
...
+  is not set in the asset, the amount
+  `bought_amount * market_fee_percent` 
+  (capped by `max_market_fee`)
+  will enter the asset's market fee sharing process;
+* otherwise (if the order is a *taker* and the
```
sschiessl-bcp, line 48: Would it be more precise to include 

"otherwise (the order is a *taker*) the amount `bought_amount * order_fee_percent` (capped by `max_market_fee`) will enter the asset's market fee sharing process where `order_fee_percent` is `taker_fee_percent` if it is set, otherwise `market_fee_percent`"

bsip-0081.md
abitmore: I'm not sure if you've understood the logic. I guess @MichelSantos has.
It's just how we describe it, as below:
```
if A
  do action 1
else if B
  do action 1
else
  do action 2
```
or
```
if A or B
  do action 1
else
  do action 2
```

abitmore: We're not talking about `market_fee_percent` here but `taker_fee_percent`, so your suggestion doesn't make sense.

bsip-0081.md
MichelSantos: I agree that this is the same idea.  It's a matter of different ways of explaining it.  I support whichever text will be most clear.

bsip-0081.md
MichelSantos: > You could add in the first sentences
> 
>     * "Before the consensus upgrade, when market fee is enabled and an order buying that asset got filled ..."
> 
>     * "After the consensus upgrade, when market fee is enabled and order buying the asset got filled, ..."
>       to clarify that both fees will only be collected if market fee is enabled.

This phrasing is possible too, and is technically correct.  The distinction being made in the current text is between the maker and taker fee, and it presumes that the the market fee is enabled.

sschiessl-bcp: Thanks. I think I misread the or in the first if.

sschiessl-bcp:
Since we are touching this BSIP can we add the requirement that market fee is enabled for the market fees to be taken?

This sentence is not true strictly, only if market fee flag is enabled (market fee can be set but flag disabled, correct?)
"Before the consensus upgrade, when an order buying that asset got filled, the amount [...] will enter the asset's market fee sharing process."

abitmore:
To be honest I'm not a fan of spending time on tweaking the wordings. On one hand English is not my native language, on the other hand the spirit of the BSIP and the change is clear enough, the most important thing to put our efforts on is to push forward fast. The things didn't describe in the text is simply **won't be changed**. Since the system is already quite complex and it lacks of detailed documentation, IMHO this BSIP is not the best place to explain all the details and to finish all the docs. KISS.

Actually I already feel the "capped by max market fee" sentence is already annoying as we have to repeat it again and again but it's not the most important thing to focus on in this document, I'd rather avoid mention it at all. The "market fee sharing process" phrase is similar.

If I'm not getting productive help here, I don't mind to merge it without anyone else's approval.

abitmore:
IMHO the "market fee sharing process" implies several things, one of them is "if the asset doesn't enable the charge_market_fee flag, no fee will be charged".

sschiessl-bcp:
Sounds good. Update makes it clearer

PR #218, opened 2019-09-20
BSIP 75 - Asset Owner defines MCR and MSSR values

jmjatlanta:
See Issue #96

- [ ] Add specifications
- [ ] Verify rationale
- [ ] Research risks

PR #277, opened 2020-05-28
BSIP77: Permissions and their forfeiture

MichelSantos:
Adding a permission to update the new initial collateral ratio (ICR) parameter in the same spirit as the [discussion](https://github.com/bitshares/bsips/pull/218#issuecomment-628882624) in #218


abitmore:
@MichelSantos please add yourself to the author list. Thanks.

PR #287, opened 2020-09-09
Create BSIP9999 Distributed Bitcoin Cross Claim Gateway

litepresence:


BSIP9999 Distributed Bitcoin Cross Claim Gateway
```diff
@@ -0,0 +1,110 @@
...
+Rational
+----------------
+
+A smart contract XCLAIM style gateway ensures contractual performance through collateralization
+
+Specifications
+----------------
+
+Witnesses upload bitcoin headers
+Anyone can issue "polkaBTC" to anyone by locking BTC on Bitcoin Mainnet
```
abitmore, line 51: Technically, how to achieve "locking BTC on Bitcoin Mainnet"?

abitmore (changes requested):

shulthz:
Some illustrations?...

BSIP9999 Distributed Bitcoin Cross Claim Gateway
litepresence: locking - in this sense - means Alice (vault client) has given Bob (vault operator) full possession of her BTC. 
Bob is now contractually obliged to maintain (BTS) collateral on the issuing chain of vaultBTC.

litepresence: further reading:

https://interlay.gitlab.io/polkabtc-spec/spec/issue.html#requestissue
https://interlay.gitlab.io/polkabtc-spec/spec/issue.html#executeissue

I'm working on a full Alice/Bob thought experiment write up of the issue/redeem functionality. 
I've probably made some misstatements, as my time is currently limited; hence "Draft".

matle85: My reading is Bob can only issue vaultBTC by locking up enough BTS so it's sort of like creating bitBTC and then selling it for BTC.

xeroc: > Bob is now contractually obliged to maintain (BTS) collateral

This hasn't worked out well for bitAssets in (BTS-)bear marekts in the past. Why would it work better for collateralizing off-chain assets? If you want to make it "secure" you need a high collateral ratio which makes this thing super capital **in**-efficient. The more volatile your collateral is, the worse this becomes.

matle85:
This is similar to bitBTC but maybe (/hopefully) with some HTLC or similar functionality that means BTC has to be committed in order for polkaBTC to be created (i.e. a cross chain trade and the new asset is only released / created if the two requirements - BTC sent, collateral locked up, are met) if not it is exactly the same as bitBTC pretty much?

It's better to have some buy in from BTC in creation of the asset. A far superior option is locking up BTC but has anyone got close to figuring out a road map for how that could be done?

I need to read up on how all the wrapped ETH contracts do it.

BSIP9999 Distributed Bitcoin Cross Claim Gateway
litepresence: in theory - you could back a "polkaBTC" vault construct with honest.btc as easily as you could with core token.


now bob wants to be a vault operator; he buys honest.btc on the open market - he's not a borrower at all
now he puts up his - owned - honest.btc as collateral 1:1 and that rate stays pegged so long as honest.btc retains 1:1 value with real btc; his loan would never require any maintenance.    the oracle of the peg could simply be the dex market price of honest.btc:bts vs the feed price of honest.btc:bts

litepresence: > My reading is Bob can only issue vaultBTC by locking up enough BTS so it's sort of like creating bitBTC and then selling it for BTC.

except that there is no way to sell bitBTC for BTC without an htlc arrangement or cex

here with polkaBTC bob has real BTC in hand after the tx and all accounting has been handled by the issuing chain (Bitshares) with awareness of bitcoin headers via a network of spv witness oracles.

litepresence:
> This is similar to bitBTC but maybe (/hopefully) with some HTLC or similar functionality that means BTC has to be committed in order for polkaBTC to be created (i.e. a cross chain trade and the new asset is only released / created if the two requirements - BTC sent, collateral locked up, are met) if not it is exactly the same as bitBTC pretty much?

with the mpa construct bob wishes to get bitBTC and he locks core token collateral; he "self issues" bitBTC to himself.   This is not real BTC... its just an mpa on the dex

with the polka construct bob wishes to get real BTC.   So he sets aside some collateral and fishes for a client by advertising a fee to do a swap.    alice approaches him and says I'll give you real BTC so long as you give me polkaBTC.   

alice transfers real BTC to bob
bob has core token equal to that amount then locked in an mpa type margin contract.
in exchange alice gets polkaBTC
prior to the transaction bob had core token; now he has real BTC
prior to the transaction alice had real BTC; now she has polkaBTC

alice can get her real BTC back via a redeem operation with bob - or charlie
likewise bob can get his core token back via a redeem operation with alice - or charlie







> 
> It's better to have some buy in from BTC in creation of the asset. A far superior option is locking up BTC but has anyone got close to figuring out a road map for how that could be done?
> 
> I need to read up on how all the wrapped ETH contracts do it.

abitmore:
> alice can get her real BTC back via a redeem operation with bob - or charlie

A fact is the redeem operation with real BTC has to be done on the Bitcoin blockchain. If nobody responses to her request she simply can't get any real BTC. Since she owns pBTC only on the p chain, the only thing she can do with her pBTC is to settle it for the core token of the p chain -- that's essentially what bitBTC is.

>alice transfers real BTC to bob
bob has core token equal to that amount then locked in an mpa type margin contract.
in exchange alice gets polkaBTC

If no HTLC is involved in the process Bob is able to run away when he got Alice's real BTC. Or vise versa.

Issue #295, opened 2024-04-04
New BSIP: Increase maximum total bitshares supply to boost reserve pool funds


grctest:
BSIP: 0088
    Title: Increase maximum total Bitshares supply to grow the reserve pool fund
    Authors: R
    Status: Draft
    Type: Protocol
    Created: 2024-04-18

# Abstract

Currently the Bitshares reserve pool grows through the collection of BTS fees for broadcast operations.

This BSIP proposes to substantially raise the amount of BTS tokens held in the reserve pool through a possible future core mandatory upgrade.

# Motivation

With a larger amount of BTS in the reserve pool, we can afford to pay witnesses at a higher rate, encouraging greater witness election campaigning & improving competition for these roles, strengthening the DPOS consensus mechanism.

With a larger reserve pool the development of core code and next-gen user interfaces can be funded for full-time development, as opposed to on a voluntary basis.

# Rational

The current circulating supply of BTS is: 2,995,301,686 BTS
The max circulating supply of BTS is: 3,600,570,502 BTS

Meaning there's 605m BTS in reserves (16.82% of total supply).

Considering the market cap on coinmarketcap is calculated from the circulating supply, issuing additional BTS supply to the reserves does not directly affect the market cap calculation, so even a large increase in max total supply will not directly alter the market cap calculation.

Competitor cryptocurrencies such as Dogecoin have no maximum total supply limit; there is real market tolerance for a growing core token supply.

Initial crypto whitepaper economics can change, the fees are regularly changed by committee decision as the price of BTS fluctuates for example.

Fundraising through any other means introduces various forms of risks, whether that's regulatory risk, centralized counterparty risks or can suffer from lack of sufficient participation.

There's a great deal of historical precedent for traditional publicly traded companies to issue additional shares for open market purchase, in order to fund their continued operations, this proposal is somewhat similar except the increased reserve pool isn't immediately liquid - approved worker proposals would be required to make the reserve funds liquid.

By democratically distributing these newly raised funds through the worker proposal system we will be improving the decentralization of the network. At the moment there are several centralized exchange accounts with a substantial amount of BTS in their balances, by distributing worker funds to key stakeholders these whale accounts power is reduced substantially.

# Specifications

Create a worker proposal to evaluate whether people are for or against such an increase in core token supply for reserve pool purposes.

If it passes, create secondary worker proposals to vote on the amount to increase the total supply by.

[...]

PR #296, opened 2024-04-18
BSIP 88: Increase maximum total Bitshares supply to grow the reserve pool fund

grctest:
The concept of this BSIP was introduced 2 weeks ago:
https://github.com/bitshares/bsips/issues/295
https://bitsharestalk.org/index.php?topic=34038.0

If you have any input on this idea, please do comment in this pull request, in the issue and even in bitsharestalk.

### DIFF
--- a/bsip-0088.md
+++ b/bsip-0088.md
@@ -1,0 +1,85 @@
+    BSIP: 0088
+    Title: Increase maximum total Bitshares supply to grow the reserve pool fund
+    Authors: R
+    Status: Draft
+    Type: Protocol
+    Created: 2024-04-18
+
+# Abstract
+
+Currently the Bitshares reserve pool grows through the collection of BTS fees for broadcast operations.
+
+This BSIP proposes to substantially raise the amount of BTS tokens held in the reserve pool through a possible future core mandatory upgrade.
+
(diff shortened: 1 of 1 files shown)


brekyrself:
This should move forward as soon as possible so it can be voted on by bts holders as indicated in the bsip.  BTS needs development funds and this is a very democratic way to jumpstart new workers.

"Create a worker proposal to evaluate whether people are for or against such an increase in core token supply for reserve pool purposes.

If it passes, create secondary worker proposals to vote on the amount to increase the total supply by."

grctest:
I've updated the BSIP with some simplifications, removed discussions which didn't invoke comment, and added a second option to the proposed polling mechanism.

I'm feeling more in favour of option B - just voting on whether or not to increase the maximum total supply value rather than voting on the intent, then voting on the specific amount.

Discussing the amount in here could suffice too - 100 Billion BTS up from ~3.6 Billion BTS - IMO with such a change there would be no need for further maximum total supply increases probably for decades to come.

Since there's the refund400k reserve pool barrier to entry, it doesn't really matter what the increased amount is after a certain point, but the amount we increase it by can send a powerful message.

